PR #37, opened 2020-06-03
Allow pan, rotate and zoom without changing tools

turtletooth:
Previously, if you used the keyboard shortcuts for pan, space or rotate, it would switch to those tools.  If you had a selection made, you would lose the selection.

This bypasses the tools and performs those tasks in the viewer.

Note: this currently hard codes the commands to:
Pan: Space
Rotate: Ctrl + Space
Zoom: Shift + Space

It is possible to catch custom shortcuts for those commands, but I don't want to potentially introduce latency in looking up the shortcut.


turtletooth:
This also fixes a failing assert in the camera settings popup because the hide event is being called twice.  Who knows why?

PR #47, opened 2020-06-08
Project Management Overhaul

turtletooth:
This allows projects to be created anywhere and removes the concept of project roots.

This should aid in easier project management and portability.

Changes are made to the create project dialog and to the startup popup.  Project settings is temporarily hidden, while a decision is made about whether or not to allow custom paths for project folders.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,9 +57,9 @@ after_build:
 
     copy /Y "C:\Tools\opencv\build\x64\vc14\bin\opencv_world412.dll" %CONFIGURATION%\Tahoma
 
-    mkdir "%CONFIGURATION%\Tahoma\portablestuff"
+    mkdir "%CONFIGURATION%\Tahoma\tahomastuff"
 
-    xcopy /Y /E ..\..\stuff "%CONFIGURATION%\Tahoma\portablestuff"
+    xcopy /Y /E ..\..\stuff "%CONFIGURATION%\Tahoma\tahomastuff"
     
 artifacts:
 - path: toonz\$(PLATFORM)\$(CONFIGURATION)
--- a/stuff/config/reslist.txt
+++ b/stuff/config/reslist.txt
@@ -4,94 +4,12 @@ HD 1080, 1920x1080, 16/9
 
 HD 720, 1280x720, 16/9
 
-HD 480, 853x480, 16/9
+Square, 1000x1000, 1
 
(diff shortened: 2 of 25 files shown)


artisteacher:
I always use the default locations, so I don't have strong feelings about having custom paths for the project sub-folders. I think the original new project dialog was just confusing for new learners. 

If the original custom path capability is kept, it might be nice for the custom path options to be hidden from "New project" dialog by default with a "more options" button. The button would expand/reveal the custom path options.


Looking at asset management in other programs:

Blender doesn't allow any customization of the organizational structure within a .blend file, though cache & render folders can be set anywhere, at any time. 

Unity is for game development but it does allow for custom locations to be added as needed on the fly. As you are going, you can create new folders & drag & drop in the Project window to re-organize. New scripts, new materials & etc. are created in whatever directory is currently active. Adobe Animate's library for each file is similar. This allows for great flexibility, but less automation & is likely more time consuming.

turtletooth:
I hear you.  I'm still weighing what route I want to go.  I tend to lean toward simpler, even if a little less flexible.  I think the original project management was neither simple nor as flexible as many wanted it.
I didn't get rid of the code for custom folders.  It's just hidden for now, so it could easily come back.

It's a hard balance.   OpenToonz is very powerful, but that sometimes leads to confusion and feeling overwhelmed.  I'd like a program that the user can just open and use and things feel intuitive.

turtletooth:
BTW -Did you get your build issues fixed?

artisteacher:
> I tend to lean toward simpler, even if a little less flexible.

I'm on board with that. There is still some flexibility that exists - new levels from the new level dialog can be saved anywhere. Similarly, pre-existing levels can be loaded from anywhere. Renders can be saved anywhere. So anyone that feels strongly about a different setup, can have at least levels and renders in custom locations.

No, not yet. It builds & runs fine on my own computer - I just can't move the app to a different one.
Screen Shot 2020-06-08 at 11 52 54 PM

turtletooth:
Hmm
Sounds like it is missing some qt files.  Did you bring over all the needed files?

I'm reworking the layout of the popup.  This was @DarrentTAnims idea:
newSP

artisteacher:
No, I was hoping everything would be part of the app. That must take some extra setup before building.

I like the reworked layout - there's a better sense of hierarchy in the new scene settings.

PR #48, opened 2020-06-12
Fix variety of panels

turtletooth:
This simplifies the output settings panel, adds a new style button to the palette viewer, and fixes some ui issues with the style editor.


artisteacher:
The new style button in the level palette is fantastic! In the style editor, I think the texture & trail brushes (which ignore the active color) should have the icon repeated instead of showing the average color or black. The mypaint icon is definitely way better though.

turtletooth:
Good point.  I'll take care of that.

PR #50, opened 2020-06-15
Status Bar

turtletooth:
New status bar at the bottom shows tips and help.


artisteacher:
This will be super helpful for schools. What's with "Hi John"? 
mainwindow.cpp - line 3308

artisteacher:
It would be great for this to extend to style types. Ie. a warning would be helpful if a style is active but won't work for the level type - ie. a blending raster brush on a smart raster level or a vector trail brush on a raster level. Likewise, some of the vector styles only work on strokes or only work as fills.

turtletooth:
The hi John was for @manongjohn when I first made the status bar.  I forgot I put it in there.  Good to know someone's looking at the code.

Eventually most things will have a description that pops up down in the status bar when you hover over them.  I can see the need on the styles you mentioned being important for clarity.

artisteacher:
I saw the message pop up & got curious.

PR #51, opened 2020-06-16
Perspective Assistant for Smart Raster and Raster levels

turtletooth:
This adds a perspective assistant for the brush tool in Smart Raster and Raster levels.

To use this;
With the brush tool active:
Alt + Ctrl + Click to set a vanishing point. (Multiple points are supported)

Hold Alt while drawing to snap to perspective drawing.  The assistant will figure out which vanishing point you are aiming for.

To clear the points - Shift + Alt + Click.

If no vanishing point is placed, it will use the center as a vanishing point.

### DIFF
--- a/toonz/sources/include/toonz/levelproperties.h
+++ b/toonz/sources/include/toonz/levelproperties.h
@@ -55,6 +55,8 @@ public:
       //!  are not).
       m_isStopMotionLevel;
 
+  std::vector<TPointD> m_vanishingPoints;
+
 public:
   LevelOptions();  //!< Constructs with default values.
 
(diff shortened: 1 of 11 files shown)


artisteacher:
This works really well - particularly in combination with the field guide if you want even measurements for a tile floor or etc. I love that the line isn't actually created until the alt+click is released - so you can make adjustments on the fly if you realize you drew too far or etc. It is helpful that the vanishing points can be off canvas - so important for accurate looking multi-point perspective. It's also really nice that the vanishing points are linked to their layer/level. In my opinion, it's easier to use than the equivalent functionality of the assistant tool that's in the works for OpenToonz.

It would be helpful to be able to move a vanishing point - if it didn't end up in quite the right place. In some cases it might also be helpful to remove just one vanishing point instead of clearing them all.

turtletooth:
I'm working on adding the ability to save vanishing points in the scene.  Would it be better for all levels to share vanishing points or for each level to have it's own?

I think I might add the ability to remove points to ctrl + alt instead of just using that to place them.  So if you are close enough to a point, ctrl + alt will remove the point.  If not, it will add a point.

I think that for now, I'll use removing and placing instead of making points be movable.  It just easier to code. :)

turtletooth:
I added saving assistant points per level.  You can also now remove individual points with a ctrl alt click.

PR #53, opened 2020-06-19
More Misc Fixes

turtletooth:
This adds more status messages for some menu items.
It updates some ui elements
The viewer bg color now has an option to be set from the theme


artisteacher:
You have some really helpful additions here! I would add one more message when hovering on the Raster tab - "Blending & eraser brushes only work on Raster levels."

turtletooth:
Good idea.

PR #69, opened 2020-07-03
Grab Bag

turtletooth:
New splash screen
Hex values in the style editor menu
Fix a handful of ui bugs


artisteacher:
So nice to be able to use hex values! Hex shorthand - ie. #000 instead of #000000 doesn't work, but that's not a big deal.

PR #85, opened 2020-07-13
Fix incorrectly using None interpolation type

manongjohn:
This PR fixes an problem in Tahoma that also exists in the Official OpenToonz version noted in issue https://github.com/opentoonz/opentoonz/issues/3168

When a new keyframe is created, the keyframe may inadvertently be created as a `None` interpolation type. There may be multiple ways to do this, but 1 known way is to create a new keyframe before the 1st keyframe in the column and then copy/paste the newly created keyframe after it.  This was due to not updating `m_prevType` in the next keyframe properly.

A `None` interpolation keyframe is written to the scene file as `<n>...<\n>` which can lead to a crash while loading the scene or create strange behavior with the keyframes in the column.

This fix should
1. Prevent the creation of a `None` interpolation type keyframe
2. Throw out `None` interpolation keyframes when loading the scene in case it encounters one.

### DIFF
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -757,6 +757,9 @@ void TDoubleParam::setKeyframe(int index, const TDoubleKeyframe &k) {
     dst.m_prevType = TDoubleKeyframe::None;
   else
     dst.m_prevType = keyframes[index - 1].m_type;
+
+  if (getKeyframeCount() - 1 != index)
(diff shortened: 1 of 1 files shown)


turtletooth:
Do you know why the none type exists?

turtletooth:
BTW - Thanks for doing this.  I really appreciate it.  I'm building this to test right now.

turtletooth:
Confirmed the fix.  Thanks a ton @manongjohn

jeremybullock:
Just making a note to myself.  Keyframe data is written at tdoublekeyframe.cpp

PR #87, opened 2020-07-14
Remove Letter Ports

turtletooth:
This removes the letter ports from the schematic (except for B - which now stands for Base).  Numbered ports are still available for parenting to hooks.  

This also fixes a visual glitch in the change parent and port popups in the xsheet.


artisteacher:
Is the alignment supposed to be a little lower on the right?
Screen Shot 2020-07-14 at 12 01 50 PM

PR #95, opened 2020-07-23
Display Default Palette

manongjohn:
This PR provides the following

1. Allows one to define personal default palettes for raster, smart raster and vector levels.

You can modify a default palette while on an empty column. Your `Preferences -> Drawing -> Default Level Type` setting will determine which default palette you are modifying.

Alternatively you can save an current level's palette as a default palette for that level type.

2. A new action to "Save As Default XXXX Palette" is available in the Palette window's tab context menu and in the options dropdown menu.

The palette will be saved as a default for the current level type you are on or your Default Level Type preference if you are on an empty column.  Example: If on a vector level, you will save as a Default Vector Palette.

Default palettes are saved under personal settings in tahomastuff/profiles/layouts/settings.xxx as

raster_default.tpl
smart_raster_default.tpl
vector_default.tpl

3. New levels will be created using a default level palette if one exists.

A level's palette will be created based on whatever is in the default palette at the time of creation.  So if you create a level, then modify the default palette, the next level will have the new changes, but older levels will not.

The default raster palette is only used when a project raster palette (+palettes\Raster_Drawing_Palette.tpl) doesn't already exist.
  
4. When Autocreation is enabled and you are on a blank frame in the xsheet/timeline, a palette will be shown

On columns with a vector, smart raster or raster level in it, shows the palette of the level just before the current frame or the palette of the 1st level in the column

On empty columns, shows a default palette based on your Preferences -> Drawing -> Default Level Type

### DIFF
--- a/toonz/sources/include/tpalette.h
+++ b/toonz/sources/include/tpalette.h
@@ -194,6 +194,8 @@ private:
   int m_currentFrame;         //!< Palette's current frame in style animations.
   bool m_isCleanupPalette;    //!< Whether the palette is used for cleanup
                               //! purposes.
+  bool m_isDefaultPalette;   //!< Whether the palette is a level default palette
+  int m_defaultPaletteType;  //!< Indicates default palette's type
 
   TImageP m_refImg;
   TFilePath m_refImgPath;
@@ -354,6 +356,20 @@ between RGBA color components.
       bool on);  //!< Sets the palette's identity as a cleanup palette.
   bool isCleanupPalette() const {
     return m_isCleanupPalette;
+  }  //!< Returns whether this is a cleanup palette.
+
+  void setIsDefaultPalette(bool on) {
+    m_isDefaultPalette = on;
+  }  //!< Sets the palette's identity as a cleanup palette.
+  bool isDefaultPalette() const {
+    return m_isDefaultPalette;
+  }  //!< Returns whether this is a cleanup palette.
+
+  void setDefaultPaletteType(int levelType) {
+    m_defaultPaletteType = levelType;
+  }  //!< Sets the palette's identity as a cleanup palette.
+  int getDefaultPaletteType() const {
+    return m_defaultPaletteType;
   }  //!< Returns whether this is a cleanup palette.
 
   const TImageP &getRefImg() const {
--- a/toonz/sources/include/tpalette.h
+++ b/toonz/sources/include/tpalette.h
@@ -194,8 +194,8 @@ private:
   int m_currentFrame;         //!< Palette's current frame in style animations.
   bool m_isCleanupPalette;    //!< Whether the palette is used for cleanup
                               //! purposes.
-  bool m_isDefaultPalette;   //!< Whether the palette is a level default palette
-  int m_defaultPaletteType;  //!< Indicates default palette's type
+  bool m_isDefaultPalette;    //!< Whether the palette is a default palette
+  int m_defaultPaletteType;   //!< Indicates default palette's level type
 
   TImageP m_refImg;
   TFilePath m_refImgPath;
@@ -360,17 +360,17 @@ between RGBA color components.
 
   void setIsDefaultPalette(bool on) {
     m_isDefaultPalette = on;
-  }  //!< Sets the palette's identity as a cleanup palette.
+  }  //!< Sets the palette's identity as a default palette.
   bool isDefaultPalette() const {
     return m_isDefaultPalette;
-  }  //!< Returns whether this is a cleanup palette.
+  }  //!< Returns whether this is a default palette.
 
   void setDefaultPaletteType(int levelType) {
     m_defaultPaletteType = levelType;
-  }  //!< Sets the palette's identity as a cleanup palette.
+  }  //!< Sets the default palette's level type.
   int getDefaultPaletteType() const {
     return m_defaultPaletteType;
-  }  //!< Returns whether this is a cleanup palette.
+  }  //!< Returns the default palette's level type.
 
   const TImageP &getRefImg() const {
     return m_refImg;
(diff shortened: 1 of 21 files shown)


DoctorRyan:
Awesome work my friend.  This will be SOOOO useful as I hate having to resetup my default pallette all the time.  Especially, when starting a new project.  More please.  Lol.

artisteacher:
Nice! This will definitely be a time saver. I am super excited about #4, btw!

On a related note, what do you all think about having "project palettes" for each level type? Probably as an optional feature. There could be a checkbox in the new level window to "Use project palette for level type" and/or a preference. This would behave just like the raster palette does now. All vector levels created with this enabled would share a single palette for the project. In the same way, smart raster levels would share a single palette. Or perhaps this should be at the scene level instead of the project level to prevent changed styles from unintentionally changing styles in other scenes.

manongjohn:
I also considered a project level palette for the other types while I was working on this.

I may be able to implement that for Smart (Toonz) Raster levels, with a preference setting somewhere. We'd need to account for several things like turning that setting on/off in the middle of working a level/scene/project, importing other Smart Raster levels from other sources and such. That would complicate matters to allow for backwards compatibility.  It may perhaps be best as a Level setting.  That is probably best handled in a future PR.

I don't think that would work as easily with Vector levels since the palette is stored inside the PLI file. A lot more changes are likely to support that so as not to break the current vector palette information in the file and would also have the same backwards compatibility issues the Smart Raster levels would.

artisteacher:
The project palettes do sound complicated - especially for the vector levels.

> So if you create a level, then modify the default palette, the next level will have the new changes, but older levels will not.

Perhaps the context menu could have a "replace with default palette" option for cases where you want to update an old palette to use the new default. Or I guess default palettes could display in the studio palette window. The positive side is there are a variety of options to merge/replace palettes. On the negative side, it's more of a hassle.

manongjohn:
I can look at adding a "Personal Palettes" category in Studio Palette window and preload any palettes it finds in the settings.xxx directory.  Once there, users can use the actions provided in the Studio Palette window for maintaining palettes.

turtletooth:
Dude!  This is awesome. Testing now.

turtletooth:
This is pretty neat.  I think this is going in 1.0.

toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1774,6 +1774,10 @@ void MainWindow::defineActions() {
   createRightClickMenuAction(
       MI_OverwritePalette, tr("&Save Palette"), "",
       tr("Save the current style palette as a separate file."));
+  createRightClickMenuAction(MI_SaveAsDefaultPalette,
+                             tr("&Save As Default Palette"), "",
+                             tr("Save the current style palette as the default "
```
turtletooth, line 1779: Nice use of the Status Bar.

toonz/sources/toonzlib/fullcolorpalette.cpp
```diff
@@ -43,10 +44,15 @@ TPalette *FullColorPalette::getPalette(ToonzScene *scene) {
   m_palette->addRef();
   TFilePath fullPath = scene->decodeFilePath(m_fullcolorPalettePath);
   if (!TSystem::doesExistFileOrLevel(fullPath)) {
-    // Per I francesi che hanno il nome vecchio della paletta
-    // Verra' caricata la vecchia ma salvata col nome nuovo!
+    // For the French who have the old name of the headstock
```
turtletooth, line 47: Great comment DV?

turtletooth:
> I can look at adding a "Personal Palettes" category in Studio Palette window and preload any palettes it finds in the settings.xxx directory. Once there, users can use the actions provided in the Studio Palette window for maintaining palettes.

@manongjohn Is that something you want in the scope of this PR or in a separate PR?

toonz/sources/include/tpalette.h
```diff
@@ -356,6 +358,20 @@ between RGBA color components.
     return m_isCleanupPalette;
   }  //!< Returns whether this is a cleanup palette.
 
+  void setIsDefaultPalette(bool on) {
+    m_isDefaultPalette = on;
+  }  //!< Sets the palette's identity as a cleanup palette.
```
turtletooth, line 363: Copied comments.  Not a big deal.

toonz/sources/toonz/iocommand.cpp
```diff
@@ -3001,6 +3002,82 @@ class OverwritePaletteCommandHandler final : public MenuItemHandler {
...
+      levelType = palette->getDefaultPaletteType();
+    } else {
+      TXshLevel *level = TApp::instance()->getCurrentLevel()->getLevel();
+      if (!level) {
+        DVGui::warning("No current level.");
+        return;
+      }
+      TXshSimpleLevel *sl = level->getSimpleLevel();
+      if (!sl) {
+        DVGui::warning("Current level is not a simple level.");
```
turtletooth, line 3033: Might be more clear as "Current level is not a drawing level"?

turtletooth:
@manongjohn I'm fine merging as is.  Any little tweaks can happen after.  This is going to make getting up and running on new projects much smoother for people who have favorite brushes or standard palettes.  Thanks a ton.

toonz/sources/toonzlib/fullcolorpalette.cpp
manongjohn: lol. I Google translated it like I normally do when I come across non-english comments and paste it as is. Sometimes it makes sense. :)

Would you like me to revert it?

toonz/sources/toonz/iocommand.cpp
manongjohn: Will fix

manongjohn:
I've already made the change to add Personal Palettes to the Studio Palette Window and have a few other changes which I will commit shortly.

Note: I moved the default palette location to `settings.xxx\palettes` just to keep the settings folder a little more organized.

turtletooth:
Thanks a ton!  Great work.

Issue #97, opened 2020-07-24
Task and command line rendering issues


artisteacher:
**Describe the bug**
Currently, task rendering & rendering from command line fails on macOS. There seems to be multiple issues that need to be corrected for rendering to be successful:
- [x] dylib files can't be found due to an issue with `@rpath` 
@otakuto's PR https://github.com/opentoonz/opentoonz/pull/3031 includes changes to the dylib paths that correct the `@rpath` issue for macOS. I can submit a PR with those path corrections.

With the path issue corrected:
- [ ] command line rendering succeeds but tcomposer crashes at the end due to `Illegal instruction: 4`
[cl_render_crash.txt](https://github.com/turtletooth/tahoma/files/4970591/cl_render_crash.txt)
- [ ] task rendering still fails & tcomposer crashes
[tasks_render_crash.txt](https://github.com/turtletooth/tahoma/files/4970590/tasks_render_crash.txt)

Newly discovered:
- [ ] FFmpeg formats fail during command line rendering - `Toonz Exception` message displays in Terminal
<img width="425" alt="Screen Shot 2020-07-24 at 10 06 14 PM" src="https://user-images.githubusercontent.com/21159116/88446934-29811700-cdfc-11ea-8167-5f99a24e7eee.png">

Since the command line argument needs to be adjusted for a successful render, perhaps there's a similar issue when rendering through the task window?

**To Reproduce**
List steps to reproduce the behavior.
1. In the Tasks window, add a scene to render
2. Press "Start" --> render fails, icons turn red
3. Copy the command line data
4. Paste command line data into a .txt file to complete the tcomposer path and also add a toonzroot location
5. Copy & paste adjusted command line argument into Terminal
6. Press enter --> gif or mp4 render fails

**Expected behavior**
Both from the tasks window and from command line, the render should succeed without any crashing.

**Device Information:**
 - OS: macOS Catalina and Mojave
 - Tahoma Version: build from current master (7/23/20)

turtletooth:
Does this happen on both the beta and your current build?

artisteacher:
Yes, this is a problem with both.

artisteacher:
Without path corrections, this is the error that shows up:
<img width="600" alt="Screen Shot 2020-07-24 at 12 34 24 PM" src="https://user-images.githubusercontent.com/21159116/88414070-45f46380-cdaa-11ea-8eac-0e850000e3d7.png">

turtletooth:
[...]

PR #99, opened 2020-07-25
Make command line rendering function

artisteacher:
This uses parts of PR [opentoonz/opentoonz#3031](https://github.com/opentoonz/opentoonz/pull/3031) by @otakuto with some small additions so that command line arguments can function. It resolves some of #97 - command line rendering works for default image formats on macOS.

Changing the cmake flags & using `@executable_path` instead of `@rpath` allows for libraries like tnzcore.dylib to be found & utilized.

Known issues:
1. There is a crash at the end of the render process (though default image formats do render successfully before the crash)
2. FFmpeg formats don't render from command line yet
3. Task rendering still fails


turtletooth:
Thanks @artisteacher

PR #100, opened 2020-07-25
Add Save Scene warning popup

manongjohn:
This PR adds a Save Scene warning popup, primarily for new users who may try to use Save Scene and not understand its impact.

image

The popup can be disabled by checking `Do not show again` option.  It can be disabled/re-enabled in `Preferences -> Saving` as well.

Feel free to suggest a better message.

### DIFF
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -197,6 +197,8 @@ enum PreferencesItemId {
   // LineTestFpsCapture,
   // guidedDrawingType,
 
+  doNotShowPopupSaveScene,
+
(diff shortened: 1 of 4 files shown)


artisteacher:
It would be good to simplify a little. What about:
"This only saves scene data, not unsaved level changes."

manongjohn:
Should I keep the question? Shorten it? Maybe keep just your message and change the buttons to `OK` and `Save All Instead`?

turtletooth:
I’d keep the question.

artisteacher:
If you want to shorten the question, what about this:
"Did you want to save only the scene, or save all?"

turtletooth:
Great!

PR #103, opened 2020-07-26
Swap columns in the command bar and fix wording.

turtletooth:
This swaps the two main columns in the command bar.  That seems to be a more expected workflow.  I also changed some wording to clarify a bit.
cb


adamearle:
brilliant

PR #105, opened 2020-07-27
Make short play work universally

### LINKED ISSUE #9: Tweak Short Play Feature
From @DarrenTAnims:
The short play feature, as described in opentoonz/opentoonz#2644 plays the last 8 frames in the current column, which works great when using it with the stop motion feature as you're almost always adding frames to the end. But it would be useful to change this to play the last 8 frames (possibly this number could be a preference?) up to the current frame. Then as you're editing the timing in the middle of your animation, you could select a frame and short play to see the timing of that.

turtletooth:
Closes #9 

Allows short play to work even in the middle of the timeline/xsheet.  No longer only works from the end.

@manongjohn Do you have a couple minutes to look at this?

### DIFF
--- a/toonz/sources/include/toonzqt/flipconsole.h
+++ b/toonz/sources/include/toonzqt/flipconsole.h
@@ -286,6 +286,7 @@ public:
   }
   UINT getCustomizeMask() { return m_customizeMask; }
   void setCustomizemask(UINT mask);
+  void setStopAt(int frame);
 
   // the main (currently the only) use for current flipconsole and setActive is
   // to
@@ -362,6 +363,7 @@ private:
   std::vector<int> m_gadgetsMask;
   int m_from, m_to, m_step;
   int m_currentFrame, m_framesCount;
+  int m_stopAt = -1;
   ImagePainter::VisualSettings m_settings;
 
   bool m_isPlay;
--- a/toonz/sources/stopmotion/stopmotion.cpp
+++ b/toonz/sources/stopmotion/stopmotion.cpp
@@ -417,7 +417,7 @@ void StopMotion::onPlaybackChanged() {
 
   int frame     = TApp::instance()->getCurrentFrame()->getFrame();
   int lastFrame = TApp::instance()->getCurrentFrame()->getMaxFrameIndex();
-  if (frame == 0 || frame == lastFrame) {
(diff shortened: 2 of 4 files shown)


manongjohn:
Testing this, is Short Play supposed to be constrained by Start/Stop markers?

I'm seeing the following odd behavior, to me at least, when using Short Play in regards to Start/Stop markers

Test scenarios for examples below: I have a 79 frame animation, and I arbitrarily set my Start at 25 and Stop at 60, and set Short Play to 10.

- If you are on a frame before the Start Marker, it will jump to the Start Marker and then move to the next frame up then stop.
   - Example: Go to Frame 10 of the animation and use Short Play. Jumps to Frame 25, then stops at 26
- If you are within Start/Stop markers, it will only go as far back as the Start marker.
   - Example: Go to Frame 30 and use Short Play. Plays only between 25 - 30
   - Consequently, if you are on the Start Marker (Frame 30), and hit Short Play, it will just move to the next frame.
- If you are on a frame after the Stop marker...
  - If the furthest frame back is before the Stop marker, it will play from that frame to the Stop marker, then jump to Start marker and stop
     - Example: Go to Frame 66 and hit Short Play.  It will play 56-60, then jump to Frame 25
  - If the furthest frame is after the Stop marker, it will play everything between Start and Stop markers then jump back to Start marker
     - Example: Go to Frame 72 and hit Short Play.  Plays between Frames 25-60 then jumps back to 25.

turtletooth:
I see what you are saying @manongjohn 
The jumping to the start marker is because rewind after playback is on.

I guess the question is what should be the default behavior when short play is called outside of the markers.  If you hit play outside the markers, it plays the range inside the markers.  

With markers set, it is reasonable that even short play should be constrained to the markers.

turtletooth:
I made changes to take play markers into account.  If short play is pressed before a play marker, playback will start at the first marker and play for as long as the short play preference tells it to.  If play is pressed after the last marker, short play will go back and play until the last marker.  This also keeps rewind after playback from kicking it back to the beginning.

manongjohn:
Part of me wishes this would ignore the current start/stop markers and just play the last set of frames based on where I am in the animation.

turtletooth:
I agree @manongjohn - I changed it.

PR #107, opened 2020-07-29
Fix saving default palette

### LINKED ISSUE #106: Reproducible crash with Save As Default Smart Raster Palette
Video:
https://streamable.com/k2lg7b#

Repro steps (upon starting Tahoma):
1. Create a Smart Raster Level and use the `Save As Default Smart Raster Palette` command.
2. Create a new Scene via `File -> New Scene`
3. Select a cell in Col1 / Select the camera column's header then Col1's header

Notes:
I was only able to get the crash with Smart Raster, following the same steps with the other 2 level types didn't trigger it.
Although a Smart Raster level does not need to exist to call the `Save As Default Smart Raster Palette` command, the creation of a Smart Raster level in step 1 is necessary for the crash to occur.
Overwriting an already present default palette does not seem to be a factor, the crash still occurs from a fresh installation.
This crash was initially discovered by Rodney, also on a Windows machine.

Tahoma version: 1.0 Beta 4 (Jul 27)
OS: Win 10

manongjohn:
This PR fixes #106 

Issue: When saving a level palette as a default palette, it was storing the level palette as the default palette.  When switching scenes, the level palette would be destroyed, leaving a non-existing palette as the default which resulted in a crash when accessing it.

Resolution: When saving a level palette as a default palette, a clone of the level palette is stored as the default palette instead so it would not be destructed when switching levels.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -3100,6 +3100,13 @@ public:
       if (ret == 2 || ret == 0) return;
     }
 
+    if (!palette->isDefaultPalette()) {
+      palette = palette->clone();
+      palette->setPaletteName(L"Default " + displayStr.toStdWString() +
+                              L" Palette");
+      palette->setIsDefaultPalette(true);
+    }
+
     StudioPalette::instance()->save(palettePath, palette);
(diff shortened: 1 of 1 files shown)


turtletooth:
Just tested.  Works great.  Thanks for the quick fix @manongjohn

PR #109, opened 2020-08-02
Fix undo guided strokes clearing frames

### LINKED ISSUE #108: Vector Guided Drawing Inbetween - Undo Deletes Drawing
**Describe the bug**
Undoing a `Create Blank Drawing` action under some circumstances causes the last drawing in a level to be deleted. This action cannot be undone (redone?).

Saving and resetting Tahoma before undoing prevents this from happening.

**To Reproduce**
- In settings: `Enable Autocreation` set to 'On' and `Numbering System` to 'Incremental'.
- Create a new vector level.
- Create a second drawing using `Create Blank Drawing`.
- Draw a line in each drawing.
- Create a new drawing between the two with `Create Blank Drawing`.
- Inbetween the stroke it using `Select Guide Strokes - Tween Mode`.
- Undo the inbetween action.

Replicated in Tahoma Beta 4 release.

_demo

manongjohn:
This PR fixes #108 

Issue: When auto inbetweening occurs and the strokes are drawn and added to undo stack, information from the tool's previous action, creating and renumbering frames, is also stored.  As a result, when undoing the auto inbetween, OT thinks it has to also undo the create/renumber action.

Fix: Reset tool's internal information back to baseline prior to performing the auto inbetweening.

### DIFF
--- a/toonz/sources/tnztools/tool.cpp
+++ b/toonz/sources/tnztools/tool.cpp
@@ -1303,6 +1303,7 @@ void TTool::tweenSelectedGuideStrokes() {
   if (vbTool) {
     m_isFrameCreated = false;
     m_isLevelCreated = false;
+    vbTool->touchImage();
     vbTool->setViewer(m_viewer);
     vbTool->doFrameRangeStrokes(
         bFid, bStroke, fFid, fStroke,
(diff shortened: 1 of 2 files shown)


melieconiek:
Seems to fix the issue. Thank you @manongjohn. 👍

turtletooth:
Confirmed the fix.  Thanks @manongjohn 

On a side note, do you know why creating a blank drawing in between vector frames creates a no icon in the level strip?

manongjohn:
It appears to happens with all level types.  It appears I forgot to invalidate the icons so it forces it to refresh after it is created.

I'll submit a PR shortly to fix it.

PR #111, opened 2020-08-03
Fix no icon for newly created blank frames

manongjohn:
This PR fixes an issue where newly created blank frames create a `no icon` frame in the level strip.  If you selected multiple cells and used create blank frame, the icons would be larger than expected, creating a mess in the level strip.

Issue: Icon manager not triggered to refresh the icons after the blank frame was created.

Fix: Invalidate level frame upon initial creation and during redo so the icon manager knows to refresh the icon in the level strip.


turtletooth:
Awesome. Thanks.

PR #115, opened 2020-08-05
Update tahoma_build.sh

artisteacher:
Automatically copy and place FFmpeg and the Canon SDK framework within the Tahoma app for macOS release.

### DIFF
--- a/tahoma_build.sh
+++ b/tahoma_build.sh
@@ -5,6 +5,14 @@ export TOONZDIR=toonz/build/toonz/Release
 echo ">>> Copying stuff to $TOONZDIR/Tahoma.app/tahomastuff"
 cp -R stuff $TOONZDIR/Tahoma.app/tahomastuff
 
+echo ">>> Copying FFmpeg to $TOONZDIR/Tahoma.app/ffmpeg"
+# This assumes an FFmpeg directory is in Downloads & that the executables are directly in that directory
+# (not in ffmpeg/bin or etc.) - path may need to be adjusted
+cp -R ~/Downloads/ffmpeg $TOONZDIR/Tahoma.app/ffmpeg
+
+echo ">>> Copying canon framework to $TOONZDIR/Tahoma.app/Contents/Frameworks/EDSDK.Framework"
+cp -R thirdparty/canon/Framework $TOONZDIR/Tahoma.app/Contents/Frameworks
+
 echo ">>> Configuring Tahoma.app for deployment"
 
 $QTDIR/bin/macdeployqt $TOONZDIR/Tahoma.app -verbose=0 -always-overwrite \


tahoma_build.sh
```diff
@@ -5,6 +5,14 @@ export TOONZDIR=toonz/build/toonz/Release
 echo ">>> Copying stuff to $TOONZDIR/Tahoma.app/tahomastuff"
 cp -R stuff $TOONZDIR/Tahoma.app/tahomastuff
 
+echo ">>> Copying FFmpeg to $TOONZDIR/Tahoma.app/ffmpeg"
+# This assumes an FFmpeg directory is in Downloads & that the executables are directly in that directory
+# (not in ffmpeg/bin or etc.) - path may need to be adjusted
+cp -R ~/Downloads/ffmpeg $TOONZDIR/Tahoma.app/ffmpeg
```
manongjohn, line 11: Copying ffmpeg like this might work for local builds, but will that still allow it to work on other machines that don't have ffmpeg installed?  My concern here is that all the libs that the ffmpeg executables need may not be included in the build by doing it this way and if a user doesn't have those, it will cause Tahoma to crash when it tries to use it.

I'm wondering if the executables for ffmpeg should actually be included in the macdeployqt command list, so it pulls the ffmpeg executables + any supporting libraries it needs.

artisteacher: I did test on another mac & everything was fine. I've always just given my students a directory with only the ffmpeg executables & they always seemed to work fine. We did have a good deal of render crashing on complex scenes but it happened with any render format, not just gif/mp4.

manongjohn: Ok. I know I've tried doing something similar with other applications and had problems.  Since you've confirmed it work as is, I'm ok with it.

I need to set up the Linux package builds in my other PR to do the same thing, though I can't do much for the Canon stuff.

Can we move this tahoma_build.sh script to the `ci-scripts/osx` ?  Can also name it travis-buildpkg.sh, similar to the linux one I created for this and use in OTX for building.  We'd need to put logic around the canon piece to only include it if it is detected.

turtletooth: Sorry for being away for a few days.  Our dishswasher broke and flooded the kitchen and into the basement ceiling.  I've been working on getting our kitchen back in order.

From what I understand, this script is only going to be used to make the release and beta builds.  I'm not sure this needs to be incorporated into the travis builds.  Please correct me if I'm wrong.  Since we can't include the Canon SDK in the repo, any build for release or beta needs to be compiled on a local machine.

manongjohn:
No it doesn't need to be used by travis. I would use it for making local builds and testing on other machines though I'd need to comment out the Canon piece unless the scripts only did that if it detected Canon libraries somewhere.

If you wanted travis to build "nightlies" (when stuff merges to master) but without Canon support for general testing, it could be used for that.

My apologies if I'm overstepping. Just looking at things from many different angles.

turtletooth:
No need to apologize at all.  It's good to have other perspectives.  I've never launched a program before, so I'll take all the help I can get.  

I like the idea of nightlies.  That would great for testing.  If you know how to set up the process, would you be willing to help with that?  

I think using a simple if to check if the canon directory exists would work:
https://www.cyberciti.biz/faq/howto-check-if-a-directory-exists-in-a-bash-shellscript/

I'm not a bash scripter though,

artisteacher:
Thanks, that page was helpful. The If statement seems to be working. Moving the script is proving fussier. I tried updating  paths on lines 3, 6, & 16:
Screen Shot 2020-08-07 at 9 24 27 AM

And now I'm getting this in Terminal:
Screen Shot 2020-08-07 at 9 26 19 AM
It wasn't happening before. Is there a path change I missed? Or should something be written differently?

manongjohn:
If you don't mind, I'd like to submit my working build package script for macOS. It will automatically download and install static builds of ffmpeg/ffprobe 4.3 from ffmpeg.zeranoe.com which we can use for automated builds in the future.

artisteacher:
That's fine with me. I'll close this then.

PR #126, opened 2020-08-26
UI update and Icons from Konero

turtletooth:
Thanks @konero for the amazing work.


artisteacher:
@konero - The icons are gorgeous!

@turtletooth - I liked your vector color better than the old OT color. I hope you'll add it back!

Otherwise, the key icon really should be the same in the Timeline/Xsheet as it is in the viewer & etc. 

In all of the dark themes, when render visibility is turned off for a column, it's hard to see the off state of the render visibility, lock, and settings buttons in the column header. The level of contrast is fine in lighter themes.
Screen Shot 2020-08-26 at 10 57 31 PM

turtletooth:
@artisteacher I put the colors back to the way they had been in Tahoma up til now.

artisteacher:
Neutral & light themes should have the vector color adjusted too.

turtletooth:
@konero Thanks again for all your work.  I plan on merging this later today.  Have you received any feedback that indicates that I should hold off?

PR #130, opened 2020-08-31
Display Fx Icon as timeline/xsheet thumbnail

manongjohn:
With the recent changes to the UI, there was a request to replace the text displayed in thumbnails for Fx columns (Zerary Fx) that appear in the xsheet/timeline with some image.  If was agreed to take the icon used in the Fx Schematic node view and make that the thumbnail for these columns.

image

The current icons are PNG will look a little pixelated since they are smaller than the thumbnail size. It's not too bad for timeline.  There is an endeavor to replace them with SVG icons, but will most likely be done separately.


turtletooth:
That's awesome!  Thanks for this.
I think 1.0 will drop this week.

PR #135, opened 2020-09-04
Allow timeline to be fixed height on starting

### LINKED ISSUE #134: Timeline clipped when starting Tahoma maximized
**Describe the bug**
The top of the timeline appears to be clipped by the panes above it when I have my room arranged a certain way and Tahoma starts up maximized.

image

I recall another user on Discord also encountered a similar issue, though I don't know what their room layout looked like.

**To Reproduce**
1. Add a Command Bar and rearrange the 2D room to look like this
image
2. Maximize Tahoma
3. Restart Tahoma -> timeline becomes clipped

It only appears to happen when starting maximized in the 2D room with this configuration.

It corrects itself when you unmaximize.  When you restart Tahoma unmaximized, it's also ok.  If you start Tahoma maximized but in another room, then switch to this one, it's also ok. 

**Device Information:**
 - OS: Windows 10
 - Tahoma Version: 1.0

turtletooth:
This fixes #134 

This removes the code in mainwindow.cpp that restored window states after starting.  That code was originally added to keep the timeline from expanding vertically upon restarting.  In some instances like stop-motion, the timeline doesn't need to be very tall at all, but Tahoma kept expanding the timeline vertically.

I am trying another solution to this problem.  This reuses the code that makes the level strip a fixed width on resizing horizontally to do the same thing for the timeline vertically.

@manongjohn When you get the chance, would you mind looking at this to see if I made any glaring mistakes?

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1240,7 +1240,7 @@ void MainWindow::onMenuCheckboxChanged() {
     FieldGuideToggleAction = isChecked;
   else if (cm->getAction(MI_RasterizePli) == action) {
     if (!QGLPixelBuffer::hasOpenGLPbuffers()) isChecked = 0;
-    RasterizePliToggleAction = isChecked;
+    RasterizePliToggleAction                            = isChecked;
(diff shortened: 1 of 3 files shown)


manongjohn:
LGTM

PR #148, opened 2020-09-07
Remove unnecessary click with Vendor name

jeremybullock:
This removes the vendor category under plugins.  Now the plugins section will simply show the list.  This isn't meant to detract from the work of the author of the plugins.  It is meant to save a click.

If eventually the list of plugins becomes large enough, this can be reverted to organize the plugins better.


jeremybullock:
Now that we are including other plugins, I think this can be closed.

PR #154, opened 2020-09-10
Allow Smart Raster and Raster brushes to snap every 15 degrees

jeremybullock:
This changes the control/command behavior with raster and smart raster brushes.  Instead of snapping at 45 degrees, this will allow snapping every 15 degrees.  This adds greater flexibility and will allow for isometric workflows.


jeremybullock:
I'm planning on adding all the vanishing point and control/command abilities to vector levels too.  
Anyone want to chime in on what commands should be used?  

Right now, alt on a vector level toggles snapping.  On the raster levels, alt snaps to a vanishing point.  

What about Ctrl + Shift for one of those?  That is the only combination not used I think.

artisteacher:
Consistency between level types is definitely helpful wherever possible.

I'm fine either way but I'm slightly leaning towards:

Ctrl+shift - Snap to vanishing points
Alt - Vector snapping

jeremybullock:
I just remembered that ctrl is also used right now to continue a frame range section.  I'm thinking of moving that to Ctrl+Shift+Alt since that would get used less frequently (except by beeheemooth).

Issue #112, opened 2020-08-04
Some default window sizes are too small


artisteacher:
**Describe the bug**
When opening Tahoma for the first time, the Timeline, Style Editor, and Stop Motion Controls windows are too small. Expanding the program window/going fullscreen does expand the Timeline. However the width of the Style Editor & Stop Motion Controls remain too narrow without manual adjustment. The Stop Motion Controls window is the worst - it's barely visible at all.

**To Reproduce**
List steps to reproduce the behavior.
1. Open Tahoma for the first time using a computer without adjusting the program window size
2. Go fullscreen

**Expected behavior**
The Timeline should show at least 1 row by default. The Stop Motion Controls should be visible. Ideally, all the tabs in the Style Editor should be visible. Personally, I'd also like the program window to fill the whole screen on open.

**Screenshots**
<img width="960" alt="Screen Shot 2020-08-04 at 11 12 57 AM" src="https://user-images.githubusercontent.com/21159116/89311347-dc931100-d643-11ea-8516-b462548d501b.png">
<img width="961" alt="Screen Shot 2020-08-04 at 11 13 15 AM" src="https://user-images.githubusercontent.com/21159116/89311350-ddc43e00-d643-11ea-9e15-d3f39971024c.png">

**Device Information:**
 - OS: macOS 10.14.6 & 10.15.4
 - Tahoma Version: Beta (8/4/20)

turtletooth:
Thanks for the info.  What is the resolution of those screens?  I built the rooms to default to a good size at 1920 x 1080 but we can tweak them.

artisteacher:
1680 x 1050 & 1440 x 900. Thanks!

PR #155, opened 2020-09-11
Fix Mac Crash on Close

### LINKED ISSUE #141: Crash on close in macOS
**Describe the bug**
Sometimes closing Tahoma from the 2D room causes a crash. When this happens, it resets any size adjustments to the windows in the 2D room back to the default. A crash notification doesn't come up, but there's a crash report visible in Console. If this happens once, it happens every time closing from the 2D room.

**To Reproduce**
1. Close Tahoma with the 2D room open (closing method makes no difference)

**Screenshots**
Screen Shot 2020-09-05 at 12 04 49 PM

It looks like this might be the problem:
Screen Shot 2020-09-05 at 12 09 09 PM

**Device Information:**
 - OS: [macOS 10.14.6 & 10.15.6]
 - Tahoma Version: [1.0.1]

jeremybullock:
Closes #141 Hopefully.

Adds a destructor to Orientation and derived classes.

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -218,6 +218,7 @@ protected:
   map<PredefinedFlag, bool> _flags;
 
 public:
+  virtual ~Orientation(){};
   virtual CellPosition xyToPosition(const QPoint &xy,
                                     const ColumnFan *fan) const = 0;
   virtual QPoint positionToXY(const CellPosition &position,
--- a/toonz/sources/toonzlib/orientation.cpp
+++ b/toonz/sources/toonzlib/orientation.cpp
@@ -52,6 +52,7 @@ class TopToBottomOrientation : public Orientation {
 
 public:
   TopToBottomOrientation();
+  virtual ~TopToBottomOrientation(){};
 
   virtual CellPosition xyToPosition(const QPoint &xy,
                                     const ColumnFan *fan) const override;
@@ -117,6 +118,7 @@ class LeftToRightOrientation : public Orientation {
 
 public:
   LeftToRightOrientation();
+  virtual ~LeftToRightOrientation(){};
 
   virtual CellPosition xyToPosition(const QPoint &xy,
                                     const ColumnFan *fan) const override;


jeremybullock:
@artisteacher When you get the chance, would you mind testing this out?

jeremybullock:
Do you know when this started happening?  If you add another panel to the 2D room, close and restart, does it stick around?

#135 shouldn't have affected things that way, but I can make it Windows only if it seems to.

jeremybullock:
If you want to try commenting out lines 524-527 and 552-557 in docklayout.cpp, that would check to see if #135 was the issue.

artisteacher:
Commenting out those lines does stop the timeline from shrinking upon reopening. The stop motion controls & the style editor do still get narrower. The stop motion controls are barely visible, so that's the worst. The style editor & palette shrink a little, but not nearly as much:

Adjusted width size:
Screen Shot 2020-09-11 at 5 15 23 PM

Width after reopening:
Screen Shot 2020-09-11 at 5 15 11 PM

jeremybullock:
Do you want me to make all the fixed size panels be windows only?

artisteacher:
What do you mean by windows only?

artisteacher:
Oh, do you mean the OS? I like that the fixed windows stay the same size when resizing the window. It's preferable to the old behavior where they stretched out disproportionally. 

It's just strange how they resize in the active room upon reopening. Ie. if I make make the stop motion controls window wider and then close Tahoma from the 2D room, the stop motion controls window stays the same when I switch back to the stop motion room after opening. However, if I close from the stop motion room instead of the 2D room, the stop motion controls are practically hidden upon reopening. The active room on close/open shouldn't change the window size.

jeremybullock:
@artisteacher I'm a little confused too.  Did it do that in the first Mac beta you made?

Ok if I merge this to at least solve the crash on close?

artisteacher:
> Did it do that in the first Mac beta you made?

I think so - that's likely why the size of the windows seemed so off #112.

Sure, it's fine to merge this for the crash & deal with the resizing glitch in a new PR. That issue definitely doesn't happen on OT. Could it relate to #59?

PR #164, opened 2020-09-12
Fix override warnings

artisteacher:
There are 100+ warnings due to ______ overriding a member function without being marked override. This adds override in those cases to silence the warnings.
Screen Shot 2020-09-11 at 10 59 25 PM


jeremybullock:
Thanks for doing this.  Right now, there are around 1500 warnings that happen on a Windows build.  This will put a little dent in that.

PR #168, opened 2020-09-14
Allow Transparency on Camera View

jeremybullock:
This creates a new slider when you right click the camera view icon in the viewer.  The slider allows the outside camera view to have variable transparency.  This can be helpful in doing stop motion or in planning shots that use backgrounds wider than the camera.  This way, you can emphasize the camera view, but still see what is happening outside the view.

Here is a badly drawn example:
cam


artisteacher:
With this feature, do we need a separate camera stand & camera view? What if one of those controls was replaced with one for transparency?

jeremybullock:
Good question.  I think it still might be nice to have both views for quickly switching.  But I'm open to other ideas.

artisteacher:
Personally I'd replace one with a toggle for the transparency slider. The slider could include a reset button (trash icon like reseting the brush settings?) to return to full opacity quickly. It would be 2 clicks instead of 1, but it would still be fairly fast.

artisteacher:
Or the slider itself could replace one of the buttons...

artisteacher:
It would also be nice for the transparency level to update faster. It takes a couple seconds to update, which makes refining the level of transparency a little fussy. Great feature though!

jeremybullock:
@artisteacher I'll take a look at seeing if I can make it respond.   Mine didn't seem laggy, but didn't change until I had closed the slider dropdown.  I'll test again.

jeremybullock:
#172 Should take care of the speed.

DarrenTAnims:
I really like this feature, but it's not very discoverable. I wonder if we should replace those 2 buttons with the slider direct on the toolbar.

PR #172, opened 2020-09-15
Make viewer transparency respond faster and add to combo viewer

jeremybullock:
This makes adjusting the transparency slider respond faster and also adds the feature to the combo viewer.


artisteacher:
The response is so much better now. Thanks for the update!

PR #182, opened 2020-09-18
Fix copy/paste keyframe causing bad data

### LINKED ISSUE #178: Found the location of the cause of <> tags being written.
I managed to create a scene that would consistently create a <> keyframe.

If you download the scene and move the keyframe on frame 7 to frame 20(ish?) it creates a <> keyframe.

Reverting opentoonz/opentoonz#2647 seems to keep the <> from being written.  Something in that pr is confusing the logic.

You can watch the debug values in setkeyframe in tstageobject.cpp to see the m_type of the keyframe during paste.

[sandbox.zip](https://github.com/turtletooth/tahoma/files/5237506/sandbox.zip)

manongjohn:
This PR fixes #178  and should stop the generation of `<n>` or `<>` keyframe data when saving.

The logic that was there assumed and worked correctly for full keyframes, but not partial keyframes.  When copy-pasting partial keyframes, it would set m_prevType on channels that were not keyframed.  In addition to that, the newly pasted key's interpolation was being incorrectly updated which would lead to the `<n>`/`<>` when pasting before a partial keyframe.

Added logic to only update m_prevTypes if the channel was keyframed on the newly pasted key and the next/previous key also had that channel keyframed.

### DIFF
--- a/toonz/sources/toonz/keyframedata.cpp
+++ b/toonz/sources/toonz/keyframedata.cpp
@@ -137,9 +137,12 @@ bool TKeyframeData::getKeyframes(std::set<Position> &positions,
         // interpolation between them using preference setting
         TStageObject::Keyframe prevKey = pegbar->getKeyframe(kL);
         for (int i = 0; i < TStageObject::T_ChannelCount; i++) {
-          prevKey.m_channels[i].m_type =
-              TDoubleKeyframe::Type(Preferences::instance()->getKeyframeType());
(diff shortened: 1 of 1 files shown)


jeremybullock:
I confirmed the fix.  Thanks @manongjohn for the fast fix.  It we see these things happening again, we can also add a check that m_type is an actual usable value.

LGTM.

PR #185, opened 2020-09-19
Add perspective grids to vanishing points

jeremybullock:
This adds an option to have a perspective grid shown with vanishing points.  It is toggle-able via the right-click menu in the viewer.

colors2

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -335,6 +335,7 @@ void FullColorBrushTool::leftButtonDown(const TPointD &pos,
     if (!deletedPoint) m_assistantPoints.push_back(pos);
     simLevel->getProperties()->setVanishingPoints(m_assistantPoints);
     level->setDirtyFlag(true);
+    invalidate();
     return;
   }
   if (e.isAltPressed() && e.isShiftPressed() && !e.isCtrlPressed()) {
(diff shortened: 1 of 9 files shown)


jeremybullock:
The PR makes this look a lot bigger code wise than it is.  This is probably around 150 lines.  The rest is just clang formatting.

jeremybullock:
Builds are failing because I am out of Git LFS bandwidth.

artisteacher:
Great addition!

artisteacher:
Should the position of the grid guidelines shift on zoom in/out?

manongjohn:
I like this feature, however the guidelines shift when rotating/moving/zooming the canvas, destroying any frame of reference you are trying to maintain. I would expect once they are set, they would scale/rotate accordingly as if they were drawn on the canvas.

I would recommend either allowing a box selection when trying to remove perspective points, or make the points a little larger so they are easier to target.

jeremybullock:
I have some ideas for tweaking the guides so they stay in a relative position to the canvas.  Right now, they are being drawn at intervals to the viewer.

jeremybullock:
I made the lines fixed relative to the vanishing point in #193 

I'll take a look at selection issues too.

manongjohn:
Great!

As for the selection, i think it would be good if worked like control points, where when you hovered close enough, it will expand large enough to indicating you are at that point and make it easier to select.

Issue #181, opened 2020-09-17
Improve tabbing between numerical boxes for ColorField


WolfInABowl:
The ColorField UI element is used in many FX nodes, it allows you to type in values for the `Red`, `Green`, `Blue`, and `Alpha` channels. When pressing Tab to switch between the numerical boxes, you need to press Tab twice as every other Tab press selects a slider. 
As the amount of times a user would purposefully focus a slider via the keyboard is likely low, I believe it would make sense to change the focusing behavior.

PR #186, opened 2020-09-20
Make tabbing through color fields easier.

jeremybullock:
#181 

Now single tab will get through the fields in the color field.


WolfInABowl:
Works great, thanks Jeremy!

PR #188, opened 2020-09-20
Allow drawing behind vector lines on brush and geometry tools.

jeremybullock:
This will allow you to draw behind a vector image using the brush and geometry tools.

Edit:  This has been added as an option to the tool options bar.


melieconiek:
The use of caps lock is typically avoided in software. It's primary function cannot be disabled and often ends up being accidentally on--which can cause a lot of frustration for the user.

I'd recommend creating a set of toggle and direct shortcuts similar to what is available for other tools. This way users can map this however they prefer.

image

jeremybullock:
This now works for the geometry tool too.

jeremybullock:
@manongjohn Do you want to take a look at this?

PR #193, opened 2020-09-22
Grids and Overlays

jeremybullock:
This changes how the field guide button works.

It now toggles grids and overlays.  

Grids and overlays are customizable:
options2

grid

iso


melieconiek:
The thirds and golden ratio grids should move with the camera. They're used to help compose the scene within the camera view.  It would be nice if we had a center cross guide too.

image 
image


It might be helpful to insert a horizontal line in the grid ui to separate camera relative and table relative guides. ...and despite some peoples assertions that 'no one uses the field guide.' Well. I use the field guide. Would it be possible to include it as an option here?

Very nice work dude, thank you. :)

jeremybullock:
@melieconiek Thanks for the feedback.

I fixed the rule of thirds and golden overlays to the camera.  The others still follow the table.

I put the field guide in the options menu.

The options menu and the camera view opacity slider are now accessed through dropdown arrows next to the icon:
pane

The overlay opacity is now adjustable.
The horizontal and vertical grids now have an adjustable offset.
menu

jeremybullock:
Eventually I'd like to add snapping to this, but that will another time.

PR #194, opened 2020-09-22
Fix crash on exploding sub xsheet with only sound.

jeremybullock:


RodneyBaker:
REF in Opentoonz:  https://github.com/opentoonz/opentoonz/issues/3488

PR #197, opened 2020-09-23
Auto close and auto fill options for vector brush

jeremybullock:
This adds the option to automatically close or fill a vector stroke.  If the stroke is filled, it will automatically be self grouped.


jeremybullock:
Frame range should work now too.

melieconiek:
- Enable `Auto Close`, `Auto Fill`, and `Draw Under`.
- Draw a shape, then enter the group for that shape.
- Try to draw another shape with all the same options enabled.

_Causes a crash._ Haven't fiddled with the settings yet to see if it's one of the options in particular.

jeremybullock:
@melieconiek Thanks for the crash info.  I fixed it.

PR #202, opened 2020-09-24
Paste clipboard images (from outside Tahoma) into Tahoma

jeremybullock:
This enables the pasting of outside images into Tahoma.

More to do.

Workflow:
Select the frame you want it pasted to, and paste.
If on the selection tool, you will get the transform tools.

This is still WIP so tweaking to workflow and ease of use is still being done.  Undo currently is borked in a few cases.

### DIFF
--- a/toonz/sources/include/tools/rasterselection.h
+++ b/toonz/sources/include/tools/rasterselection.h
@@ -56,9 +56,8 @@ class DVAPI RasterSelection final : public TSelection {
   bool m_noAntialiasing;
 
 private:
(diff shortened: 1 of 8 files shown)


jeremybullock:
Removed the WIP.

Images pasted to a selected cell will get the option to paste in place or create a new level.  If the image is too big and paste in place is selected, the image will automatically be scaled to fit.

You can also paste from the selection tool and the level strip.  With the selection tool, you get the resize and placement handles.

Issue #158, opened 2020-09-11
Record audio crash


artisteacher:
**Describe the bug**
Clicking on the record button on the "Record Audio" window causes a crash

**To Reproduce**
List steps to reproduce the behavior.
1. Open the "Record Audio" window
2. Click record

**Expected behavior**
Audio to record without a crash.

**Crash Report**
[record_audio.txt](https://github.com/turtletooth/tahoma/files/5208925/record_audio.txt)

**Device Information:**
 - OS: macOS 10.15.6 & 10.14.6
 - Tahoma Version: 1.0.1

jeremybullock:
Does this happen on OpenToonz.  The audio code between the two is identical.  If this doesn't happen in OT, it may be due to a different version of Qt.  Either way, I'll take a look.  The windows and mac audio code is handled differently though.  So it may be a case of "Hey can you test this. . ."

artisteacher:
No crash on OT. But there is a crash on OTX. Could being portable make a difference?

> So it may be a case of "Hey can you test this. . ."

I'm happy to test as needed.

manongjohn:
Might be a QT version issue. OT still builds on an older version I believe. OTX (mac) uses the latest QT.

What version are you building with?

For OTX, I have found there have been issues with more recent QT versions in both macOS and linux versions since 5.13, I think. With Linux I was able to force builds to use QT 5.9.7 like I do with Windows buildd, but I can't do that with macOS builds through Travis. If I could, I would probably use 5.12.

artisteacher:
I've been using 5.15.0.

jeremybullock:
@artisteacher could you test this on OTX and see what happens?

artisteacher:
There's an immediate crash after pressing record.
<img width="835" alt="Screen Shot 2020-10-05 at 1 24 42 PM" src="https://user-images.githubusercontent.com/21159116/95111923-354b5c80-070e-11eb-9a28-40783db03d5c.png">

manongjohn:
Division by 0 error?

manongjohn:
@artisteacher Does Tahoma2D have permission to use your microphone on your mac?  Did it ever prompt you for it? That might explain the crash if it is trying to use something it doesn't have access to.

I saw a report that when someone using QT Creator tried to access the microphone on macOS and it prompted for access, however their installed program wouldn't prompt for access and couldn't record, though they didn't have a crash.

artisteacher:
[...]

PR #228, opened 2020-09-26
Ask to use microphone on macOS

artisteacher:
On later versions of macOS, permission needs to be granted by the user in order for Tahoma to access the microphone. This has no impact on the crash from recording audio #158.

https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos


jeremybullock:
Thanks for this @artisteacher 
I still haven't forgotten about the mic crash on Mac.  I'll try to tackle that this coming week.

PR #240, opened 2020-09-27
Do not remove unused styles 0 and 1 on default palettes

### LINKED ISSUE #231: Empty level allow "Delete Unused Styles" to delete color 0
**Describe the bug**
Empty column allow "Palette > Delete Unused Styles" to delete color 0 (Transparency).

**To Reproduce**
1. Select an empty column
2. On the Palette window, right click "colors" tab and select "Delete Unused Styles"
3. Tahoma ask if you want to delete indices 0, 1

**Expected behavior**
Tahoma shouldn't allow to delete color 0

**Screenshots**
GIF Video
MP4 Video

**Further testing**
This issue doesn't occur on OT since it doesn't allow default palette to show without a level being created.

manongjohn:
This PR fixes #231 

The `Delete Unused Styles` command assumed there was a level and marked off all styles used by the current level, excluding style id 0 only (normal behavior since day 1). For Default Palettes, there is no level so it didn't exclude style id 0 from the list and allowed it to be deleted.

Modified behavior such that 
- If it is a default palette, style id 0 and 1 are immediately excluded.
- Enhanced logic to also exclude style id 1 on normal level palettes in keeping with the inability to directly delete style id 1 normally.

### DIFF
--- a/toonz/sources/toonzqt/styleselection.cpp
+++ b/toonz/sources/toonzqt/styleselection.cpp
@@ -703,6 +703,11 @@ void TStyleSelection::eraseUnusedStyle() {
   // Verifico quali stili sono usati e quali no
   std::map<int, bool> usedStyleIds;
   int pageCount = palette->getPageCount();
+  // We're on a default palette.  Never delete Style 0 and 1
+  if (palette->isDefaultPalette()) {
+    usedStyleIds[0] = true;
+    usedStyleIds[1] = true;
+  }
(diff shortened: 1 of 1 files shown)


jeremybullock:
Thanks!  That was fast and I love the addition of style 1 to the protected styles.

justburner:
Well done, didn't expected a day zero fix :D

PR #248, opened 2020-09-28
More paste improvements

jeremybullock:
Pasting an image with the selection tool enabled will now give the transformation box around the paste even without an active selection. 

This also takes care of a few paste crash issues.

### DIFF
--- a/toonz/sources/common/tvrender/tpalette.cpp
+++ b/toonz/sources/common/tvrender/tpalette.cpp
@@ -278,7 +278,7 @@ TPalette::~TPalette() {
 
 TPalette *TPalette::clone() const {
   TPalette *palette = new TPalette;
-  palette->assign(this);
+  if (this && this->m_styles.size() > 0) palette->assign(this);
   return palette;
 }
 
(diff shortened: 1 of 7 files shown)


artisteacher:
Can vector data be pasted?

jeremybullock:
It can be pasted, but I'm working on making the selection handles appear right now.

jeremybullock:
Alright.  I added the vectors to the mix too.  So now any drawing data that is pasted while the selection tool is active should paste with the transform tools around it.

There are a couple other bug fixes I found as I worked on this.

@manongjohn There are currently 3 different locations that handle pasting drawing data.  I could make it so that any time drawing data is pasted, the selection tool is activated and the data is pasted via the selection tool.  This would give the benefit of always getting the selection transforms when you paste, so that users don't wonder why they sometimes get the controls and other times not.

It would also let us remove the paste commands from the cellsection and filmstripcommands.  This would simplify the paste code significantly.  Part of why this was a pain is I had to make sure that traffic was being sent to the correct paste commands depending on what tool was active and what type of drawing data we were dealing with.  3 level types is a blessing and a curse.

What are your thoughts on simplifying the paste commands into 1 location?

jeremybullock:
My suggestion above for consolidating the paste code won't work because we can paste on blank cells.  We would lose that ability if I got rid of the extra paste code.  Bummer.

I'd prefer to get a bit leaner on the code side.

PR #253, opened 2020-09-28
Horizon overlay and some vanishing point customization

jeremybullock:
This adds a horizon option in the overlays, with lines getting closer together as they approach the horizon line.  The position of the horizon, the angle and the spacing between lines is configurable.

This also moves the toggle for the vanishing points to the overlay menu.  You can now customize how many degrees between each ray and the opacity of the vanishing points separately from other overlays.


jeremybullock:
This is probably the last new feature that will be added to 1.1.  After this (unless @manongjohn has anything up his sleeve) I think a feature freeze is in order.  I will then try and knock out a bunch of bugs before 1.1 is released.

artisteacher:
Some people might find it confusing that the Grid/Overlay button turns visibility of all grids/overlays on/off except for the vanishing point rays.

On a related note, enabling a grid option should automatically toggle on the Grid/Overlay button if it is off.

PR #267, opened 2020-10-01
Change glut to freeglut on Windows

jeremybullock:
The original glut hasn't been updated in 20 years.  
https://www.opengl.org/resources/libraries/glut/glut_downloads.php

This uses the still updated and newer freeglut.

I don't plan on merging this until after 1.1 is released.

Comments are welcome.


jeremybullock:
Side note:
It looks like glut is only used to draw text to the viewers and opengl surfaces (poorly).  If we had an alternate way to draw text to those surfaces, it doesn't look like we would need glut.

RodneyBaker:
Very interesting observations!

PR #268, opened 2020-10-01
Remove libusb

jeremybullock:
libusb was used by the scanner support which was removed from T2D.


jeremybullock:
Again, not going to mess with merging until after 1.1

RodneyBaker:
It seems that scanning in general has been deprecated from Opentoonz.
I'm not sure what other options there are for use of scanners outside of GTS.
This has been mentioned several time with the idea presented to remove the scanning option from the menu as they do not appear to be used by anyone to include Japanese studios.

My own preference would be to see direct scanner support for Opentoonz and Tahoma2d but when last I checked those developing the TWAIN standard itself had stalled out on their efforts.  I'll check again and report back any pertinent observations.

manongjohn:
Need to remove libusb from the Linux install scripts also.

jeremybullock:
Will do.  Thanks!

PR #275, opened 2020-10-02
Make change parent dropdown visible.

### LINKED ISSUE #272: Change parent from the xsheet dropdown is hidden.

jeremybullock:
This fixes #272 

It also keeps children of a column from showing up in the dropdown.  The circular parenting was already not allowed in the parent logic, this just keeps it from being shown as a choice in the ui.


artisteacher:
@jeremybullock - There's an unfortunate offset on macOS that wasn't there before. Also, I think if there's no port dropdown, the target area & resulting dropdown should take up the full column.

Screen Shot 2020-10-03 at 10 56 53 AM

jeremybullock:
Hmm.  I'm seeing that now too.  I don't know what made it different for a bit.  Reverting the change, and it's working again as expected.

jeremybullock:
The dropdown is variable width and will resize to fit the column names, but I can make the minimum width bigger.  I'll see about making the target area bigger if the parent is a table, camera or peg.

PR #282, opened 2020-10-02
Fix missing shortcuts

### LINKED ISSUE #274: Include Save (Project/Global) Palette as a keyboard shortcut
Hi.  I'm spending a lot of time adding to and making adjustments to my project palette (which I have all my levels using and linked to)... and to see the changes, I need to save the project palette via the menu at the bottom right of the Global Palettes window.

A great time saver would be to be able to set that save as a keyboard shortcut, so I can frequently update the colours in the project without having to use the mouse and the menu pulldown.

I see there is a Save Palette item in the shortcuts list, but it appears it might be the local (level) palette(s), rather than the global one... 

Having it as a shortcut would definitely help speed up the colouring process for me.

jeremybullock:
When I added the vector auto close and draw under options, I didn't create the ability to assign them to a shortcut.  This fixes that.

I also added a shortcut for saving the studio palette.
closes #274


melieconiek:
Thank you for this!

PR #294, opened 2020-10-04
UI Updates

jeremybullock:
This fixes some inconsistencies in Tahoma

The Windows menu renamed to Panels
You can now pan the function graph with space
Double click an fx name to rename, double click outside of the name to open settings
Changed XSheet to Xsheet in the UI - capitalization difference.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -282,7 +282,7 @@
         <command>MI_MaximizePanel</command>
         <command>MI_FullScreenWindow</command>
     </menu>
-    <menu title="Windows">
+    <menu title="Panels">
         <command>MI_OpenCommandToolbar</command>
         <command>MI_OpenToolbar</command>
         <command>MI_OpenToolOptionBar</command>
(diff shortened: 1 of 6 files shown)


artisteacher:
Would you like the manual updated to use panel instead of window (floating) and pane (docked)?

I’ve also been wondering if perhaps the Xsheet menu might be renamed to Scene? Now that the default orientation is Timeline, a menu name that’s less orientation specific might be beneficial.

jeremybullock:
Updating the manual would be great @artisteacher 

I also agree about changing xsheet to something like scene, since the xsheet is only one way of visualizing the scene now.

jeremybullock:
Question. . . Do subXsheets then become subscenes?

artisteacher:
They could.

jeremybullock:
Just asking because I think it would be good to be consistent.  But I would like feedback before I proceed.
I do think it's good to changes like this before Tahoma gets too big.

artisteacher:
Panel should be all set in the manual, at least for English.

It's definitely best to make these sorts of changes now.

jeremybullock:
Thanks!  The xsheet menu and other non-actual-xsheet references are changing to scene or subscene tonight.

PR #300, opened 2020-10-05
WIP:  Fix Mac audio crash. . .

jeremybullock:
@artisteacher Could you try this and see if it make any difference in recording audio.


artisteacher:
Still getting the same crash. In case it helps, here's what shows up in Console when opening the record audio window up till pressing record & getting the crash:
Screen Shot 2020-10-05 at 9 15 34 PM

jeremybullock:
Bummer. . . I'm not sure what's going on.

PR #301, opened 2020-10-05
Fix resizing windows on start

### LINKED ISSUE #192: Some windows shrink on open in active room
**Describe the bug**
On macOS, when Tahoma opens, the Timeline, Style Editor, and Stop Motion Controls windows shrink if they are in the active room upon opening. It's possible that other windows could have the same issue, but these are the only ones that are a problem in the default setup. Rooms that were inactive on opening keep their sizes as expected.

**To Reproduce**
1. Open Tahoma after install - the 2D room's Timeline is shorter than the other rooms & the Style Editor is narrower than expected
2. Adjust window sizes to fit the screen better - or switch to the Stopmotion room
3. Close Tahoma
4. Reopen Tahoma - changes to the 2D room are reset if it is active on open; if the stopmotion room is active, the Stop motion controls & Timeline are barely visible.

**Expected behavior**
The widths/heights of windows should not change on opening just because they are in the active room.

**Screenshots**

2D room after adjustments:
Screen Shot 2020-09-21 at 6 11 49 PM

2D room on install & after reopening if active:
Screen Shot 2020-09-21 at 6 09 55 PM

Stopmotion room if not active on open:
Screen Shot 2020-09-21 at 6 10 26 PM

Stopmotion room if active on open:
Screen Shot 2020-09-21 at 6 13 53 PM

**Device Information:**
 - OS: [macOS 10.14.6 & 10.15.6]
 - Tahoma Version: [master-9-21-20]

jeremybullock:
fixes #192 

I had made a change earlier that would not lock in a windows fixed width if it was being docked, to make docking windows from a floating state place the windows a little more naturally, but on mac that was causing issues with window sizes on restarting.

I reverted the change.

### DIFF
--- a/toonz/sources/toonzqt/docklayout.cpp
+++ b/toonz/sources/toonzqt/docklayout.cpp
@@ -422,14 +422,9 @@ void DockLayout::applyTransform(const QTransform &transform) {
 
 //------------------------------------------------------
 // check if the region will be with fixed width
-bool Region::checkWidgetsToBeFixedWidth(std::vector<QWidget *> &widgets,
-                                        bool &fromDocking) {
+bool Region::checkWidgetsToBeFixedWidth(std::vector<QWidget *> &widgets) {
   if (m_item) {
-    if (m_item->wasFloating()) {
-      fromDocking = true;
-      m_item->clearWasFloating();
-      return false;
-    }
+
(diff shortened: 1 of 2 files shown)


artisteacher:
So much better, thanks!

PR #302, opened 2020-10-05
WIP: Add webcam resolutions on linux

jeremybullock:
This is the start of getting webcam resolutions working on linux.  So far it's a simple attempt.

@manongjohn When you get the chance can you see if this at least gets a couple resolutions working on linux?  I plan on adding more once we confirm this solution is working.


manongjohn:
I tested this, though the max resolution for my laptop's webcam is 1280x720 so I couldn't test switching to a different resolution because there wasn't one lower and can't go any higher.  The image was at least bigger than when I didn't have any resolution to choose from.

This is what Windows gives me for my integrated webcam: 
image

Not sure what would be considered a standard set and how useful the lower resolutions would  be for Stop Motion work, but I can add them to  my list just to test switching resolutions.

jeremybullock:
I don't think anything below 640 x 360 is really worth messing with.

jeremybullock:
Thanks for testing and confirming.  Glad to get this working on Linux.

jeremybullock:
I added the resolutions you listed as well as the resolutions listed by my webcam.  It may be overkill, but until we can poll the supported resolutions on linux, this will have to do.

PR #303, opened 2020-10-06
Change Xsheet references to Scene (Unless it's actually an Xsheet)

jeremybullock:
Unless something is actually an xsheet, it is now referenced as a scene.
Examples: 
In the fx schematic, the xsheet node is now scene
Sub-xsheets are now sub-scenes
The xsheet toolbar is now the quick toolbar.

This is purely a cosmetic change.  It only changes visible references to xsheet.  Internally, all the variables and classes are still the same.  The only exception is the quick toolbar.


jeremybullock:
@manongjohn Do you want to look at this beast?  I don't blame you if the answer is no.

manongjohn:
Well, this certainly looked fun to do. lol

I've reviewed the changes and I think you got everything that needed to be changed.

PR #308, opened 2020-10-07
Build script corrections

manongjohn:
This PR does the following with the macOS/linux build scripts

1. Fixes macOS build script issues and suggestions brought up by @artisteacher  (thanks!)
2. Fix re-running the buikdpkg scripts to only create certain things when needed or remove certain things before redeploying.
3. After building the shared ffmpeg library needed for opencv compiling, it builds a static ffmpeg to be used for deployments.

### DIFF
--- a/ci-scripts/linux/tahoma-build.sh
+++ b/ci-scripts/linux/tahoma-build.sh
@@ -5,7 +5,10 @@ popd
 
 cd toonz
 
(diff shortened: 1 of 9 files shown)


manongjohn:
I cancelled Travis build as it's not necessary, but currently do not have access to stop Appveyor build.

PR #314, opened 2020-10-07
Fix Plastic Rendering on 16 bit

jeremybullock:
This fixes the rendering issues that plastic rendering had when using 16 bit depth.


jeremybullock:
Note: This doesn't process the plastic portion of the image with 16bits.  That is still handled with 8 bits, but it converts the image data back to 16bit data to avoid the render issues.

Some of the code here will be able to be simplified once we switch to a newer Qt version.

PR #324, opened 2020-10-08
Make setting markers easier

### LINKED ISSUE #321: Keyboard Shortcut for "Set Start Marker" / "Set End Marker"
In other software for video editing you can set a keyboard shortcut for In / Out loop points on the timeline, so far I can't seem to find the proper term / feature to set these to keyboard shortcuts from within Tahoma.

Being able to find / set "Set Start Marker" / "Set End Marker" in the keyboard shortcuts settings menu. This would allow me to setup my graphics tablet / external control buttons for faster timeline editing between a loop area, as I'd be able to expand / shrink the zones with a fast button press vs a few mouse clicks.

image

image

image

image

jeremybullock:
closes #321 

This makes it possible to assign as shortcut to set and remove markers and auto markers.

It also allows you to:
Ctrl + Click in the frame number area to set the in marker
Alt + Click in the frame number area to set the out marker
and Ctrl + Alt + Click in the frame number area to clear the markers.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -2094,6 +2094,12 @@ void MainWindow::defineActions() {
                                    tr("Remove Multiple Keys"), "");
   menuAct->setIcon(createQIcon("remove_multiple_keys"));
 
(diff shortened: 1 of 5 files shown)


melieconiek:
Thank you for this, very nice. :)

PR #327, opened 2020-10-09
Fix colors on various checks.

jeremybullock:
This makes onion skin colors still show for Toonz Raster levels with Transparency Check or Fill Check on.

It also makes the Gap Check color on Toonz Raster levels still show of the above options are on.


jeremybullock:
Also fixes opentoonz/opentoonz#2504

PR #338, opened 2020-10-10
Fill with gaps on Smart Raster levels.

jeremybullock:
morefill

This allows smart raster drawings to be filled even if there are gaps.  The gap tolerance will be configurable in the tool options bar along with the option to fully ignore gaps.

The tool options part of this is yet to be implemented, but it works and (for testers)you can configure the gap size with the tape tool.


jeremybullock:
This is ready for testing.  There are three options for dealing with gaps: ignoring them (old behavior), filling the area only, or closing the gap with a line and filling the area.

jrhyme:
The function does not work for me, as soon as the radius is not set, the background is filled......

jrhyme:
The error remains ..

jeremybullock:
Are you using vector levels or smart raster levels?

jrhyme:
> Are you using vector levels or smart raster levels?

 smart raster levels

jeremybullock:
@RodneyBaker You tried this earlier.  Any issues with it?

jrhyme:
> @RodneyBaker You tried this earlier. Any issues with it?

there were problems then and now

melieconiek:
Seems to work alright on my end. No more lag when drawing with gap check on either, which is nice.

It might be nice if gap check updated as you were pulling the slider, though.

Very nice dude!

jeremybullock:
Thanks for the confirmation @melieconiek 
I agree on gap check updating as you go.  I'll see what I can do.

RodneyBaker:
I didn't experience any issues.  Works quite well.

@jrhyme If you experienced problems you should detail those so they can be addressed/fixed.

toonz/sources/toonzlib/fill.cpp
```diff
@@ -374,18 +384,33 @@ bool fill(const TRasterCM32P &r, const FillParameters &params,
   int paint = params.m_styleId;
   int fillDepth =
       params.m_shiftFill ? params.m_maxFillDepth : params.m_minFillDepth;
+  TRasterCM32P tempRaster;
+  int styleIndex                                 = 4094;
+  int fakeStyleIndex                             = 4095;
+  if (autoCloseDistance < 0.0) autoCloseDistance = AutocloseDistance;
+  if (fillGaps) {
+    tempRaster            = r->clone();
+    TPalette *tempPalette = params.m_palette->clone();
```
shun-iwasawa, line 393: This line seems to be unnecessary since `tempPalette` is not used in this scope.
I found that OT crashes here as OT estimates `params.m_palette` should not be NULL in `TPalette::clone()` .

jeremybullock: Totally right.  That line was left over from a previous idea.  I'll take it out.

artisteacher:
This works great on macOS. You might want to hide the gap options when the mode is set to lines, since it has no impact. 

I have gotten interesting results if I use "fill gaps" with a high gap distance & then try to fill the area again with either "ignore gaps" or with the gap distance set down to 1. 
Screen Shot 2020-10-16 at 4 52 11 PM
Unsurprisingly, using "close & fill" is more reliable if refilling with other settings later.

jeremybullock:
I like the idea of hiding it when lines are selected.  I'll do that.

jeremybullock:
I think this is ready.

jeremybullock:
@manongjohn If you are interested, this is ready for review.

jeremybullock:
I'm going to merge this so I can start bringing in some of the recent OT updates.

PR #342, opened 2020-10-10
Fix running through Linux shortcut

### LINKED ISSUE #340: Having problem on launching
I'm using the appimage on linux mint. I added the shortcut on my menu but when launching it from there an error pos up. It's launching normely if I Double click on the Appimage file but I want it to launc from the menu for daily use.
Screenshot_20201010_155754 Though I'm not sure if it's a bug or what but an issue for me.
Thank U so much for making Tahoma and I wish you can help me with my problem

manongjohn:
This PR fixes #340 

For Linux systems, when launching from a shortcut (symbolic link), the working directory is where the application was started from so Tahoma2D cannot find the tahomastuff directory.

Added logic for Linux systems only, to attempt to convert a symbolic link to the true path of the application and set the working directory to that.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -220,7 +220,17 @@ public:
   std::string getSystemVarPrefix() { return m_systemVarPrefix; }
 
   void setWorkingDirectory() {
-    QString workingDirectoryTmp  = QDir::currentPath();
+    QString workingDirectoryTmp = QDir::currentPath();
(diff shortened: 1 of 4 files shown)


manongjohn:
Just realized I need to fix the ffmpeg detection

jeremybullock:
thanks for the quick fix.

manongjohn:
This should be good to go now.

jeremybullock:
Awesome.  Thanks,

PR #360, opened 2020-10-15
Allow render without saving.

jeremybullock:
This has been requested a bunch, so this will be going into 1.1.


RodneyBaker:
Thanks Jeremy!
As the primary instigator of the Save before Rendering option I am also a big fan of having the option to bypass that as well.
This will do just that.

For me personally this option will be used mostly for smaller sketchings and doodles that I don't really have any need to save the scene or level for and the rendered image is sufficient.

Much appreciated.

artisteacher:
On macOS I don't see a menu option for Rendering without saving. It does show up in the Output Settings.

PR #361, opened 2020-10-15
WIP - Use the schematic hierarchy for ordering drawings in the viewer.

jeremybullock:
This is the first step toward bringing in live matte fx.  

This gives an option in a drop down next to the camstand and camera view buttons to draw the viewer based on the hierarchy of the fx schematic.

This will allow fx like over fx to change the drawing order shown in the viewer.

Eventually this will also allow matte in/out nodes to do live matting.

This isn't even test worthy right now, but I am making this PR to get comments and feedback about the desired functionality of this.

It does currently build the viewer based on the schematic, but does not take into account any actual fx.  But if a node is disconnected from the schematic, it will not show with this option on.

This will not become the new default and will always be an option.  Once it starts working with more fx, there could be a performance hit, so I want the current behavior to always be available.

### DIFF
--- a/toonz/sources/common/tgl/tstencilcontrol.cpp
+++ b/toonz/sources/common/tgl/tstencilcontrol.cpp
@@ -285,10 +285,9 @@ void TStencilControl::Imp::popMask() {
 //---------------------------------------------------------
 //---------------------------------------------------------
 
-// questo forse e' un po' brutto:
-// La maschera e' chiusa.
-// Se e' abilitata,    open = push+open
-// Se e' disabilitata, open =      open
+// this is perhaps a bit ugly : the mask is closed.
+// If it's enabled,    open = push+open
+// If it's disabled, open =      open
 
 void TStencilControl::beginMask(DrawMode drawMode) {
   glPushAttrib(GL_ALL_ATTRIB_BITS);
--- a/toonz/sources/common/tgl/tstencilcontrol.cpp
+++ b/toonz/sources/common/tgl/tstencilcontrol.cpp
@@ -43,7 +43,7 @@ public:
   int m_stencilBitCount;
   int m_pushCount;
   int m_currentWriting;  // current stencil bit plane.
-  // 0 is the first bit plane ; -1 menas no stencil mask is writing
+  // 0 is the first bit plane ; -1 means no stencil mask is writing
 
   int m_virtualState;
 // the state of the (eventually virtual) top mask.
@@ -53,7 +53,7 @@ public:
 #ifdef _DEBUG
   std::stack<bool> fullState;
 // state of each mask in the stack (except top mask).
-// 'true' means opend; 'false' means close and enabled
+// 'true' means opened; 'false' means closed and enabled
 // Used only for assert
 #endif
 
--- a/toonz/sources/include/toonz/stage.h
+++ b/toonz/sources/include/toonz/stage.h
@@ -63,7 +63,7 @@ DVAPI void visit(Visitor &visitor, ToonzScene *scene, TXsheet *xsh, int row);
 
 //-----------------------------------------------------------------------------
 
-DVAPI void visitSimpleLevel(Visitor &visitor, TXshSimpleLevel *level,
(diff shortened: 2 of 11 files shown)


jeremybullock:
@shun-iwasawa This is another PR that I thought you might like in OT.  If you have any thoughts about getting matte in/out to preview for toonz raster or raster levels, I'd love to hear it.  There is already masking for vector levels which works well using the OpenGL stencil buffer.

shun-iwasawa:
Thank you for letting me know. Wow, it seems to be very ambitious endeavor.
Applying the mask in easy & handy way would be very welcome.

I'm not yet observing the source very much, but am a little concerned that:
The Fx Schematic is not a tree structure, but actually a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Which means that one column can be connected to multiple ports of the layering fxs. In such case the viewer may need to draw the same image multiple times?

jeremybullock:
@shun-iwasawa Thanks for the response.  I'm aware of the potential of images being drawn multiple times.  If I continue with this PR, I plan on making a vector of images that have been called to be drawn and keep them from being drawn again.  I'm not sure yet if I plan on continuing this PR, but I do think it would be nice to be able to show images in the render order and to have masking available without a full render.  

Still not sure the best way to go. . .

jeremybullock:
I'm going to close this.  I think this will be too error prone and hard to maintain.

artisteacher:
Too bad!

manongjohn:
My wish would be to get the mask column fully working on the timeline and rendering out appropriately. Would also like if we could somehow make it work with any column instead of vector columns only. 😁

jeremybullock:
That's why I started with this - to get masking working.  I think the mask columns made sense in the versions of Toonz that didn't have the schematic, but with that, having masking columns and matte effects could get confusing pretty quickly.  That's why I started going the route of trying to pull the schematic hierarchy into the viewer and was then going to do masking from there.

I would still love to be able to visualize mask/matte without preview turned on, but I'm not sure how to go about it.

adamearle:
> That's why I started with this - to get masking working. I think the mask columns made sense in the versions of Toonz that didn't have the schematic, but with that, having masking columns and matte effects could get confusing pretty quickly. That's why I started going the route of trying to pull the schematic hierarchy into the viewer and was then going to do masking from there.
> 
> I would still love to be able to visualize mask/matte without preview turned on, but I'm not sure how to go about it.

I would love to know what was preventing you from implementing this. After reading what you have said. It seems that it was more about how to implement it. Not so much that it couldn't be implemented?

do you think we could take another stab at it.

I would love to sit down with you and try to work it out if we can. https://github.com/opentoonz/opentoonz/issues/4329

manongjohn:
Doubt Jeremy will answer you as he's left github and the project...but I will...

I actually independently started working on column based masking some time ago and pretty much came to similar conclusions and thoughts about it.

The current column masking works 1 way, but the node masking does it a completely different way.  Combining both masking would be very confusing...things to consider:

- Current column masking works only with Vectors as the mask and only masks the column directly below/left in the stacking order.  Node masking uses any level type, which is nice.
- Column masking is not applied to final render.  I'm guessing the code was taken out before OT went opensource.
- The node render logic is vastly different than how live column masking works...may because it's Vector based mask? Regardless, how live view and preview works is quite different which is probably for performance reasons.
- Render logic follows the nodes to build images. If it encounters a column mask, how/when does it apply the column mask, and what happens if the column it masks had other Fx on it? what's the impact of fx on a column used as a mask. (these are probably minor issues)
- When I compared the current column masking results to the node masking...there was a difference.  In one case the mask including the line work where as the other, it didn't (forget which one did which).

It would be nice to have live masking, but it would likely need to be rewritten and should...
- be able to use any level type as a mask
- be able to apply the mask in such a way as not to be physically required to be next to each other (perhaps like plastic meshes?)
- be able to apply column mask the same way it does in node mask, though may need to do so in such a way as not to cause performance issues.
- somehow handle applying column masking and node based masking appropriately.

adamearle:
@manongjohn Thanks for answering mate. I had no idea that was Jeremy.

PR #368, opened 2020-10-16
Add link to Google Group

jeremybullock:


artisteacher:
@jeremybullock - on a related note, what do you think about having the "What's New" link from the Help menu go to the "What's New" page of the manual instead of GitHub?

jeremybullock:
Sounds good @artisteacher

PR #370, opened 2020-10-18
Correct mac shortcuts and missing menu item

artisteacher:
For Tahoma on macOS, we use Cmd instead of Ctrl and Opt instead of Alt. This fixes the display of those shortcuts in the status bar on macOS.


artisteacher:
#360 added an additional render option into the menu but missed the macOS menu. This also adds the additional render option in the Render menu for macs.

jeremybullock:
Thanks for the fix @artisteacher

PR #374, opened 2020-10-18
Block fixed with panels on dock and undock

### LINKED ISSUE #325: Undocking Scene Viewer doesn't cause Palette and Style Editor to expand Horiztonally
**Describe the bug**
Normally when undocking things adjacent panels surrounding it will resize to fill the space. Normally if no adjacent panel can be resized, the panel cannot be undocked.

In the 2D room, you can undock the Scene Viewer but the Palette and Style Editor doesn't expand horizontally like it's supposed to, leaving a hole in the layout.

When this happens, you cannot undock anything else and if you try to resize anything vertically, Tahoma2D crashes.

NOTE: This happens in OT also

**To Reproduce**
1. Open Tahoma2D to 2D room
2. Unlock the panels
3. Undock the Scene Viewer   -> Palette and Style Editor do not expand to fill the space and it should
4. Try to undock something else. -> You can't
5. Try to raise/lower the timeline -> Crash

**Screenshots**
image

jeremybullock:
fixes #325 
This prevents panels from being fixed width if an item is docked or undocked.

@artisteacher 
Would you mind testing this on Mac to see if it breaks anything?  It's a pretty straight-forward fix, so it should be fine, but I'd like to make sure.

### DIFF
--- a/toonz/sources/toonzqt/docklayout.cpp
+++ b/toonz/sources/toonzqt/docklayout.cpp
@@ -424,7 +424,6 @@ void DockLayout::applyTransform(const QTransform &transform) {
 // check if the region will be with fixed width
 bool Region::checkWidgetsToBeFixedWidth(std::vector<QWidget *> &widgets) {
   if (m_item) {
-
     if ((m_item->objectName() == "FilmStrip" && m_item->getCanFixWidth()) ||
         m_item->objectName() == "StyleEditor" ||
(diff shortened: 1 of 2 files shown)


artisteacher:
Seems fine to me.

jeremybullock:
Thanks for checking @artisteacher

PR #376, opened 2020-10-18
Add individual reset room layout options

manongjohn:
For those times when a user decides to play with their room layouts or even add new rooms, this PR adds the following:

1. Individual Room Layout resets

image

- The Saved layout option reverts to the last time the room layout was saved (when quitting)
- The Default layout option reverts to the original installation version. It is only active for rooms that are part of the original installation.

2. Changes `Reset to Default Rooms` to `Reset All Default rooms`

image

- Only affects original rooms. Custom rooms (rooms added by user) are not touched.
- This will restore deleted original rooms to the end of the tab list.

In both cases, the room layout resets are immediate and does not require restarting Tahoma2D.


jeremybullock:
Dude!  This is awesome!  I have long wanted the ability to reset rooms without restarting.

jeremybullock:
Just tested and man! This works really nice.
I did get one crash on updateToolbar in toolbar.cpp, but couldn't duplicate it.  I'll keep poking, but overall, I love this.

manongjohn:
Thanks for checking it out.  

Initially I had some crashes when the toolbar was being updated.  Didn't happen for every tool and largely depended on what controls were displayed.  For a while I was able to crash reliably but then I changed how I was removing the pane. I thought I fixed since I couldn't reliably recreate it anymore.  If you manage to get a reliable way to produce it again, let me know.

jeremybullock:
I was looking at this today.  I am still getting some toolbar crashes, but haven't yet nailed down a series of steps.  
I did find a reliable crash-
From a clean room set-
Unlock the rooms
Delete the timing room
Reset all rooms.

jeremybullock:
Actually that just created an assertion failure, but after ignoring the assertion, I deleted the 2D room and that triggered a crash.

jeremybullock:
K, I got a reliable toolbar crash.  

With a clean set of rooms, 
go to the stop motion room
delete the timing room
delete the 2d room
reset all rooms.

I'm still really hopeful for this one.  It would be awesome.

manongjohn:
Ok.  I was able to duplicate with the artifact.  I'll look further into it.

jeremybullock:
Just tested and I can confirm the crashes are fixed.  Thanks @manongjohn

Issue #184, opened 2020-09-18
Menu disappears after deleting new room on macOS


artisteacher:
**Describe the bug**
The menu disappears after deleting a room. Upon restarting the menu is back. This happens on OT as well. 

I think it's because on Windows, each room has(or had?) it's own unique menu. In that case, the room's menu should be deleted with the room. On macOS however, there is only one menu that all the rooms share. It should never be deleted.

**To Reproduce**
List steps to reproduce the behavior.
1. Unlock the workspace
2. Create a new room & name it
3. Add a window to the room
4. Switch to a different room
5. Delete the new room

**Expected behavior**

**Screenshots**
Before deleting a room:
![Screen Shot 2020-09-18 at 9 23 40 AM](https://user-images.githubusercontent.com/21159116/93602552-c1c8f180-f990-11ea-9a29-bbc9fbf1e795.png)
After deleting a room:
![Screen Shot 2020-09-18 at 9 23 19 AM](https://user-images.githubusercontent.com/21159116/93602573-cab9c300-f990-11ea-94b8-12494433bad6.png)

**Device Information:**
 - OS: macOS 10.14 & 10.15
 - Tahoma Version: Build from master on 9/11/20

artisteacher:
Creating a new room (after deleting a new room) also brings the menu back.

jeremybullock:
Oh, that's fun.

artisteacher:
That's one way to describe it!

PR #377, opened 2020-10-18
Remove creation and use of room_menubar.xml

manongjohn:
This PR removes the creation and use of room_menubar.ini files. It primarily impacts Windows users.

Issue:  Tahoma2D currently does not allow menu customizations, however it still creates and uses room_menubar.ini files.  Depending on how a user upgrades to a newer version, this becomes a problem because the room_menubar.ini files are not updated and will hide any menu changes we've made.  Users would need to reset rooms in order to get the changes which also has the bad side effect of resetting the room layouts and deleting custom rooms.


jeremybullock:
Awesome!

jeremybullock:
I'd actually be fine with losing the whole feature instead of ifdef'ing it out.  I don't see bringing it back.

jeremybullock:
Should this go into 1.1?

artisteacher:
Yes! Should it have any impact on #184?

manongjohn:
@jeremybullock I honestly can't imaging us bringing it back either.  I was half tempted to remove everything related to it. Would take me a bit to do.

@artisteacher I just tested it and no, it does not fix that unfortunately.  I think what is happening is when it deletes the room, it blindly deletes the menubar. For macOS, these should only be 1 so it's deleting it.

Surprisingly, despite my change, this doesn't happen on Windows and I would expect it to since it's supposed to be just 1 menubar now. This is prompting me to look a little more into how the menubar is being handled.  I'll try and fix it with this PR.

manongjohn:
I've removed the logic that managed 1 menubar per room, even though they were the same. I've also removed related to customizing them.  Now there is only 1 menubar for the entire application. 

This change will fix #184.

jeremybullock:
Awesome work.  LGTM.

PR #378, opened 2020-10-18
Remove Camera Capture Popup Window

### LINKED ISSUE #322: Camera Capture crash on open
**Describe the bug**
Immediate crash on opening the camera capture window.

Do we still need camera capture since we have the stop motion controls working for all supported OS?

**Screenshots**
Screen Shot 2020-10-08 at 1 38 28 PM

**Device Information:**
 - OS: macOS 10.14.6 & 10.15.6
 - Tahoma2D Version: Alpha 2

jeremybullock:
Closes #322

I moved  the histogram and color modes to the webcam part of stop motion, then removed the camera capture popup.


artisteacher:
The build fails on macOS - just like in the travis build.

jeremybullock:
Do you have errors?

artisteacher:
Screen Shot 2020-10-18 at 10 35 46 PM

jeremybullock:
Thanks.  I think I took care of it.

artisteacher:
It works! The histogram & color modes are working great. I'm not seeing any obvious impact from the other settings on the settings tab when using a built-in camera.

jeremybullock:
The function of the settings on the settings tab varies based on the camera vendor's decision to support them.  I have some cameras that work with some settings but not others.  I don't think there is a way to check if a setting is supported besides testing it out.  I'll keep it in mind as something to look out for in the future, but I'm aware that for some cameras, not all settings will work.

PR #389, opened 2020-10-20
Remove platform specific audio code and use Qt.  Plus other audio stuff.

jeremybullock:
This removes the Windows specific audio code and unifies all platforms to use the Qt audio stuff. The windows code added to the complexity of the code base without any real benefit.

There are some things that also come with this PR:
Most(?) audio distortion on Windows should be significantly better.
Scrubbing is more responsive and seems to keep up better.
The play audio button now resets when audio is finished playing on Windows (This already worked on other platforms)
The memory leak from audio levels is now fixed.


artisteacher:
Will this help with https://github.com/opentoonz/opentoonz/issues/1739?

melieconiek:
@artisteacher An update of the scene audio is not triggered after some actions where audio exposures are edited. If the bug you described there is caused by this, then no this PR doesn't seem to address that.

melieconiek:
I'll fuss with this and see if the resolves the memory leak issue. Thanks, @jeremybullock. :)

melieconiek:
Been working with audio today, doing some lipsyncing--I loaded in this PR. Audio playback no longer lags behind, but audio quality, especially when scrubbing is much worse. Compared side-by-side, pops and clicks happen much more often. It's bad enough that it makes it a little difficult to discern phonemes.

jeremybullock:
@melieconiek Thanks for the feedback.  I think I will go back to the original sound.  But this does fix ( to a large degree ) the memory leak.  It is in how Tahoma handles audio frames.  There may be a little residual, but it should be significantly better.  The good part is that the fix wasn't related to which audio code we were using.

PR #394, opened 2020-10-20
Centralize user preference and configuration files

manongjohn:
This PR moves user configuration/preference files saved in various places under 1 directory.  Keeping everything centrally located will help with upgrading and potentially sharing in the future.

Files will be saved under `tahomastuff\profiles\users\xxx`, where `xxx` is the user's username logging into the machine.

Here are the following files that will be relocated:

What|Old location|New Location
----|------------|------------
Preferences, config files and default palettes|profiles\layouts\settings.xxx\\*|profiles\users\xxx\\*
Application Variables|profiles\env\xxx.env|profiles\users\xxx\env.ini
Room Layouts|profiles\layouts\personal\Default.xxx\\*|profiles\users\xxx\layouts\Default\\*
Browser Room's history|config\xxx_history.txt|profiles\users\xxx\file_history.txt
Brush presets|config\brush_*.txt|profiles\users\xxx\brush_*.txt
Camera output presets|config\outputpresets\*|profiles\users\xxx\outputpresets\*
Resolution presets|config\reslist.txt|profiles\users\xxx\reslist.txt
Cleanup resolution presets|config\cleanupreslist.txt|profiles\users\xxx\cleanupreslist.txt

It should be noted that presets currently saved in the `config` directory by 1 user are available to all users on the machine because the filename does not indicate which user it is for. With this change, they become personal settings and are no longer shared by default.

A future enhancement for shared machines or networked installations would be to create a `profiles\users\template` folder to be used when creating new users. It would contain initial configurations/layouts set up by an admin user.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -158,8 +158,8 @@ public:
         getSystemVarPathValue(getSystemVarPrefix() + "PROFILES");
     if (profilesDir == TFilePath())
       profilesDir = getStuffDir() + systemPathMap.at("PROFILES");
-    m_envFile =
-        profilesDir + "env" + (TSystem::getUserName().toStdString() + ".env");
+    m_envFile = profilesDir + "users" + TSystem::getUserName().toStdString() +
+                "env.ini";
   }
 
   void init() {
--- a/toonz/sources/include/toonz/toonzfolders.h
+++ b/toonz/sources/include/toonz/toonzfolders.h
@@ -44,6 +44,8 @@ DVAPI TFilePath getLibraryFolder();
 DVAPI TFilePath getReslistPath(bool forCleanup);
 DVAPI TFilePath getCacheRootFolder();
 DVAPI TFilePath getProfileFolder();
+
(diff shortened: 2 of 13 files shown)


artisteacher:
> A future enhancement for shared machines or networked installations would be to create a profiles\users\template folder to be used when creating new users. It would contain initial configurations/layouts set up by an admin user.

Love this idea!

jeremybullock:
I'd like to include this in 1.1 to get folks used to this.  Thanks for your work here.
I think it could also be nice to move the layouts folder out of profiles, so that profiles only contain user stuff.

In layouts, I think we can delete:
check.bmp
menubar.xml
the toolbars folder

What are your thoughts?  I think this is ready to go as is, but though I would ask about that.

manongjohn:
Thanks. I think the sooner this is introduced the better. I will need this for the Import Preferences feature I am working on.

Ideally I would like to remove `config`, `doc` and, `profiles` out of the stuff folder which I hope to address in the near future.  I try to remember to remove files that are no longer needed when I do that.

jeremybullock:
Thanks @manongjohn LGTM.

PR #401, opened 2020-10-21
Fix current brush preset tracking

manongjohn:
This PR fixes 2 issues related to creating and deleting brush presets. 

**Issues**
1.  Creating brush preset names does not flag it as the current brush when saving

To Reproduce:
- Select Brush Tool
- Add Preset and name it BrushA
- Close and Reopen Tahoma2D
- Select Brush Tool -> Preset should be BrushA, but restores back to <Custom>

2. Deleting the current preset and then using the brush tool after restart causes crash

To Reproduce:
- Select Preset brush BrushA from above
- Delete preset BrushA
- Close and Reopen Tahoma2D
- Select Brush Tool -> Crash

Helped a user some time ago with # 2, but didn't know how that happened until recently.

**Fixes**
1. Added logic to update the environment variable to an appropriate value immediately after adding or deleting a preset.
2. Added logic to check if the value exists in the preset list and default it to "<custom>" if not found.

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -205,7 +205,9 @@ void FullColorBrushTool::onActivate() {
         QString::fromStdString(FullcolorBrushPreset.getValue()).toStdWString();
     if (wpreset != CUSTOM_WSTR) {
       initPresets();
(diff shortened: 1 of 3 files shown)


jeremybullock:
Confirmed.  Thanks for the fix.

PR #402, opened 2020-10-21
Fix audio memory leak and fix windows distortion

jeremybullock:
This takes care of the bulk of the memory leak that happens when using audio.  The sound cells that are created with getCell are never released.  And they build up over time.  It's slow, but noticeable.  

There are times when getCell needs to be called, but this avoids calling getCell on sound cells in many cases if the information needed could be obtained another way.

This fix also changes the default internal sample rate to 44100 instead of 22050 for Windows.  This takes care of some of the big audio distortion that was present on some files.

A question that the community needs to decide. . .
We can make scrubbing be really responsive, but it will sound more choppy as it cuts itself off as scrubbing switches frames, or we can have smoother sounding scrubbing that isn't as responsive.  In the second option each frame of audio is played in its entirety.  It's more pleasing to work with, but not quite as accurate as the first option.

The second option is what we have had previously.


melieconiek:
Thank you for the memory leak fix. :)

On the PR with the unified audio code the pops and snaps can make it a little difficult to pick out phonemes, and caused me some ear fatigue (it's irritating.) It's not just scrubbing--the distortion can also be caused during normal playback by viewer lag. I was able to cause consistent popping by enabling all of the levels in a project. The extra precision is nice, but it's not as if the current implementation is bad enough to warrant replacement on it's own, and there is a pretty severe downside.

I can certainly understand wanting to simplify the code base--is there a way the choppiness during playback could be reduced with this change?

jeremybullock:
this new pr as it is uses the old method of scrubbing and the old windows audio code, so this one should be about the same as ot minus the memory leak- or at least most of it.

the one thing that could use a little more testing from users is throwing audio that used to distort at it and seeing if its better.

PR #405, opened 2020-10-23
Fix create blank frame not using user default palette

### LINKED ISSUE #404: Create Blank Drawing command does not use default palette.
**Describe the bug**
After triggering `Create Blank Drawing` in an empty column the resulting new level is not assigned the default palette.

**To Reproduce**
1. Select an empty column.
2. Trigger `Create Blank Drawing`

**Expected behavior**
The new level should use the default palette.

**Screenshots**
_demo

**Device Information:**
 - OS: Windows 10
 - Tahoma2D Version: Latest Commit (10-21)

manongjohn:
This PR fixes #404 

Added a check to only initialize the level with a base palette (clear, black) during new frame creation is there isn't one at the time of frame creation.

Added logic to try and set the user default palette on a level if it is missing a palette when creating a new frame via touchImage().  This is more of an enhancement of the OT crash fix so it uses the user default palette instead of the base palette.

### DIFF
--- a/toonz/sources/tnztools/tool.cpp
+++ b/toonz/sources/tnztools/tool.cpp
@@ -349,6 +349,15 @@ TImage *TTool::touchImage() {
   TXshSimpleLevel *sl = cell.getSimpleLevel();
 
   if (sl) {
+    // If for some reason there is no palette, try and set a default one now.
+    if (!sl->getPalette() &&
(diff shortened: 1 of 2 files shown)


jeremybullock:
Thanks for the fix.  Sorry for breaking it in the first place.

manongjohn:
No prob. As the code diverges more, it's bound to happen from time to time when incorporating OT fixes.

PR #415, opened 2020-10-26
Recent Projects and Default Project Location

### LINKED ISSUE #380: Add Recent Project list in Switch Project dialogue box
Everytime when you want to switch to previous project, you have to repeat the process of navigating into your project folder in File Browser

Add Recent Project list in `Switch Project` dialogue box

Recents Project list tahoma

jeremybullock:
This adds recent projects to the file browser and to the switch project startup.

It also adds a default project location to the preferences- which sets the starting location when you choose create project.

You can clear the recent projects from the file menu.

closes #380


jeremybullock:
Is this important enough to go into 1.1 or should it wait until 1.2?

artisteacher:
It would be amazing for this to make 1.1, though I'm sure we'd all survive if it works out better for 1.2.

artisteacher:
I'm getting a crash after trying to change the project location from the startup pop-up. As soon as I click on "Documents" in the sidebar, there's a crash.

Trying to change the default path in preferences also causes a crash as soon as I click on "...". 

[crash_project.txt](https://github.com/turtletooth/tahoma2d/files/5442661/crash_project.txt)

artisteacher:
If I change the default location in preferences by typing the location, everything works fine afterwards with no crashes.

jeremybullock:
@manongjohn I don't have access to discord right now, so I can't check your response there.  Did you want to take this over and add the combo box yourself, or would you like me to do it?

manongjohn:
If you want to move forward with this you can.

I can always submit a new PR to bring back the old project dropdown using recent projects for the list and also add a text box to show actual location.

jeremybullock:
@artisteacher What happens when you click in Documents in the browser?

jeremybullock:
@artisteacher I have the same issue as you on another computer.  On my other computer I changed the location of my documents and desktop folders, and Qt doesn't point to the right location.   I need to figure that one out.

artisteacher:
So the problem with the dropdown right now, is that you can only select recent projects. What if you want to switch to a project that hasn't been used recently enough to display in the list? If you want to create a new scene in an old project, there's not an immediate way to do so.

jeremybullock:
This isn't quite done yet. I'm going to make the last option in the dropdown a choice to browse. You can then choose an existing project from anywhere.

jeremybullock:
I think this is done.

PR #420, opened 2020-10-28
Motion Path Panel

jeremybullock:
This will create a new panel for selecting and modifying motion paths.
This is mostly for stop motion work.  Instead of tracking an object to a motion path, this makes motion paths visible to serve as a guide for a stop motion object.

You can set a number of steps for a motion path to be divided into, and tick marks will be shown on the path at each step.
I will be adding a way to change the interpolation of the tick marks.

spline

This also makes it possible to edit motion paths even if they are not connected to an object.

This is very early WIP.  Posting here so I can see my changes as a reference.


RodneyBaker:
This looks to be VERY useful. Checking out the current iteration now.  :)

beeheemooth:
Cool!

jeremybullock:
@RodneyBaker The current iteration doesn't actually do much.  The panel is empty (I'm working on that now.)

You can see the changes in the schematic however.  You can now edit a motion path without it being connected.

jeremybullock:
I'm planning on making this available in two places- a stand alone panel, and also as a tab in the stop motion panel.

jeremybullock:
Some progress:
linear

artisteacher:
I kinda want a motion room now! A motion path panel, the stage view of the schematic...

I love that you are adding access to the relevant tools directly into the panel.

artisteacher:
The build currently fails on macOS:
Screen Shot 2020-11-07 at 10 37 29 AM
Changing qOverload to QOverload gives the following error:
Screen Shot 2020-11-07 at 10 41 44 AM

jeremybullock:
When using the capitalized version of qoverload, it needs to be QOverload<int>::of
If you can confirm that works, I'll update the pr.

jeremybullock:
Github is removing some the text. The correct form is here https://forum.qt.io/topic/93385/how-to-use-qoverload/8

artisteacher:
That fixes it. This is a great feature!

If you rename the path in the Stage Schematic, the name does not update in the motion path panel. It would probably be nice to be able to rename from the motion path panel too.

Style wise, the path name/visibility feels a little disconnected from the width, color and step values. Perhaps a subtle shift in background color (alternating for each path) could be beneficial? Or dividing lines?

I had been hoping we'd be able to easily get something like this with the interpolation:
Screen Shot 2020-11-07 at 11 31 14 AM
But it looks like that is only possible right now if each arc is a separate guide.

If the geometric tool is active, it might be helpful to change the type directly from the panel.

I wonder if the new path button should use the same icon as it does in the stage schematic window - for consistency. Though the plus sign might be more understandable...

Similarly, the existing delete icon (trash can icon) could be beneficial. Or maybe it would look too busy...

RodneyBaker:
The appveyor check finally passed so was able to try this out.
First look... very impressive Jeremy.

I'm still exploring but want to give my first impression.

jeremybullock:
@artisteacher Thanks for the feedback.

I think there are some paths that are best created by hand, or as a series of paths.

I think the + and - are simple and clear, so I'd prefer to stick with them.

The separation of the path name from the other controls is to give room for path names.  I'll try and think about this some more.

I'll look into the renaming stuff and choosing the geometric tool from the panel.

jeremybullock:
You can now play a sample animation along the path:
play

jeremybullock:
Alright, so I think this is pretty much ready.  You can now rename paths from the panel, and changing the name in the schematic updates the panel.  If the geometry tool is selected, you can choose the type of tool from the panel (except for the polygon tool.)

doneish

artisteacher:
Changing the name works great & it is really helpful being able to change the geometric type. I have found a few problematic behaviors:

If you press enter to complete the name change on the stopmotion path tab, it tries to capture. 

I found a crash from moving/selecting a control point on the curve and then trying to change the step number
[path_crash.txt](https://github.com/turtletooth/tahoma2d/files/5520647/path_crash.txt)

The motion path panel should probably show the curve area on open & be a bit wider so "Step" isn't cut off.
Screen Shot 2020-11-10 at 5 46 04 PM

Changing the geometric tool type works well on the panel, but doesn't work on the stopmotion tab. The name changes, but the behavior continues to act as the rectangle.
Screen Shot 2020-11-10 at 6 00 46 PM

If you use multiple path panels - the geometric type doesn't always display. To reproduce:
1. Work with a path on the stopmotion room's path tab
2. Choose the geometric tool & change the type
3. Open a motion path panel 
Result: geometric tool is active, but type dropdown is missing until changing away from the geometric tool & back
Screen Shot 2020-11-10 at 6 07 00 PM

jeremybullock:
@artisteacher Thanks for the feedback.  I can't duplicate the crash, can you share the steps please?

Regarding the geometric tool - it looks like the command to switch tool types only works when the tool option bar is visible.  If that is the case, I may remove the geometric tool as an option from the path panel since it wouldn't always work.

I'll look into blocking the capture on the name change.

artisteacher:
1. In the stopmotion path tab, create a path
2. Use the brush to draw a path
3. Adjust the curve - add multiple points to the curve & adjust them:
Screen Shot 2020-11-10 at 7 39 39 PM
4. Try to change the step number - beginning to type causes a crash

jeremybullock:
I can't duplicate it.  :(

jeremybullock:
Alright.  I fixed it trying to capture an image from the stop motion panel and I fixed the geometry options working even if a tool options panel was not visible.  I got polygon sides working.  

Unfortunately, I still couldn't get it to crash.  I tried.  And tried.

artisteacher:
Here is what Xcode shows after the crash is triggered:
Screen Shot 2020-11-11 at 1 57 32 PM
Screen Shot 2020-11-11 at 2 01 38 PM

artisteacher:
This is what I get in Xcode:
Screen Shot 2020-11-11 at 2 42 14 PM

jeremybullock:
Got it!

Thanks for the help @artisteacher

artisteacher:
The crash is fixed & changing the geometric type now works!

RodneyBaker:
Jeremy,  I'm getting a crash when I attempt to load a path motion file using the Right Click method on a motion path node in the Schematic.   I don't get this same crash with Opentoonz so am currently assuming this might relate to changes incorporating the motion path panel.

To test, Right Click on a motion path in the Schematic and attempt to save out the motion path file.
You should likely get a crash here.
If successful, try to reopen that path file via same methodology.

jeremybullock:
Thanks Rodney.  I forgot about saving and loading paths.  It should be fixed now.

quirinux-so:
> Some progress:
> linear

Wonderful feature, this can make people start using Tahoma instead of DragonFrame. Just one detail: I notice that the increments are uniform. Is it possible that, as in DragonFrame, the program calculates the acceleration / deceleration, modifiable by nodes? 

Reference:
https://my.opendesktop.org/apps/files_sharing/publicpreview/rRNqGFTXcTMTw3R?x=1280&y=450&a=true&file=increment-editor.png&scalingup=0

quirinux-so:
> 1. In the stopmotion path tab, create a path
> 2. Use the brush to draw a path
> 3. Adjust the curve - add multiple points to the curve & adjust them:
> 
> Screen Shot 2020-11-10 at 7 39 39 PM
> 
> 4. Try to change the step number - beginning to type causes a crash

Excuse my ignorance, this is seen in the figure, are the nodes with which we could control the acceleration / deceleration of the increments?

Reference:
https://github.com/tahoma2d/tahoma2d/pull/420#issuecomment-798882243

artisteacher:
You can use the curve to control acceleration/deceleration. The nodes seem to let you create multiple transitions for the acceleration/deceleration but they aren't the easiest to map/control when the curve gets complex. 

I think if you have a complex path/multiple transitions, it is actually better to make multiple paths.

quirinux-so:
> You can use the curve to control acceleration/deceleration. The nodes seem to let you create multiple transitions for the acceleration/deceleration but they aren't the easiest to map/control when the curve gets complex.
> 
> I think if you have a complex path/multiple transitions, it is actually better to make multiple paths.

The increment editor tool, similar to motion path, was designed to draw references for stop-motion photography on the screen. The idea is to be able to see an acceleration / deceleration guide on the screen **that we will use to take the photographs**, not the creation of automatic intermediates.

PR #422, opened 2020-10-28
Add Import Preferences option

manongjohn:
This PR adds an option to load Preferences and Settings from another installation of Tahoma2D.  This will aid in upgrading to newer versions of Tahoma2D without having to overwrite existing versions with the current.

image

For Windows and Linux builds, users would need to select the folder containing the Tahoma2D application since the `tahomastuff` folder is also there.  Technically any folder containing a `tahomastuff` folder would work. 

For macOS, they have a slightly different message to select the Tahoma2D application itself, since the `tahomastuff` folder is buried inside it.

It is backwards compatible, in that it will pull files from old locations and store in the new location (`profiles\users\xxxx`)

For non-user specific files and directories (fx, studiopalette, plugins, library, sandbox, projects, toonzfarm), it performs a destructive merge.  Files found in source but not in destination are added, common files are replaced with source and files only in destination are not touched.

A restart is required, but not enforced, afterwards for room layouts and other preference changes to take effect.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -633,6 +633,7 @@ void TEnv::setStuffDir(const TFilePath &stuffDir) {
 }
 
 void TEnv::saveAllEnvVariables() { VariableSet::instance()->save(); }
+void TEnv::loadAllEnvVariables() { VariableSet::instance()->load(); }
 
 bool TEnv::setArgPathValue(std::string key, std::string value) {
   EnvGlobals *eg = EnvGlobals::instance();
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -283,7 +283,8 @@ void TSystem::rmDirTree(const TFilePath &path) { ::rmDirTree(toQString(path)); }
 
 //------------------------------------------------------------
 
-void TSystem::copyDir(const TFilePath &dst, const TFilePath &src) {
+void TSystem::copyDir(const TFilePath &dst, const TFilePath &src,
+                      bool overwrite) {
   QFileInfoList fil = QDir(toQString(src)).entryInfoList();
 
   QDir::current().mkdir(toQString(dst));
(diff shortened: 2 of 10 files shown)


jeremybullock:
A-MAZE-ING!

artisteacher:
For user specific data, will it import from all users?

manongjohn:
> For user specific data, will it import from all users?

Currently, No.  Only the user running the application. File permission wise, that might be problematic, especially on a macOS/Linux platform.

jeremybullock:
Finished with a read-through of the code.  Will test this tonight.

jeremybullock:
Thanks for working on this @manongjohn - I think this is going to be a big time-saver for folks.

jeremybullock:
I just tested this.  The majority of this seems to be working well.  There are two things that I noticed.  
It didn't import my room layout even though that was checked.  I added a level strip to 2d and created another room with just an xsheet.  Those didn't get pulled in after a restart.

The recent scene file makes recent scenes in the old install still show up.  This could confuse folks if they think they are opening the scene that was copied to the new install, but edit the old install.

Preferences , palettes and shortcuts seem to be working well.

manongjohn:
Thanks. Will look into the issues you found.

manongjohn:
> It didn't import my room layout even though that was checked. I added a level strip to 2d and created another room with just an xsheet. Those didn't get pulled in after a restart.

This worked for me, importing from 1.0.1 and from 1.1. I used both the PR artifact and a local build. When you modified the layout in the old Tahoma version, did you close it to save the changes first before attempting to import it into the new Tahoma version?

> The recent scene file makes recent scenes in the old install still show up. This could confuse folks if they think they are opening the scene that was copied to the new install, but edit the old install.

I'll look at adding some logic to scan the recent file list and change old stuff folder path to new one, only if they also checked to import the stuff folder.

jeremybullock:
Thanks for this @manongjohn 
I think I found out what was happening with room layouts.  It isn't a problem with this pr.  It looks like Tahoma doesn't like user profiles with a period in them.  I tested this on a machine where my user account is jeremy.bullock.  All of the other preference files get stored in the right location, but the rooms got stored in a jeremy folder (without the .bullock), and the folder with the room in them is "Default.bullock".

So it looks like there is an error in the logic for the room location when dealing with users with periods in the name.  I don't think it's related to this pr.

jeremybullock:
This seems to be working well apart from the rooms logic which isn't part of this PR and can be resolved later.

Thanks so much for this @manongjohn - you are goin to save folks a lot of time.

manongjohn:
> It looks like Tahoma doesn't like user profiles with a period in them.

Interesting find.  Just so I understand.  For username = "jeremy.bullock"

Tahoma 1.0.1 paths (same in OT?):
- Preferences: profiles\layouts\settings.bullock
- Room layouts: profiles\layouts\personal\Default.bullock

Tahoma 1.1 paths:
- Preferences: profiles\users\bullock
- Room layouts: ???

jeremybullock:
It looks like it got introduced in 1.1.
I'm uploading profile folders for 1.1 and 1.0.1
[1.1.zip](https://github.com/turtletooth/tahoma2d/files/5519423/1.1.zip)
[1.0.1.zip](https://github.com/turtletooth/tahoma2d/files/5519429/1.0.1.zip)

manongjohn:
Thanks.  Should probably open a new issue for that.

PR #451, opened 2020-11-01
Fix loading brush presets or last brush used

manongjohn:
This PR does the following

1.  Fixes a crash when loading a custom preset Smart Raster brush with a `Hardness` setting < 100. 

Issue: Some internal variables to handle Smart Raster brush with `Hardness` < 100 were not initialized correctly when loading a preset and crashed when used.

Resolution: Ensured when loading a preset, the appropriate logic was followed as if the `Hardness` value was changed so that internal variables were initialized accordingly.

2. Logic for loading preset brushes or reloading the last brush used enhanced

Issue: When certain brush options are changed on the tool's option bar, it can trigger additional logic which can impact how the brush works.  When restoring presets or restoring the last brush used, this logic was not being performed and could impact the behavior of the brush. I reviewed all 3 brush types and found potential issues with Vector and Smart Raster brushes.

Resolution:  Updated the `loadPreset()` and `loadLastBrush()` logic to include any additional logic that would normally be triggered when changing brush properties on the tool's option bar.

3. No longer set you back to `<custom>` preset when modifying an option not saved with preset.

Issue: Was forcing back to custom on any property change, which it didn't need to be.

Resolution: Added condition only when a preset value changes.

4. `notifyTool` was not properly updating some of the Vector checkboxes

Issue: The checkboxes were part of Property set 0, but when setting up the controls, it was binding them under Property Set 1

Resolution: Moved the binding of controls to properties in the section that handles Property set 0.

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -1891,7 +1891,6 @@ BrushToolOptionsBox::BrushToolOptionsBox(QWidget *parent, TTool *tool,
     builder.setSingleValueWidgetType(ToolOptionControlBuilder::FIELD);
 
     addSeparator();
-    if (tool && tool->getProperties(1)) tool->getProperties(1)->accept(builder);
     m_drawOrderCheckbox =
         dynamic_cast<ToolOptionCheckbox *>(m_controls.value("Draw Under"));
     m_autoCloseCheckbox =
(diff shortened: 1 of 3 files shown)


melieconiek:
As requested on Discord:

1. Enter a size into 'Min' setting on the vector brush tool (10px width) so that it's registered as the <custom> preset.
2. Switch to another preset with a different size (2px width.)
3. Enable Auto Fill --> Vector Brush will switch to the <custom> preset.
4. Restart Tahoma.

The brush tool will start with the <custom> preset, but it will not be set to 2px width as expected--the active brush settings when you closed Tahoma.

It either needs to ignore the Auto Close/Auto Fill settings or correctly register them to the <custom> preset (or whatever it's doing wrong.)

manongjohn:
Ok. The last commit should fix tracking how properties change when switching back from a preset to `<custom>`.

Also, if you change an option that is not saved as part of the preset, it won't kick you back to `<custom>`.  This mostly affects vectors options.

melieconiek:
Thank you @manongjohn. :)

jeremybullock:
Just tested, and this works nicely.  Thanks @manongjohn

PR #479, opened 2020-11-10
Replace coded default Studio Ghibli rooms with Tahoma2D rooms

manongjohn:
This PR replaces the hardcoded Studio Ghibli rooms used as a default when for whatever reason, the room files are not found.  It will now create the current Tahoma2D rooms.

To test:
1. Navigate to `profiles\users\xxx\layouts\Default` folder and delete all the `room*.ini` files only.  Leave the `layouts.txt` file.
2. Open Tahoma2D

In older versions, the Studio Ghibli default rooms (plt, inkandpaint, etc) would load.  With this PR, it should match what our current Tahoma2D rooms are.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -704,25 +704,25 @@ void MainWindow::readSettings(const QString &argumentLayoutFileName) {
                                  TFilePath("fliphistory.ini"));
 
   if (rooms.empty()) {
-    // PltEditRoom
-    Room *pltEditRoom = createPltEditRoom();
-    m_stackedWidget->addWidget(pltEditRoom);
-    rooms.push_back(pltEditRoom);
+    // 2D Room
+    Room *room2D = create2DRoom();
+    m_stackedWidget->addWidget(room2D);
+    rooms.push_back(room2D);
 
-    // InknPaintRoom
-    Room *inknPaintRoom = createInknPaintRoom();
-    m_stackedWidget->addWidget(inknPaintRoom);
-    rooms.push_back(inknPaintRoom);
+    // Stop Motion Room
+    Room *roomStopMotion = createStopMotionRoom();
+    m_stackedWidget->addWidget(roomStopMotion);
+    rooms.push_back(roomStopMotion);
 
-    // XsheetRoom
-    Room *xsheetRoom = createXsheetRoom();
-    m_stackedWidget->addWidget(xsheetRoom);
-    rooms.push_back(xsheetRoom);
+    // Timing Room
+    Room *roomTiming = createTimingRoom();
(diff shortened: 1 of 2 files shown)


jeremybullock:
You are killing it.  I'm stuck on one pr and you just keep knocking it out of the park.

toonz/sources/toonz/mainwindow.cpp
```diff
@@ -800,151 +800,246 @@ void MainWindow::writeSettings() {
...
+  // LevelPalette
+  TPanel *levelPalettePane = TPanelFactory::createPanel(room, "LevelPalette");
+  if (levelPalettePane) {
+    levelPalettePane->setGeometry(QRect(34, 392, 259, 368));
+    geometries.push_back(levelPalettePane->geometry());
+    room->addDockWidget(levelPalettePane);
+    layout->dockItem(levelPalettePane);
+  }
+
+  DockLayout::State state(geometries, "-1 1 [ 4 [ 3 [ [ [ 0 5 ] 2 ] 1 ] ] ] ");
```
jeremybullock, line 868: I'm not questioning this one. . . just trying to learn.  What does this line do?

artisteacher:
I wonder what causes the rooms to be deleted? I've had it happen once or twice.

toonz/sources/toonz/mainwindow.cpp
manongjohn: Your guess is as good as mine. LOL.

Honestly I'm not sure entirely what it is.  I haven't really looked too deeply into it.  It some hierarchical notation of the regions the different panes sit in.

manongjohn:
> You are killing it. I'm stuck on one pr and you just keep knocking it out of the park.

What you are doing is big vs the little stuff I do here in and there. It's all good :)

manongjohn:
I'm not sure what causes that particular issue either and I'd be interested to find out. This would typically only happen when quitting since that's really the only time those files get updated.

Historically the way things are updated in OT is the original file is blindly  removed first and then a new file is written.  This has proven to be problematic in the past when the write runs into an issue.  By then it's too late since the original file is gone.

artisteacher:
> Historically the way things are updated in OT is the original file is blindly removed first and then a new file is written. This has proven to be problematic in the past when the write runs into an issue. By then it's too late since the original file is gone.

I would think it would be better to rename the original file first - ie. somename.filetype.old, and delete it only after the new file is written.

manongjohn:
> I would think it would be better to rename the original file first - ie. somename.filetype.old, and delete it only after the new file is written.

Yeah. Normally when I run across that, prompted usually by a bug report, i change it to write 1st to a temp file, then rename the temp to the original file, overwriting it. Will need to see what's happening here.

jeremybullock:
LGTM.  Thanks!

PR #483, opened 2020-11-10
Fix username room layout directory path

### LINKED ISSUE #482: Rooms path not set correctly for users with a period in the username. (Windows)

manongjohn:
This PR fixes #482 

Issue: When building certain paths, it was using a TFilePath function `getWideName()` that was stripping extensions from the filename.  For usernames with periods in it, anything after the last period was considered an extension and omitted incorrectly.

Resolution: Reworked the logic so it didn't need to use  `getWideName()`


jeremybullock:
This fixes it.  Thanks!

PR #489, opened 2020-11-12
OpenToonz changes as of 11-11-20

jeremybullock:
This includes 
opentoonz/opentoonz#3083
opentoonz/opentoonz#3583
opentoonz/opentoonz#3582
opentoonz/opentoonz#3570

I'm skipping 
opentoonz/opentoonz#3559 for now.  Do we need it?

I think the only one I need to do now is
opentoonz/opentoonz#3375


artisteacher:
Personally, I don't need 3559.

manongjohn:
For now, I think we can pass on 3559. If we start seeing reports from the community about the lag with studio palette lookups, we can look to adopt it then.

PR #499, opened 2020-11-13
Incorporate Rhubarb Lip Sync

jeremybullock:
Here's the process so far:

It searches the columns in Tahoma for sound levels and lets you choose one.
You can play it to preview the level first.
Next you generate the dat file with a button (This runs rhubarb in the background)
This auto-populates the file field with the new dat file.
All that's left is to choose your mouth shapes and hit apply.

I have it working, but now I need to figure out how I want it to look.

### DIFF
--- a/toonz/sources/include/toonz/txsheet.h
+++ b/toonz/sources/include/toonz/txsheet.h
@@ -499,7 +499,7 @@ in TXsheetImp.
   */
   int getFirstFreeColumnIndex() const;
 
-  TSoundTrack *makeSound(SoundProperties *properties);
+  TSoundTrack *makeSound(SoundProperties *properties, int col = -1);
 #ifdef BUTTA
   /*! Returns \b TSoundTrack with frame rate \b \e frameRate computed calling
           \b TXshSoundColumn::mixingTogether() of all sound column contained in
--- a/toonz/sources/toonz/lipsyncpopup.cpp
+++ b/toonz/sources/toonz/lipsyncpopup.cpp
@@ -27,6 +27,7 @@
 #include "filebrowsermodel.h"
 #include "xsheetdragtool.h"
 #include "historytypes.h"
+#include "tsystem.h"
 
 // Qt includes
 #include <QHBoxLayout>
@@ -37,6 +38,8 @@
 #include <QPainter>
 #include <QSignalMapper>
 #include <QComboBox>
+#include <QProcess>
+#include <QTextEdit>
 //=============================================================================
 /*! \class LipSyncPopup
                 \brief The LipSyncPopup class provides a modal dialog to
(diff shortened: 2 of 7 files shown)


jeremybullock:
sync

artisteacher:
I'm looking forward to trying this!
Screen Shot 2020-11-13 at 4 59 09 PM

justburner:
Great, the crash has been solved and works multiple times now!
But sadly found a few more bugs:
1. Choosing a file with relative path to the project does not work. `+extras\recordedAudio1.wav` will fail but absolute path such as `bahbah\Tahoma2D\tahomastuff\sandbox\extras\recordedAudio1.wav` works.
2. Choosing a file and changing the filename field doesn't re-enables the Generate button.
3. **Major!**: Generating data and then clearing `Lip Sync Data File` text field will make Tahoma2D crash on `Apply`.
4. *Minor*: Re-opening the Lip Sync window resets the combobox to index 0 instead of the last used audio source.
5. *Minor*: I see duplicated files on the File Browser window when picking Lip Sync Data File, likely an OpenToonz bug.

Small suggestion, it should be more intuitive for the user to see the actual column name in the combobox but analyzing the source code seems it's not directly possible... maybe you can add a QList\<int\> to translate currentIndex into column number, that way you don't have to split and convert text to number either (that's bad if locale is implemented later in the future).

artisteacher:
It's working well! I have noticed that the next/previous buttons are only clickable in the small "box" close to their icons:
Screen Shot 2020-11-14 at 3 15 52 PM

It would be great for the play button to change to a stop button when an audio track is being played. We might not always need to listen all the way through.

The first time I opened Tahoma, before I had rhubarb in the correct directory, there was some temporary visual distortion when the lip sync window first opened (from hiding the rhubarb relevant fields?). I haven't been able to trigger that again.

jeremybullock:
Thanks for the feedback @justburner 
I think I have 1-5 fixed.
#6 is beyond the scope of this.

jeremybullock:
@artisteacher 
I think the button thing may be a mac bug - I can click the buttons anywhere.
I added the ability to stop playback.
I think I fixed the window glitch on opening if Rhubarb isn't present.

artisteacher:
Looking more closely, it actually seems to be all of the buttons in my build. 1.1 doesn't have this issue. I have a build I made on Oct. 7th where the buttons are all fine. I'll see if I can narrow down when it started - or if it's due to a change in my build environment.

jeremybullock:
sync

artisteacher:
I love the new tabs! Everything seems to be working great so far on my end, though I don't think the rest drawing is actually extending to the end marker.

I'm pretty sure that Rhubarb actually uses 9 drawings instead of 10. It looks like the WQ drawing is the one that is skipped & the CDG... drawing is used instead. Perhaps it should be hidden on the audio tab.

jeremybullock:
@artisteacher I just checked and rest to end is working.  It only works on frames in the selected column that go past the selected sound.  If the column didn't start longer than the sound, it won't do anything really.

You are probably right about the phonemes.  I'm going to leave the 10 there until I better understand what is happening under the hood.  The developer of rhubarb recommends not actually using the preston blair phonemes and using the 1-6 mandatory and 7-9 optional phonemes he uses.

Eventually I can modify the dialog to use those phonemes when using audio and show the 10 for importing from papagayo.  But I'll leave that for another pr.

PR #503, opened 2020-11-15
Fix translation of menubar

### LINKED ISSUE #490: Lost translation
**Describe the bug**
The top menu is not translated, everything was fine before, until today's wizard installed

**To Reproduce**
1. Clicked in the settings - translate into Russian
2. Restart the program
3. Then you will see in the picture

**Expected behavior**
There should also be Russian as before
**Screenshots**
Screenshot_1

**Device Information:**
 - OS: Windows 10
 - Tahoma2D Version: [e.g. Nightly (2020-12-11)]
 - GPU: Rx 580

manongjohn:
This PR fixes #490 

The original `menubarpopup.cpp` provided the translation for the menubar through `ShortcutTree::tr()`. Since that was removed, the menubar stopped translating. 

As I wasn't sure if fixing all the .ts files and regenerating the .qm files would fix it, I opted to updated the code to reuse  `ShortcutTree::tr()` which fixed the problem.

Aside: Tahoma will need help at some point with translation files since the OT ones we are currently using will not work with some of the UI changes we've made.

### DIFF
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -8,6 +8,7 @@
 #include "cellselection.h"
 #include "mainwindow.h"
 #include "docklayout.h"
+#include "shortcutpopup.h"
 
 // TnzQt includes
 #include "toonzqt/tselectionhandle.h"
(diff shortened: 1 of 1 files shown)


jeremybullock:
Awesome! Thanks @manongjohn

PR #504, opened 2020-11-15
Fix crash moving cell with shaderfx

### LINKED ISSUE #488: Shader FX with inputs can cause OT/Tahoma2D to crash
**Describe the bug**
Any Shader FX with inputs can cause OT/Tahoma2D to crash.
Occurs in both Timeline and XSheet.
(Theory: may seem related to dragging frames while rendering effect?)

**To Reproduce**
Steps are a bit convoluted in order to make it 100% reproducible so I've recorded a video with audio:
https://youtu.be/XVIF5oG5IcY

**Expected behavior**
Tahoma2D shouldn't crash, OpenToonz is also affected so this is an old bug.

**Device Information:**
 - OS: Win7
 - Tahoma2D Version: v1.1
 - GPU: 1050 Ti
 - GPU Driver Version: 456.71

manongjohn:
This PR fixes #488 

Issue: ShaderFx are derived from ZeraryFx, which normally creates a column with no input when used in an animation.

In the case of `GPU Radial Blur` and `GPU Spin Blur`, they don't generate a column.  When moving a level with an attached shader in the timeline, it was was hitting logic that expected the shader to have a column and would lead to a crash when it didn't.

Modified the logic to check for a column and act accordingly.

### DIFF
--- a/toonz/sources/toonzqt/stageobjectsdata.cpp
+++ b/toonz/sources/toonzqt/stageobjectsdata.cpp
@@ -75,7 +75,8 @@ FxCanGenerateState canGenerate(const std::set<TFx *> &fxsSet, TFx *fx) {
 
 //! Returns whether the specified fx is a downstream node of the xsheet node.
 bool hasTerminalUpstream(TFx *fx, TFxSet *terminalFxs) {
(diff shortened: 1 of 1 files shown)


jeremybullock:
Thanks @manongjohn 
Great work.

PR #505, opened 2020-11-16
Fix motion path issues

jeremybullock:
This does a number of user experience things:
It removes the dialog asking the user to confirm that they want to replace the motion path.  Changing a motion path is undoable, so that dialog isn't really necessary.

This dialog was causing an issue with using pens.  Clicking an option of the dialog was registering as a pen press in the viewer.  I'm curious if there are other instances where this is an issue.

It adds text to the viewer when a motion path is selected that says "Motion Path Selected" in the top corner.  This is to help clarify the situation when the brush or geometry tool is used while a motion path is selected,

It adds a status bar message indicating when a motion path is selected and how to get out of motion path editing.

It fixes some motion path visibility issues.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -1285,14 +1285,16 @@ public:
       bool isSpline        = false;
       if (TTool::getApplication()->getCurrentObject()->isSpline()) {
         isSpline = true;
-        if (!ToolUtils::isJustCreatedSpline(vi.getPointer())) {
-          m_primitive->setIsPrompting(true);
-          QString question("Are you sure you want to replace the motion path?");
-          int ret =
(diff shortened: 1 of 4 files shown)


DarrenTAnims:
Thanks for fixing these issues. This will really make a huge difference to using it.

One question. Do get out of motion path editing, I had to click in the timeline. I hoped that selecting the brush tool from the main toolbar, to start drawing would do so as well. It didn't. Is it possible to make selection of a tool from the main toolbar also exit motion path editing?

jeremybullock:
It's not, since those tools can also be used to modify paths.  These tools need to be able to modify paths since paths can also be selected from the schematic without the motion paths panel open.

But there is a new status bar message explaining how to get out of motion path editing, and a big text on screen that lets you know when you are using a motion path.

artisteacher:
That's such a better experience, not having the dialog.

PR #506, opened 2020-11-16
Fix guidedstroke in sceneviewer

### LINKED ISSUE #502: Vector guide not working in timeline
**Describe the bug**
The direction of vectors only works on the first stroke in the timeline.
**To Reproduce**
1. I draw the first frame, then I create empty ones, including the vector guide
2.  I turn on the onion skin and draw a stroke, where it shows the direction of the vector
3. After completing the quest, no more directions appear

**Expected behavior**
With each stroke made, there must be a new direction of vectors, it only works in level strip, when it should work and on a timeline like in opentoonz. 

**Screenshots**
bandicam-2020-11-15-01-04-29-937

**Device Information:**
 - OS: [e.g. Windows]
 - Tahoma2D Version: [e.g. Nightly (2020-13-11)]
 - GPU: RX 580

manongjohn:
This PR fixes #502 

Issue: A change made for #402 prevented the population of the current frame variable for sound columns when drawing the frame in the sceneviewer. However, `app->getCurrentColumn()->getColumn()` may not return an actual column object even though the `getColumnIndex()` is populated properly. Since the current frame variable was not populated, guided controls didn't know what frame it was on.

Resolution: Switched to use `getCurrentLevel()->getLevel()` instead when determing if the current frame variable should be populated or not.

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -2130,8 +2130,8 @@ void SceneViewer::drawScene() {
       args.m_osm         = &osm;
       args.m_xsheetLevel = xsheetLevel;
       args.m_isPlaying   = frameHandle->isPlaying();
-      if (app->getCurrentColumn()->getColumn() &&
-          !app->getCurrentColumn()->getColumn()->getSoundColumn())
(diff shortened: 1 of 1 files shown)


jeremybullock:
Sorry, I broke it.

jeremybullock:
I can confirm the fix.  Thanks @manongjohn

PR #511, opened 2020-11-16
Fix vector skeleton picking

### LINKED ISSUE #500: Skeleton tool focus issue with vector levels
**Describe the bug**
It is tricky to change focus to certain vector levels in the Skeleton tool's Animate mode. This isn't an issue with raster formats. 

On raster levels, clicking anywhere on a different drawing changes the focus. On vector levels however, access to levels that are behind can be blocked. 

**To Reproduce**
1. Create a rigged character (side view) with vector levels
2. Use the Skeleton tool's "Animate" mode to select and move the back thigh & the back upper arm

**Expected behavior**
I expect to be able to click on any drawing to select it, without needing to use the Timeline/Xsheet to change focus.

**Screenshots**
Screen Shot 2020-11-13 at 1 16 46 PM
The back thigh in this screenshot can be clicked on to change focus if the levels are smart raster. If they are all vector levels, clicking on the thigh won't change focus. It doesn't seem to matter if you click on stroke or fill.

**Device Information:**
 - OS: macOS
 - Tahoma2D Version: 1.1\

jeremybullock:
Fixes #500 

This seems to fix the issue of picking the right layer with the skeleton tool.  Same fix I did earlier with the shift and trace tool.

### DIFF
--- a/toonz/sources/tnztools/skeletontool.cpp
+++ b/toonz/sources/tnztools/skeletontool.cpp
@@ -383,7 +383,7 @@ void SkeletonTool::leftButtonDown(const TPointD &ppos, const TMouseEvent &e) {
   if (m_device < 0) {
     // No gadget clicked.  Select the column
     std::vector<int> columnIndexes;
-    getViewer()->posToColumnIndexes(e.m_pos, columnIndexes, getPixelSize() * 5,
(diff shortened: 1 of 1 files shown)


artisteacher:
This is working for me. Thanks for a fast fix!

PR #518, opened 2020-11-19
Segment Eraser for Smart Raster Levels

jeremybullock:
This adds a segment eraser mode for smart raster levels.

This is pretty accurate, but not as accurate as the segment eraser for vectors.

Since it is dealing with pixels, there is some guess work around where the segments are.  It uses the same code as the segment fill.

### DIFF
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -76,7 +76,7 @@ DVAPI void inkFill(const TRasterCM32P &r, const TPoint &p, int ink,
 
 bool DVAPI inkSegment(const TRasterCM32P &r, const TPoint &p, int ink,
                       float growFactor, bool isSelective,
-                      TTileSaverCM32 *saver = 0);
+                      TTileSaverCM32 *saver = 0, bool clearInk = false);
 
 void DVAPI rectFillInk(const TRasterCM32P &ras, const TRect &r, int color);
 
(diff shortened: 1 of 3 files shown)


manongjohn:
Playing around with it briefly, if you have Mode = Areas, the Segment eraser still erases segments. Although that combination makes no sense, I still think, for consistency that it shouldn't erase any lines.  Perhaps it could just default to erasing areas?  That would also lend itself to erasing Lines & Areas as welll where it could remove segments and adjoining fills.

I do see it occasionally does erase incorrectly. Sometimes it will erase all the connected lines to the segment and sometimes just act like a normal erase, erasing the portion of the line. but beyond those few instances which a undo fixes, it works pretty well otherwise.

Nice job!

RodneyBaker:
Here's my inaugural test of the feature:
segment erase

jeremybullock:
Nice one Rodney.

jeremybullock:
@manongjohn I just changed it to disable the lines/areas dropdown if segment mode is on to indicate that it will only erase segments.

artisteacher:
This is really nice! I did have to undo a few times due to too much getting erased while testing. When that happens, using undo & then approaching from a different angle always seems to work.

You can still have Areas mode active while using segment erase, if it was the active mode when switching the eraser type. Though the mode seems to be ignored in this circumstance.

jeremybullock:
Yeah- because strokes are guessed rather than explicit as in vector, precision won't be as high as the vector segment eraser. Hopefully its worth it and won't be too frustrating.

When the eraser is in segment, it totally takes over.

PR #523, opened 2020-11-21
Snap Center Point to Motion Path

manongjohn:
This PR changes how a level's center snaps to the Motion Path.

Rather than using the level's true image center to snap to the Motion Path, it was requested that the center point, which can be moved by the Animation tool, should be used instead.

### DIFF
--- a/toonz/sources/include/toonz/tstageobject.h
+++ b/toonz/sources/include/toonz/tstageobject.h
@@ -291,7 +291,11 @@ object.
   double getParam(Channel type, double frame) const;
 
   //! Sets the center of the \e frame to \e center.
-  void setCenter(double frame, const TPointD &center);
+  void setCenter(double frame, const TPointD &centerPoint,
+                 const TPointD &frameCenter);
+  void setCenter(double frame, const TPointD &center) {
+    setCenter(frame, center, center);
+  }
 
   //! Returns the center of the \e frame.
   TPointD getCenter(double frame) const;
@@ -526,7 +530,7 @@ private:
   PlasticSkeletonDeformationP
       m_skeletonDeformation;  //!< Deformation curves for a plastic skeleton
 
-  TPointD m_center;
+  TPointD m_center, m_frameCenter;
   TPointD m_offset;
 
   // ----For Inverse Kinematics interpolation
--- a/toonz/sources/include/toonz/txsheet.h
+++ b/toonz/sources/include/toonz/txsheet.h
@@ -319,7 +319,11 @@ public:
           \b \e frame, to \b \e center; it calls \b TStageObject::setCenter().
           \sa getCenter()
   */
-  void setCenter(const TStageObjectId &id, int frame, const TPointD &center);
+  void setCenter(const TStageObjectId &id, int frame,
+                 const TPointD &centerPoint, const TPointD &frameCenter);
+  void setCenter(const TStageObjectId &id, int frame, const TPointD &center) {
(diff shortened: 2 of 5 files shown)


jeremybullock:
Thanks for doing this.  I wish there was a way to return the original value if 3 conditions were met:
1. The path being requested is for the current object.
2.  The current tool is the animation tool.
3.  The current mode is center edit mode.

But since this file is in toonzlib, I don't know of an easy way to detect those conditions.  So this will work fine.

manongjohn:
Could add a parameter in a few places like `bool useTrueCenter= FALSE`, or something like that, which calling functions would determine when it needs to be set to TRUE to get the value without the path adjustment.  Would that work?

Out of curiosity, in what scenario would you need true center when you are tied to a motion path?

jeremybullock:
I think that would be fine.

I would only want that so editing the center would still be intuitive even if the path was connected.  Other than that I don't think true center is ever really needed.  And if someone needs it, they just leave the center as is.  

But now that I'm talking about intuitive, what's worse- the strange way changing the center works with this fix as is, or having the drawing switch positions if you enter center editing mode?  Neither feel like awesome solutions.

What do you think about leaving this pr as is, and down the road change the edit tool to not send updates to the viewer in this situation until the mouse is released?  So you can actually drag the circle, and then when you release the new center snaps to the path?

manongjohn:
I'm fine leaving PR as is. 

I did think about how to be able to center when connected to path. I didnt like the idea of switching the view back to normal view when attempting to change center position. I would think it would be jarring. It would be better to move it from its current position. Would need to probably add a bit of logic to keep the current position while creating a faux center to move around as if it wasn't connected.

In some ways I like how it moves as you can position relative to where it is in real time but it can be a little confusing initially.

Users can always disconnect the path temporarily to change center as well.

DarrenTAnims:
I think aligning the animate tool's centre point to the motion path is much more flexible and actually maps to the documentation, so thanks for doing this.

But the way that the you drag the centre point to the centre of your column is very peculiar and will catch a lot of people out. I think it will be pretty common to create your drawing, then the path and THEN try to align the centre point. And moving it in the opposite direction, further than the character is strange and tricky to do. I think that it would be best if the drawing didn't move until you released the drag, so you can drop the centre point on the exact area of your drawing and then when released, have it snap to the path. It's not impossible to do the current way, but it is tricky and unintuitive.

DarrenTAnims:
Oh no. I've just noticed too. As I move the centre point for the column, because the path is a child of the column, that moves too. So I'm not just placing the centre of the drawn column, but I'm also moving the path away from where I've drawn it. If I've drawn the motion path in the right place on screen, I wouldn't want this to move.

ScreenCapture

manongjohn:
When I move the center on my column, my motion path does not move.  Not sure where your motion path is in the image.

DarrenTAnims:
> When I move the center on my column, my motion path does not move. Not sure where your motion path is in the image.

It wasn't. I check and then deleted the message from this thread. But the placing of the centre point is quite inconvenient. Tbh, I think I might prefer to use the centre of the page if that's the way we set up the placement.

manongjohn:
I have a fix for the relocation of center when attached to motion path.  Will be submitting update shortly.

jeremybullock:
Is this ready for checking?

manongjohn:
Yes

jeremybullock:
This seems to work well.  Thanks @manongjohn

Issue #467, opened 2020-11-07
Blank frames options in Preferences is misaligned


onigiri1:
![image](https://user-images.githubusercontent.com/46865843/98439830-18e47a80-2127-11eb-8d3d-d472892690b2.png)

The _Blank frames_ options is way too off from the left

Tahoma Build: v1.1 (October 29, 2020)

PR #532, opened 2020-11-24
word wrap to improve layout in preview prefs

trdenton:
this adds word wrap to labels in the preferences UI window to aid in alignment
before
image

after
image


trdenton:
addresses #467

DarrenTAnims:
I'm not sure if this really helps much. There's still a gap to the left of the 1st 2 options. And the word wrap on the "Number of Frames to Play for Short Play" makes it more difficult to read, IMO.

onigiri1:
The more I see it, I think the layout can be improved for categorizing sake, the settings orders seems to be out of the place
and also renaming _Number of Frames to Play for Short Play_ to _Playback frames for Short Play_
Preview preferences rearrange

manongjohn:
Unfortunately, I have to agree that I don't think word wrapping the label is the best way to address the issue.  I would certainly be in favor of shortening labels where possible, when the label is on the left of the controls.  Some reorganization is probably necessary also.

PR #570, opened 2021-01-31
Fix multiple levels using the same file

### LINKED ISSUE #562: Naming issue causes missing levels
**Describe the bug**
If you use the same name on 2 different levels of the same type (but capitalization is different), upon reloading the scene 1 of the levels will be missing.

**To Reproduce**
1. Create a raster level and name it "background" (or etc.)
2. Draw on the level
3. Create a 2nd raster level and name it "Background"
4. Draw on the 2nd level
5. Save All and quit Tahoma
6. Reopen Tahoma and only one of the levels will be there - the other will display as missing

**Screenshots**
Screen Shot 2021-01-21 at 11 06 33 AM
Screen Shot 2021-01-21 at 11 07 03 AM

**Device Information:**
 - OS: macOS Mojave & Catalina
 - Tahoma2D Version: 1.1

manongjohn:
This PR fixes #562 

Level names as file names is a problem on systems where OS file names are case insensitive. Certain checks in Tahoma don't consider when doing file name comparisons. This can result in multiple levels writing to the same file name overwriting each other.

Modified the following changes which also apply to Linux in case the project is shared with Windows/OSX users:

1. General change
   - When performing TFilePath comparisons, a lowercase string comparison is used for OSX/Linux. It was already doing this for Windows.
   - When checking for files on the system, if the exact case filename is not found, it switches to lowercase matching.

2. Create New Level
   - Enhanced logic to force lowercase file name comparisons using the new level name to see if any other level uses a filename with the same name.
   - Modified to no longer automatically create files on the system when creating new Raster and Smart Raster levels. 
 
3. Level Settings
   - Removed logic that gave the option to use the same path as another level. It now blocks it instead.
   - Added warning that unsaved changes on the level will be lost when changing the path.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -316,7 +316,11 @@ bool TFilePath::operator==(const TFilePath &fp) const {
 #ifdef _WIN32
   return _wcsicmp(m_path.c_str(), fp.m_path.c_str()) == 0;
 #else
-  return m_path == fp.m_path;
+  // On case insensitive systems like OSX, we need to
+  // compare using all the same case to confirm it is unique
+  // We'll force this for Linux as well since the project might
+  // be shared on other platforms.
+  return toLower(m_path) == toLower(fp.m_path);
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


artisteacher:
This works well on macOS. Thanks for the fix!

manongjohn:
Thanks for testing!

PR #573, opened 2021-02-02
Fix libmypaint version for Linux builds

### LINKED ISSUE #572: Linux build uses libmypaint 1.3 and needs to be 1.4
The Linux appimage currently includes libmypaint 1.3 libraries but needs to have at least 1.4 otherwise it crashes whenever loading brush sets that contain new brush properties added in 1.4 in support for MyPaint 2.0

This does not impact Windows or macOS builds as they are already using 1.4 and 1.6 respectively.

manongjohn:
This PR fixes #572 

The current Linux build environment is Ubuntu 16.04 and apt repositories only have libmypaint 1.3, but Tahoma2D requires 1.4 or later otherwise Tahoma crashes when opening the Raster tab if brushes with new 1.4 parameters are encountered.

This update will download and compile the source for libmypaint v1.6.1  (latest release) for building.

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -44,6 +44,8 @@ jobs:
           ci-scripts/linux/tahoma-buildopencv.sh
       - name: Build libmypaint
         run: |
+          export CC="ccache ${{ matrix.cc }}"
+          export CXX="ccache ${{ matrix.cxx }}"
           ci-scripts/linux/tahoma-buildlibmypaint.sh
       - name: Build Tahoma2D
         run: |
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -42,6 +42,9 @@ jobs:
           export CC="ccache ${{ matrix.cc }}"
           export CXX="ccache ${{ matrix.cxx }}"
           ci-scripts/linux/tahoma-buildopencv.sh
+      - name: Build libmypaint
+        run: |
+          ci-scripts/linux/tahoma-buildlibmypaint.sh
       - name: Build Tahoma2D
         run: |
           export CC="ccache ${{ matrix.cc }}"
--- a/ci-scripts/linux/tahoma-buildlibmypaint.sh
+++ b/ci-scripts/linux/tahoma-buildlibmypaint.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+cd thirdparty/libmypaint
+
+echo ">>> Cloning libmypaint"
+git clone https://github.com/tahoma2d/libmypaint src
+
+cd src
+echo "*" >| .gitignore
+
+export CFLAGS='-Ofast -ftree-vectorize -fopt-info-vec-optimized -march=native -mtune=native -funsafe-math-optimizations -funsafe-loop-optimizations'
+
+echo ">>> Generating libmypaint environment"
+./autogen.sh
+
+echo ">>> Configuring libmypaint build"
+./configure
+
+echo ">>> Building libmypaint"
+sudo make
+
+echo ">>> Installing libmypaint"
+sudo make install
+
+sudo ldconfig
--- a/ci-scripts/linux/tahoma-buildlibmypaint.sh
+++ b/ci-scripts/linux/tahoma-buildlibmypaint.sh
@@ -13,7 +13,7 @@ echo ">>> Generating libmypaint environment"
 ./autogen.sh
 
 echo ">>> Configuring libmypaint build"
-./configure
+sudo ./configure
 
 echo ">>> Building libmypaint"
 sudo make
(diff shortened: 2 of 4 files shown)


esotericDisciple:
Crashes on Manjaro:
```
bin]$ gdb Tahoma2D
GNU gdb (GDB) 10.1
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from Tahoma2D...

warning: Loadable section ".dynstr" outside of ELF segments
(No debugging symbols found in Tahoma2D)
(gdb) run
Starting program: /Tahoma2D/squashfs-root/usr/bin/Tahoma2D 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
0x00007fffeca391cf in rng_double_set_seed () from /Tahoma2D/squashfs-root/usr/bin/../lib/libmypaint.so.0
(gdb) bt
#0  0x00007fffeca391cf in rng_double_set_seed () at /Tahoma2D/squashfs-root/usr/bin/../lib/libmypaint.so.0
#1  0x00007fffeca395ef in rng_double_new () at /Tahoma2D/squashfs-root/usr/bin/../lib/libmypaint.so.0
#2  0x00007fffeca2f9c1 in mypaint_brush_new_with_buckets () at /Tahoma2D/squashfs-root/usr/bin/../lib/libmypaint.so.0
#3  0x00007ffff23d8987 in TMyPaintBrushStyle::TMyPaintBrushStyle() () at /Tahoma2D/squashfs-root/usr/bin/../lib/libtoonzlib.so
#4  0x00007ffff234ae19 in  () at /Tahoma2D/squashfs-root/usr/bin/../lib/libtoonzlib.so
#5  0x00007ffff7fe12de in call_init.part () at /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff7fe13c8 in _dl_init () at /lib64/ld-linux-x86-64.so.2
#7  0x00007ffff7fd20ca in _dl_start_user () at /lib64/ld-linux-x86-64.so.2
#8  0x0000000000000001 in  ()
#9  0x00007fffffffe0eb in  ()
#10 0x00
```

manongjohn:
Thanks for testing the artifact.  I was also having the same exact error running this artifact on Ubuntu 16.04, which is odd because I use the same scripts and the one I build works fine.  Luckily the one I will be releasing will be from my build environment, not Github Actions.

I do need to look into what is different between this build and my local builds as we should be using the same stuff.

esotericDisciple:
Do both environments have any potentially key differences regarding stuff like whether nividia is the driver for X and/or Wayland or not?

manongjohn:
@esotericDisciple Here is a link to the one I built locally: https://1drv.ms/u/s!Au-akGfpWFmGi_cDkZvHDvLdSrnr3A

See if you can download it and run it. Let me know how it goes.

esotericDisciple:
Thanks, this build runs and doesn't crash when selecting the raster brushes tab for both a smart raster layer and raster layer...

manongjohn:
The only change between this PR and prior PR linux artifacts is that I am now compiling libmypaint 1.4 instead of using apt-get which gives 1.3. 

It might work on my build machine because I originally had libmypaint 1.3 installed.  When I uninstalled it to build 1.4 I left whatever dependencies were installed with 1.3.  Maybe some dependency still remains from 1.3 that allows 1.4 to run properly after I compile it.  I did compare the outputs of compiling locally with what GA showed and I saw no significant difference, so I'm a little confused about this.

manongjohn:
> Thanks, this build runs and doesn't crash when selecting the raster brushes tab for both a smart raster layer and raster layer...

That's good news.  Thanks again.

I won't merge this PR until I figure it out, but at least I know the release build will work.

esotericDisciple:
This latest version of Manjaro (Nibia 20.2.1) uses 1.5(*something), if that helps much (I'm feeling like it wouldn't since the appimage uses the built-in libmypaint lib)...

manongjohn:
Ok.  I discovered why the build wasn't working, though not sure how to fix it.  

I cannot seem to compile libmypaint using clang. Using gcc to compile it is fine, but other things are compiled using clang which appears to be causing the problem at runtime.

For now, I've disabled the clang builds since the gcc builds work fine.  That is also what I have been building with locally and the releases will be compiled with gcc.

manongjohn:
Ok. Now I'm just confused.

I build clang version alone. It crashes. I build gcc version alone it crashes.  I build clang and gcc versions at the same time, gcc version works but the clang version crashes.

manongjohn:
Ok. Linux builds in Github Actions using both gcc and clang appear to be working now.

PR #574, opened 2021-02-05
Hide global set keys when no skeleton created

### LINKED ISSUE #552: Crash when setting keys on a plastic mesh with no skeleton
**Describe the bug**
Adding keyframes from the Viewer's context menu (when a mesh level is active), with the Plastic Tool's animate mode active (but no skeleton is actually created) causes a crash. While part of this is user error, there should be a warning notification that there's no skeleton instead of a crash.

**To Reproduce**
1. Select a level and switch to the plastic tool
2. Create a mesh
3. Switch to the plastic tool's animate mode (without building a skeleton)
4. With the mesh layer still active, right-click in the Viewer & choose "Global Rest Key" (or etc.)

**Crash Report**
[plastic_no_skeleton.txt](https://github.com/tahoma2d/tahoma2d/files/5783428/plastic_no_skeleton.txt)

**Device Information:**
 - OS: macOS
 - Tahoma2D Version: 1.1

manongjohn:
This PR fixes #552 

The `Set Global Key` and `Set Global Rest Key`  can be used when a skeleton vertices are not selected, but does require a skeleton to work or it causes a crash.  Options are now hidden if a skeleton does not exist to prevent usage and the crash.

### DIFF
--- a/toonz/sources/tnztools/plastictool_animate.cpp
+++ b/toonz/sources/tnztools/plastictool_animate.cpp
@@ -188,6 +188,8 @@ void PlasticTool::leftButtonUp_animate(const TPointD &pos,
 void PlasticTool::addContextMenuActions_animate(QMenu *menu) {
   bool ret = true;
 
+  if (m_sd.getPointer() == nullptr) return;
+
   if (!m_svSel.isEmpty()) {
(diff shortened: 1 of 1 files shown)


artisteacher:
This works great on my end.

manongjohn:
Thanks for testing!

PR #578, opened 2021-02-11
Add CrashRpt

manongjohn:
This PR sets up integration for the open source CrashRpt (v1.5) system which will create a mini dump file whenever Tahoma2D crashes.

- CrashRpt In General
   - CrashRpt is only for Windows systems and is not available for macOS or Linux.
   - Source (for v1.5) can be found here: https://github.com/CrashRpt/crashrpt2
   - Documentation (for 1.4.3) can be found here: http://crashrpt.sourceforge.net/docs/html

- CrashRpt Intergration into Tahoma2D
   - CrashRpt will only be available in Windows pre-release and releases compiled locally. It will not be active for PRs.
   - Crash logs and mini dump files will be created in `C:\Users\yourusername\AppData\Local\Tahoma2D\Tahoma2D\cache\crashrpt\`
   - Since the CrashRpt output is created in Tahoma2D's cache folder, it will be cleared when the user select the Tahoma2D menu option `File -> Clear Cache Folder`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,7 +8,6 @@ assignees: ''
 ---
 
 <!-- Please search for existing issues to avoid creating duplicates. -->
-<!-- And please test using the latest nightly build to make sure your issue has not already been fixed: https://github.com/turtletooth/tahoma2d/releases/tag/nightly/ -->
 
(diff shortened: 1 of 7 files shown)


manongjohn:
All checks passed. As there is no artifact to test and only apply to when I build Windows releases, I'm just going to merge this.

PR #583, opened 2021-02-12
Add temp tool hold switch time as preference

manongjohn:
This PR addresses the issue with temporary tool switching where it usually requires a user to hold the shortcut for 500 ms (1/2 second) in order to switch back to the original tool.

There is now a new Tools Preference setting allowing you to change that value from 500 to whatever you want it to be.

image

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -342,6 +342,9 @@ public:
   bool useCtrlAltToResizeBrushEnabled() const {
     return getBoolValue(useCtrlAltToResizeBrush);
   }
+  int getTempToolSwitchtimer() const {
+    return getIntValue(temptoolswitchtimer);
+  }
 
   // Xsheet  tab
   QString getXsheetLayoutPreference() const {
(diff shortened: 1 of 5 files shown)


beeheemooth:
Finally!
I've been waiting for this for four years! https://github.com/opentoonz/opentoonz/issues/1100
 Thank! 
Does it make sense to hard-code navigation buttons now?

manongjohn:
Tahoma's shortcuts for pan/rotate/zoom completely bypass the "Tool" object and work directly against the viewer, so it won't use this setting as it doesn't need to.  For that reason, "space" can not be used in any other shortcut if someone wanted to change things around.  

If you set a different shortcut for pan/rotate/zoom tools, you would have 2 shortcuts, space and whatever you changed it to.  The one you set would be subject to normal tools behavior.  

Personally I'm not too crazy about this behavior and would prefer to change it back so you can set pan/rotate/zoom to whatever you want so space can be used elsewhere if you want.  But that is a separate topic for discussion.

onigiri1:
I have been waiting for this feature, Thank you so much!

beeheemooth:
@manongjohn  Oh yes, you're right ... I completely forgot that navigation tools, when switched, reset the selection of previous tools. 

I think a more reasonable range limit should have been specified than 1 - 999999999 ms (ex: 50 - 5000). It is unlikely that any of the users will hold the button for 12 days.  ^_^

PR #588, opened 2021-02-14
Re-allow space shortcut

manongjohn:
This PR restores the ability to assign/reassign space to shortcuts.  The old pan/zoom/rotate shortcuts have been restored and can be changed per user's preference.

The SceneViewer has it's own version of pan/zoom/rotate commands not tied to the original Tool.  I've somewhat bound those commands to the pan/zoom/rotate Tool version allowing them to work when using the Tool's defined shortcuts. 

The Timeline/Xsheet, Function Editor and Schematic has it's own version of pan command not tied to the original tool. I've also somewhat bound those commands to the pan Tool version allowing them to work when using the Tool's defined shortcuts. 

Things to note:
- When using the pan/zoom/rotate shortcut outside the panes noted above, the normal tool shortcut behavior is in effect where a quick shortcut press makes it a permanent switch, and a long press makes it temporary.


manongjohn:
Ok,  I think I removed all the hardcoded space key panning. They now all use the Hand's shortcut...which is currently...um....space.

melieconiek:
Tested this. I can sometimes get it to loose an active selection (Selection Tool.) The cursor flickers momentarily when this happens, like it switched tools for a second.

manongjohn:
Thanks. I will see if I can duplicate and figure out what's going on.

melieconiek:
Got it.
- Make a selection.
- Hold down space and press the mouse to pan.
- Lift up the mouse first, then spacebar.

Selection will be lost when you lift spacebar.

artisteacher:
I think the reason Jeremy made the current implementation was to allow the active selection to be kept in the Viewer while panning or etc.

manongjohn:
@melieconiek I tried with both pen and mouse and cannot seem to reproduce this.  I don't lose the selection.

Curious if anyone else can easily duplicate this.

manongjohn:
I've retriggered the Linux/macOS builds. Hopefully they'll pick up the fix I just pushed and there will be working artifacts.

melieconiek:
I tried again with the latest artifact, clean, no changes to controls or anything -- I can replicate it perfectly every time. With both a drawing display and mouse, on both vector and raster levels. If you release the controls the other way around it keeps the selection just fine. No idea what's different here.

artisteacher:
I'm still getting a crash on open with this artifact. However, building from master is working properly now. Thanks for fixing that so quickly!

artisteacher:
> Lift up the mouse first, then spacebar.

I'm losing my selections too, but only if I release in that order.

On a related note, CMD+space opens Spotlight search on a mac, while Space+cmd rotates.


> The Timeline/Xsheet, Function Editor and Schematic has it's own version of pan command

It would be helpful if the pan cursor displayed in these panels too

manongjohn:
Hmm.. Steps seem straight forward, yet I can't seem to get it to lose selection. I'll keep trying to duplicate it.

> On a related note, CMD+space opens Spotlight search on a mac, while Space+cmd rotates.

Did order matter before?  I know in some areas of Tahoma, sequence of the keys pressed matters.  Usually modifier keys need to be 1st to be used.  If 2nd, it is usually ignored. I'll see what I can do to fix that but likely not a quick fix so might be something I look at after the next release.

> It would be helpful if the pan cursor displayed in these panels too
Thanks for the suggestion. Will Add

artisteacher:
I think the order did matter before too.

manongjohn:
@melieconiek @artisteacher 

Question regarding when you lift your mouse/pen...does it happen the instant you lift the pen/mouse, meaning if you just stop pressing but are still hovering over where you pressed?  Or are you moving away from the canvas when it happens? Could you be passing over another panel when you lose selection?

I'm assuming when you lose the selection, your tool actually switched to the Hand Tool until you stopped pressing space.

artisteacher:
I'm using a touchpad, but when I release the cursor is still in the Viewer.

artisteacher:
To make this more challenging, the order doesn't seem to matter every time. If it's working properly, it's hard to trigger the problem behavior. When it's not working properly, it's hard to trigger the correct behavior. So there must be some other factor...

When there is a problem, it's not actually the selection itself that seems to be at issue.  The tool isn't returning properly when the keys are released the wrong way. If I release the mouse first and then the spacebar, the tool returns to the next previous tool - ie. the brush tool instead of keeping the selection tool:

1. Draw with the brush tool
2. Switch to the select tool and make a selection
3. Pan with spacebar
4. Release mouse then spacebar = selection lost, back to brush tool

When this is happening, it doesn't happen if I release the spacebar first.

manongjohn:
Even though I could never duplicate the problem, I believe I know what is happening. So let's see if this is any better...

Now, when you use quick-pan/zoom/rotate, the toolbar version of your action will be temporarily disabled.  This should prevent switching to it and causing the behavior you were seeing.  Let me know if it still does it. *fingers crossed*

I also have it so if you are still pressing mouse/pen and activate a quick action, the tool won't switch but the quick action would take over until you release the shortcut. So let's say you are moving/deforming a selection, if you hold quick-pan shortcut, you can pan the canvas. When you let go you are still in moving selection/deforming mode.  Before it would not allow the pan operation until you stopped.

Let me know if you run into any other odd issues regarding quick actions.

Also added pan cursor icon changes timeline/xsheet, schematic, and function editor.

melieconiek:
@manongjohn This seems to have fixed it. Thanks!

artisteacher:
> Also added pan cursor icon changes timeline/xsheet, schematic, and function editor.

Thanks!

Panning or etc. in the Viewer is working fine. 

This is minor, but sometimes when I release the panning behavior, the cursor continues to display the hand for panning until the mouse moves. Then it updates properly. That only seems to happen when the mouse is released first.

I have noticed that panning in the Graph Editor works best if you click in the Graph Editor first, so that it is active. If you don't, you get a selection box rather than a pan behavior. Once that happens, the active tool switches to the pan tool, instead of being a quick action.

manongjohn:
I believe I fixed the panning mouse cursor issue.

As for the graph editor, that was because the Function Editor pane, when docked, didn't automatically gain focus when moving the cursor over it like it does when you move over the Viewer, Timeline or other docked panes.  I fixed the focus issue related to that docked pane so it should help with immediate use of quick actions.

Note: For floating panes, you still need to click that pane first to actually gain focus before trying to use the quick action.

artisteacher:
Everything is working great now on macOS.

manongjohn:
Thanks for verifying!

PR #592, opened 2021-02-15
Statusbar enhancements

manongjohn:
This PR enhances the statusbar as follows

1. Added Hover Hint displaying the same text as the status in case text doesn't fully display on smaller screens.

2. Added modifier key information for Selection tool.
image

3. Added logic to display macOS modifier key symbols instead of the text.
image

4. Cleaned up unecessary spacing in scene section + added separator '|' between command and scene sections
image

5. Command info text converted to translatable strings.

Having issues eliding (cutting and adding "...") text that doesn't fit into status bar completely.  Will submit that separately, but probably won't make upcoming release.

### DIFF
--- a/toonz/sources/include/toonz/tapplication.h
+++ b/toonz/sources/include/toonz/tapplication.h
@@ -20,6 +20,7 @@ class TPaletteHandle;
 class TFxHandle;
 class PaletteController;
 class TColorStyle;
+class StatusBar;
 
 //====================================================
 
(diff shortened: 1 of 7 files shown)


artisteacher:
This is really helpful! 

I tried lowering the resolution of my 15" screen to try to approximate the experience on a 13". If I lower the resolution too much, this is what happens:
Screen Shot 2021-02-15 at 4 40 27 PM
All of the shortcuts are still visible, it's just not necessarily easy to read.


A couple modes of the geometric tool could also use additional status messages:
Rectangle/Ellipse:
- Shift = proportion lock
- Alt/Opt = create from center

Polyline:
- Click+drag = create curve
- Ctrl/Cmd = return to straight line
- Shift = snap to angle

manongjohn:
Funny as I was submitting this I was think about replacing spacing with hard returns for hover hint. Let me see what I can do about that.

artisteacher:
Much better!

PR #593, opened 2021-02-15
Change shortcut usage and navigation while renaming cell

### LINKED ISSUE #484: Note level issues
**Describe the bug**
Editing the text of a note level is very finicky. Using the arrows changes cell focus instead of moving the cursor. Typing an apostrophe ' deletes all of the prior text. The text box is restricted to the width of the zoomed in cell, so it's hard to see what you are doing when typing longer notes. To correct an error, you have to click and drag to change the visible area of text by selecting. It's easy to overshoot.

**To Reproduce**
1. Create a note level
2. Type a sentence into a note cell & press enter
3. Double-click on the cell to edit the existing text
4. Try to navigate the text with the arrow keys
5. Double-click on the cell to edit the text again
6. Click at the end of the text and continue typing - make sure to add a contraction - ie. don't

**Expected behavior**
For arrow keys to move the cursor when editing note text. For the apostrophe to make an apostrophe, not delete prior text.

What would be ideal, is for a small text editor panel to open after double-clicking on a note cell to edit a note. Then it would be easy to see what is being typed, and you could use the cursor to navigate more easily. There could be arrow buttons to move forwards or backwards along the note level.

**Device Information:**
 - OS: macOS Catalina
 - Tahoma2D Version: 1.1

manongjohn:
This PR fixes #484 

I've modified how shortcuts and navigation is handled while renaming cells as follows:

1. If the column is a Note column or when `Enable Tahoma2D Commands' Shortcut Keys While Renaming Cells` is OFF (default), no Tahoma configured shortcuts will work.
   - If not a Note column and settings is ON, Tahoma shortcuts apply except line editing standard commands: Copy, Cut, Paste, Undo, Redo, Clear
  
2.  LEFT/RIGHT navigation arrows within a rename edit box will move between characters.
   - If you move LEFT past the 1st character, you will move to the cell to the left without renaming cell.
   - If you move RIGHT past the last character, you will move to the cell to the right without renaming cell

3. Added TAB and BACKTAB (Shift+TAB) navigation to move down and up the column, respectively without renaming cell.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -74,6 +74,7 @@
 #include <QToolTip>
 #include <QApplication>
 #include <QClipboard>
+#include <QKeyEvent>
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


artisteacher:
This is so much better!

PR #597, opened 2021-02-22
Fix loading scenes from Story Planner

manongjohn:
This PR was primarily created to fix an issue when trying to load a scene generated by Digital Video's Story Planner.

When Story Planner generates a scene for OT/Tahoma...

1. It generates image files with the naming convention Scene_<scene#>\_<panel#>\_<layer#>.png (i.e. Scene_001_001_0.png).
2. It generates a scene file storing the full name of the image file but uses it as a single image file.

Unfortunately, when Tahoma reads the scene file, the image filenames are treated as sequenced filenames, regardless of how it is used. This causes it to have trouble finding the file because it isn't expecting a sequenced filename to be used as a single image file.

Added logic when loading a level, if we are currently loading a scene, use the path stored in the file as is instead of auto-converting to level name convention.  Treat all exact file name matches, in this case, as Single Image files (no frames), regardless if the file has the sequence file naming convention.  In all other cases of loading, auto-convert to level naming convention and look for files matching the sequenced file name.

### DIFF
--- a/toonz/sources/common/timage_io/tlevel_io.cpp
+++ b/toonz/sources/common/timage_io/tlevel_io.cpp
@@ -35,7 +35,8 @@ TLevelReader::TLevelReader(const TFilePath &path)
     , m_info(0)
     , m_path(path)
     , m_contentHistory(0)
-    , m_frameFormat(TFrameId::FOUR_ZEROS) {}
+    , m_frameFormat(TFrameId::FOUR_ZEROS)
+    , m_useExactPath(false) {}
 
 //-----------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


artisteacher:
On a related note, Tahoma (and OT) can't properly load screenshots on macOS. I assume it's due to the naming convention - ie. "Screen Shot 2021-03-11 at 6.28.27 PM.png". The image acts like it's loading but an icon's not generated and the image doesn't display. If you rename the image to remove the periods, it loads properly.

manongjohn:
Ok.  I think I got this working the way it needs to work.  Would/should only impact loading levels when opening a scene in T2D.  Changes shouldn't impact loading levels via other means.

I did test loading a simple scene I created in the trial version of Digital Video's Story Planner (https://www.toonzpremium.com/downloadsp/) as well as my usual test scene and various other scenes and didn't find any obvious loading issues.

PR #612, opened 2021-03-11
Block new frames in Single Frame levels

manongjohn:
This change will prevent the creation of new frames on Single Frame Levels.

"Single Frame" levels are raster files that contain only 1 image and whose filename is not in a sequenced file format.  Because of this, a frame number is never assigned to the image. For example ABC.png would be considered a Single Frame Level and noted as such in the Level Strip.

We need to block the creation of new frames on these types of levels because when it saves, the file is not renamed to account for multiple frames. What usually winds up happening is only 1 image is saved to the original file name and the rest are discarded, resulting in loss of drawings.

I've modified the following when it comes to Single Frame levels:
- Level Strip: Prevent any navigation to the blank areas where you would normally click to create a new frame.
- Duplicate Drawing command: Blocked the command and give warning
- Create Blank Frame command: Blocked the command and give warning
- Tools: Blocks the automatic creation of new frames. Warns when attempting to us a tool on empty frames that could normally cause new frames to be created. Updated cursor to show a tool cannot be used when on an empty frame.

### DIFF
--- a/toonz/sources/tnztools/tool.cpp
+++ b/toonz/sources/tnztools/tool.cpp
@@ -350,6 +350,13 @@ TImage *TTool::touchImage() {
   TXshSimpleLevel *sl = cell.getSimpleLevel();
 
   if (sl) {
+    // For Single Frame levels, don't create anything
+    std::vector<TFrameId> fids;
+    sl->getFids(fids);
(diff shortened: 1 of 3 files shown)


artisteacher:
So this is really just for imported/loaded images that aren't named in an accepted sequential format?

It works exactly as described on macOS.

manongjohn:
Yes, it ony applies to imported/loaded images that would load into a raster level.

PR #615, opened 2021-03-15
Update external app handling

manongjohn:
The following changes are included in this PR:

- Rhubarb Lip Sync
  - Added Rhubarb directory and processing timeout settings to Preferences
  - Improved autodetection of rhubarb executable path. Stores as relative path.
  - Allow `0` timeout value = no timeout (default). Means it waits until rhubarb finishes or errors out. 
  - Display rhubarb processing errors in a popup
  - Added Rhubarb Lip Sync license
  - Updated packaging scripts to include Rhubarb executable from https://github.com/tahoma2d/rhubarb-lip-sync repo
  - NOTE: Tahoma2D uses a modified version of the official Rhubarb application.

- Ffmpeg
  - Improved autodetection of ffmpeg executable path. Stores as relative path.
  - Allow `0` timeout value = no timeout (default). Means it waits until ffmpeg finishes or errors out. 
  - Updated packaging scripts to include ffmpeg executable from https://github.com/tahoma2d/ffmpeg repo

- Preferences
  - Moved `Fast Render Path` to `Savings` page
  - Renamed `Preferences` -> `Import/Export` to `3rd Party Apps`
  - Updated/Simplified wording for ffmpeg and rhubarb preferences

image

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -48,6 +48,9 @@ jobs:
           export CC="ccache ${{ matrix.cc }}"
           export CXX="ccache ${{ matrix.cxx }}"
           ci-scripts/linux/tahoma-build.sh
+      - name: Get 3rd Party Apps
+        run: |
+          ci-scripts/linux/tahoma-get3rdpartyapps.sh
       - name: Create Package
         run: |
           ci-scripts/linux/tahoma-buildpkg.sh
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -27,7 +27,6 @@ jobs:
             /home/runner/.ccache
             /home/runner/work/tahoma2d/taoma2d/thirdparty/openh264
             /home/runner/work/tahoma2d/taoma2d/thirdparty/ffmpeg
-            /home/runner/work/tahoma2d/taoma2d/thirdparty/ffmpeg_shared
             /home/runner/work/tahoma2d/taoma2d/thirdparty/opencv
           key: ${{ runner.os }}-${{ matrix.compiler }}-${{ github.sha }}
           restore-keys: |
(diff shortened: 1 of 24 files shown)


DarrenTAnims:
Excellent. It's great to have this in here. And I think that the new renamed category makes more sense.
1 thing however. Having -1 as the "no timeout" is very code-centric. Wouldn't 0 make more sense to the layman? Then you almost wouldn't need a comment.

manongjohn:
@DarrenTAnims Good point.  I'll change that so that if it is 0, it will convert it to -1 for internal processing.  Right now if you put in 0, the default 30000 msec (30 sec) is used.

manongjohn:
So I've removed the -1 message changed it so 0 now turns off the timeouts.  Is that enough to suggest no timeout or should I really put in a checkbox to enable/disable timeout and a box for the timeout value if enabled?

artisteacher:
I mean, I think 0 is pretty clear but you can't get clearer than the checkbox...

DarrenTAnims:
For having a simpler screen, I don't think you need any text. Plus there is the ? at the top of the window, so I'd prefer to be sure that clicking this shows the help page, preferably pointing to the correct page and there are details of these options on that page.

manongjohn:
Finally got the handling of 3rd party apps more organized and somewhat standardized.

All artifacts should now include/use ffmpeg 4.3.1 (LGPL), opencv 4.51 and rhubarb 1.10.2 (custom). 

This does naturally increase the size of the downloads.  Windows release builds will be slightly bigger once the CrashRpt stuff is included, since I don't include in PRs.  Should I?

If anyone feels this isn't a good idea for PRs, let me know and I'll only include them in the master commit builds, which are kind of like nightlies.

artisteacher:
I don't use Windows, but I imagine having crash data could help with troubleshooting if a PR has an issue on someone's system.

PR #616, opened 2021-03-15
mac specific fixes from OT

### LINKED ISSUE #97: Task and command line rendering issues

### LINKED ISSUE #158: Record audio crash

### LINKED ISSUE #460: Inbetween text squashed in level strip
**Screenshots**
Screen Shot 2020-11-04 at 9 19 34 PM

**Device Information:**
 - OS: macOS
 - Tahoma2D Version: 1.1

artisteacher:
By @shun-iwasawa https://github.com/opentoonz/opentoonz/pull/3763 - this fixes #158
https://github.com/opentoonz/opentoonz/pull/3767 - this fixes #97 
https://github.com/opentoonz/opentoonz/pull/3766 - this fixes #460 

For audio recording, we already have info.plist updated properly, so we just need the sample rate condition removed for macOS. This also makes Task rendering work and fixes the display of the Inbetween button.

### DIFF
--- a/toonz/sources/tcleanupper/tcleanupper.cpp
+++ b/toonz/sources/tcleanupper/tcleanupper.cpp
@@ -491,8 +491,9 @@ int main(int argc, char *argv[]) {
   // questo definisce la registry root e inizializza TEnv
   TEnv::setRootVarName(rootVarName);
   TEnv::setSystemVarPrefix(systemVarPrefix);
+  TEnv::setApplicationFileName(argv[0]);
 
-  QCoreApplication::setOrganizationName("OpenToonz");
+  QCoreApplication::setOrganizationName("Tahoma2D");
   QCoreApplication::setOrganizationDomain("");
   QCoreApplication::setApplicationName(
       QString::fromStdString(TEnv::getApplicationName()));
--- a/toonz/sources/tcomposer/tcomposer.cpp
+++ b/toonz/sources/tcomposer/tcomposer.cpp
@@ -663,6 +663,7 @@ int main(int argc, char *argv[]) {
   // questo definisce la registry root e inizializza TEnv
   TEnv::setRootVarName(rootVarName);
   TEnv::setSystemVarPrefix(systemVarPrefix);
+  TEnv::setApplicationFileName(argv[0]);
 
   QCoreApplication::setOrganizationName("Tahoma2D");
   QCoreApplication::setOrganizationDomain("");
--- a/toonz/sources/tconverter/tconverter.cpp
+++ b/toonz/sources/tconverter/tconverter.cpp
@@ -177,7 +177,7 @@ void convertFromVI(const TLevelReaderP &lr, const TPaletteP &plt,
   }
   maxBbox = maxBbox.enlarge(2);
   if (width)  // calcolo l'affine
-    aff   = TScale((double)width / maxBbox.getLx());
+    aff = TScale((double)width / maxBbox.getLx());
   maxBbox = aff * maxBbox;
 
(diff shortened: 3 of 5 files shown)


manongjohn:
Thanks for following up and porting over the fix. 

Shun didn't mention why the fix works so I wonder why turning off the setSampleRate fixes this for macOS but the original value is ok for Windows.  Nothing mentioned in QT's on-line docs and I found no QT bugs related to it, yet it seems like a macOS bug in QT 5.15 (maybe earlier).  So long as the -1 doesn't affect the recording, I guess it's ok.

As you've already tested this fix in OT and it's was a pretty straightforward port, I'll merge it when the build is done. Odd that the Github Action builds didn't kick off with your PR.  I'll have to check why it didn't.

artisteacher:
No problem. I'm excited to have this fixed before the new release!

I did do a test build using the build scripts to make sure that it worked with our settings. There's finally a prompt to grant permission to use the microphone. It starts recording as soon as permission is granted! 

Speaking of the build scripts, we should probably remove the minimum deployment target from tahoma_build.sh for macOS since it doesn't actually work for the libraries.

manongjohn:
Thanks again for bringing the other fixes over.

Funny, i was working on the inbetween text bug and trying to do what he did to fix it, but was having trouble with it so I set it aside for next time. I was so close! Lol...anyway, glad it's fixed.

artisteacher:
> Shun didn't mention why the fix works so I wonder why turning off the setSampleRate fixes this for macOS but the original value is ok for Windows.

It looks like the original setSampleRate value is an issue in Windows too for OT with the updates to the build environment there.

PR #618, opened 2021-03-16
Minor room updates

artisteacher:
I find it helpful for my students to include a command bar in the 2d room and to have a history room for easy undo/redo access. 
Screen Shot 2021-03-15 at 8 19 03 PM
Screen Shot 2021-03-15 at 8 19 26 PM
Anyways, feel free to ignore this if it's not the direction you want to go in.


manongjohn:
I'm fine with adding the command bar to the 2D room. I usually add a command bar to that room anyway. Saves me a step when setting up. 😁 I usually put mine above the viewer since i usually also add the timeline's Quick toolbar. I use the command bar for stuff related to viewer and a quick toolbar for timeline stuff. I'm ok with leaving it below the viewer.

I do understand why you are adding a History room, bit I'm a little iffy about dedicating a whole room to it.

Would like to get opinion from others though.

artisteacher:
Would you add anything else with the history panel? Or put it elsewhere?

manongjohn:
Not sure how others use it other than to see what was done during the session. I know it may be helpful for jumping back to a specific action and undoing everything after that instead of using  ctrl-z repeatedly.

I can't really see adding it to an existing room. To me its something you open up once in a while when you really need to see when something was done, but does that warrant its own room? It is faster to get to it that way than to use menus or shortcut I guess, plus you are aware it is there.

Still mulling over the pros and cons of it.

artisteacher:
Yeah, it's really just about quick access and awareness. But I can remove it if you want.

manongjohn:
Since there isn't much discussion about the proposed History room on here or on the Discord, I'll take it as it's not really a big deal either way...which it really isn't.

I'll incorporate as is.

Thanks for the improvement @artisteacher

PR #619, opened 2021-03-16
Fix static macOS ffmpeg builds

### LINKED ISSUE #617: Ffmpeg crash on open from Actions Artifact
**Describe the bug**
While I was looking into a different crash, I noticed that Ffmpeg is generating 3 crash reports when opening Tahoma2D artifact builds. If you try to use FFmpeg (ie. do a fast render to mp4), there's a notification saying FFmpeg is not found - even though the executables seem to be in the app.

**To Reproduce**
1. Download a recent macOS artifact build
2. Open the build
3. Look in Console for the crash reports

If applicable, add screenshots or links to short videos to help explain your problem or provide logs or dmp files if it's a crash.

Screen Shot 2021-03-15 at 6 01 54 PM
Screen Shot 2021-03-15 at 6 21 58 PM

**Device Information:**
 - OS: macOS Big Sur
 - Tahoma2D Version: Tested on action artifacts for #610 and #615

manongjohn:
This PR fixes #617 by disabling libxcb library inclusion when building the static ffmpeg version for macOS.

Additionally, Homebrew of aom generates a shared library which isn't good for the static ffmpeg build.  Now building static aom from source for use in both static and shared ffmpeg builds

### DIFF
--- a/ci-scripts/osx/tahoma-buildffmpeg.sh
+++ b/ci-scripts/osx/tahoma-buildffmpeg.sh
@@ -1,5 +1,29 @@
 cd thirdparty
 
+echo ">>> Cloning aom"
+mkdir aom
+cd aom
+echo "*" >| .gitignore
+
+git clone https://aomedia.googlesource.com/aom
+
+if [ ! -d build ]
+then
+   mkdir build
+fi
+cd build
+
+echo ">>> CMaking aom"
+cmake ../aom
+
+echo ">>> Making aom"
+make
+
+echo ">>> Installing aom"
+sudo make install
+
+cd ../..
+
 echo ">>> Cloning ffmpeg"
 git clone https://git.ffmpeg.org/ffmpeg.git ffmpeg
 
--- a/ci-scripts/osx/tahoma-buildffmpeg.sh
+++ b/ci-scripts/osx/tahoma-buildffmpeg.sh
@@ -58,7 +58,6 @@ export SDKROOT=`xcrun --show-sdk-path`
       --enable-static \
       --disable-shared \
       --enable-pic \
-      --disable-shared \
       --enable-pthreads \
       --enable-version3 \
       --enable-videotoolbox \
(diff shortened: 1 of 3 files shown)


artisteacher:
I can confirm macOS artifact builds no longer have ffmpeg crashes on open.

manongjohn:
Hmm. When i bring it to my non build vm, it yells about /usr/local/opt/aom/lib/libaom.2.dylib.

Is that on your system already?

artisteacher:
Yes, it is.

manongjohn:
I don't think thats in a normal system. Possible it was placed there if you brew installed stuff to compile something related to OT or Tahoma.

Seems the homebrew formula changed and  doesn't allow static linking of aom anymore impacting when I build ffmpeg from source in Actions. My local build is old enough that it can still link statically.

Deciding what to do. I may pull the compiled version from tahoma2d/ffmpeg repo when building through Actions. Need to first figure out what I am doing with opencv. I may be compiling it for no reason when I can just use a brew'd version.

artisteacher:
Jeremy mentioned something to me about a license issue with libx264 which was a dependency for either opencv or ffmpeg - likely why he compiled his own.

manongjohn:
Oh right, x264 is needed by ffmpeg if compiled with GPL options.

Homebrew opencv is/was BSD-3. I think if you brew install opencv, it was a version compiled without ffmpeg support. You had to specify to install opencv with ffmpeg (GPL version) included.

I don't think we need ffmpeg compiled into opencv unless that is somehow needed for the Canon stuff? 

My head hurts...

manongjohn:
Nevermind...I figured it out.

The Windows version of opencv is compiled with LGPL verison of ffmpeg, but the Homebrew opencv version uses the GPL version, hence needing to compile it ourselves.

So back to square 1 with this.

manongjohn:
I've confirmed ffmpeg in macOS builds isn't trying to use any shared libraries anymore and successfully loaded a scene with an mp3 with it.

PR #621, opened 2021-03-19
Setup ccache for Action builds

manongjohn:
This is to set up Actions to use ccache to hopefully speed up macOS and Linux builds.


manongjohn:
Builds now using ccache.  Merging so the master branch ccache can be built.

artisteacher:
That is so much faster!

PR #633, opened 2021-03-28
Fix console warning messages

manongjohn:
This PR fixes the issues that causes certain console warnings to be generated when running from command line.

The following console warnings were addressed

- `QLayout: Attempting to add QLayout "" to xxxx "", which already has a layout`

Generated by: Main window at startup as well as the following popups: About, Antialias, Record Audio, Export Scene, Level Settings, Color Fade, Apply Lip Sync to Column, Stop Motion Control, Output Settings, Time Stretch, New/Switch Project,  Timeline Note Viewer

Caused by trying to see parent layout with QHBoxLayout(this) or QVBoxLayout(this) when the parent widget already has a layout.

Cleared `this` from these Layout constructors for the impacted popup.

- `QLayout: Attempting to add QLayout "" to TPanelTitleBarButtonForGrids "", which already has a layout`

Generated by: Anything with buttons in the panel's title bar, like the Viewer, ComboViewer

Caused by trying to see parent layout with QGridLayout when the parent widget already has a layout.

Cleared `this` from the Layout constructors for this object.

- `QSettings::setValue: Empty key passed`

Generated by: About popup. 

Caused by improper setup of DVGui:Dialog for the About popup as well as the DVGui:Dialog not checking to see if the key is empty before saving when hiding the popup.

Corrected About popup constructor to pass the proper parameters as well as skip saving QSettings if they key was not populated.

- `QFileSystemWatcher::addPath: path is empty` and `QFileSystemWatcher::removePath: path is empty`

Generated by: Browser room

Caused by trying to set a file watcher on an empty path when expanding/collapsing the file tree view.  Was triggered when a folder name started with a +.

Added logic to skip paths that are empty or null, in case it finds any.

### DIFF
--- a/toonz/sources/toonz/aboutpopup.cpp
+++ b/toonz/sources/toonz/aboutpopup.cpp
@@ -21,7 +21,7 @@ void AboutClickableLabel::mousePressEvent(QMouseEvent* event) {
 }
 
 AboutPopup::AboutPopup(QWidget* parent)
-    : DVGui::Dialog(parent, true, "About Tahoma2D") {
+    : DVGui::Dialog(parent, true, true, "About Tahoma2D") {
   setFixedWidth(360);
   setFixedHeight(350);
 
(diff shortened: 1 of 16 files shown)


artisteacher:
Thanks for working on some of the warnings! There have been an excessive number for a long time.

PR #639, opened 2021-04-01
Display file paths in Scene Cast

manongjohn:
This PR adds the path column to the Scene Cast to make it easy to see where all your assets are, especially if you `Load` assets instead of `Import` and/or if the level name differs from the actual file name.  Path will also be in the Export File List

image

### DIFF
--- a/toonz/sources/toonz/castviewer.cpp
+++ b/toonz/sources/toonz/castviewer.cpp
@@ -18,6 +18,8 @@
 #include "toonz/levelset.h"
 #include "toonz/toonzscene.h"
 #include "toonz/txshsimplelevel.h"
+#include "toonz/txshpalettelevel.h"
+#include "toonz/txshsoundlevel.h"
 #include "toonz/txshleveltypes.h"
 
 // TnzQt includes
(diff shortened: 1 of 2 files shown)


DarrenTAnims:
This looks really useful. And will be helpful to help diagnose issues in the future.

artisteacher:
@manongjohn - the path is abbreviated on macOS even though there should be plenty of room to see the full thing. 
Screen Shot 2021-04-05 at 11 57 25 PM

manongjohn:
I believe that is the normal behavior for displaying long text in narrow columns. If you expand the column to the right, it should reveal more and more of the filename/path.

artisteacher:
That works. Though what's the point of that last column? Could the path column take up the empty space instead?

manongjohn:
There are only 3 defined column, the rest is space the last column can expand into. Looking briefly it doesn't appear I have control over that. Seems to be the nature of that grid.  The last column width is set to 300. I could make it bigger if you want though likely it will always never be big enough to hold the full path on initial load since paths will vary.

artisteacher:
Yes, giving the 3rd column a larger initial width would be helpful.

PR #640, opened 2021-04-03
Fix missing Linux event mouseMove with tabletMove

### LINKED ISSUE #629: Geometry tool (Arc, MultiArc) graphic tablet - Linux
I originally opened this bug issue in opentoonz, but as it affects Tahoma2d as well, I believe it is necessary to put it here as each project follows a direction. 

In short, the arc and multiarc tool doesn't work correctly when drawing and  adjusting the angle of the lines with graphics tablet.

Here the link to issue https://github.com/opentoonz/opentoonz/issues/3826 if necessary I can copy paste here too.

manongjohn:
This PR fixes #629 

On Linux, a mouseMove event is not generated with tabletMove operations when pen is hovering like it does on other platforms.

I found I also had issues ending the Polyline because of this.

Modified to handle mouse move logic when tabletMove is processed for Linux systems only.

NOTE: While this fix is primarily for Linux systems, one minor change I made may impact Windows/macOS.  When processing mouse/tablet moves where a button is not pressed, internal information is reset with a call to doQuit().  I didn't see any issues in my Windows testing.


andeon:
@manongjohn Thanks, It's working now! I gonna try to test on Windows to see if there is any impact.

manongjohn:
@andeon Thanks for testing

artisteacher:
Seems fine on the macOS end.

PR #641, opened 2021-04-03
Fix loading files with multiple dots

manongjohn:
This PR fixes loading files with multiple dots in the name (i.e "abc def.ghi jkl.png") but are not sequenced files.


artisteacher:
Thanks! Screenshots on macOS load perfectly now.

PR #643, opened 2021-04-04
Disable menu icons

manongjohn:
Decided to disable menu icons for the following reasons

1. Menu icons have been hidden since the beginning. Users are probably used to it. I don't think anyone misses them.
2. Menus look cleaner without them in my opinion
3. Menus are easier to read since it is not as busy with every item having an icon
4. Won't have to deal with icon alignment issues.

### DIFF
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -309,7 +309,7 @@ QMenu {
 }
 QMenu::item {
   border: 0;
-  padding: 3 28 3 40;
+  padding: 3 28;
 }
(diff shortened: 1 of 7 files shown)


artisteacher:
I'm all for this. The menu its super busy with the icons. Also, the icons can have display issues with the native mac menus, depending on light vs. dark mode and white vs. black icons.

RodneyBaker:
A post mortem note relating to this PR.
In Opentoonz, the menu icons have been made optional via Preferences.
REF:  https://github.com/opentoonz/opentoonz/pull/3842

PR #654, opened 2021-04-07
Fix LNP generating tails

manongjohn:
This PR fixes an issue with LNP that occurs when you lift the pen causing the stroke to generate a tail between the desired end stroke position and the position of the pen.

When the pen is lifted, the releases event tells the tool to complete operation using the position of the pen on the canvas instead of where the last drawn position was resulting in the stroke catching up to the pen position.

Changed to complete tool operation using the last known drawn position.

Regression testing: I have tested this without using LNP on Windows and Linux (via VM) and did not encounter any particular issues. 

I am unable to test this change on macOS VM.  My VM doesn't quite recognize my tablet correctly and treats all events as Mouse events. Would be appreciated if someone could test this on macOS using a pen.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -989,8 +989,7 @@ void SceneViewer::onRelease(const TMouseEvent &event) {
   tool->setViewer(this);
 
   {
-    TPointD pos = tool->getMatrix().inv() *
-                  winToWorld(event.mousePos() * getDevPixRatio());
+    TPointD pos = tool->getMatrix().inv() * winToWorld(m_lastMousePos);
 
     TObjectHandle *objHandle = TApp::instance()->getCurrentObject();
     if (tool->getToolType() & TTool::LevelTool && !objHandle->isSpline()) {


artisteacher:
I'll check with a Wacom Intuos on Big Sur this evening, though I don't have LNP.

manongjohn:
@artisteacher Thanks. LNP is for Windows only, I believe.  

Just need to make sure the little change I made doesn't cause a big issue in macOS builds when using a pen. I don't think it will but macs have proven me wrong a number of times. 🙂

artisteacher:
In a quick test it seemed to behave the same as usual with intuos on macOS. 

I did notice that occasionally multi-arc makes a straight line segment. I checked and it looks like this was already happening in 1.1.

Arcs and multi-arcs have an even worse problem using an iPad through Sidecar as an extended display. You need long taps with the stylus to create the points of the arcs and even with that there are occasional straight lines. With a standard or fast tap length, straight lines are a much more common result. I'm thinking that it might be that the iPad might not be recognized as a touch device. Even with touch enabled, the pan/zoom gestures don't work. The two finger pan gesture causes zooming instead of panning. So it might be trying to use the mouse logic for arcs/multi arc, which doesn't work well.

melieconiek:
Fix works on my end, thank you for all the fixes, @manongjohn. :D

manongjohn:
@artisteacher The viewer kind of uses a combination of tablet and mouse events at times which is probably the problem. I don't like how the logic uses both since a mouse event isn't always generated with a tablet event on some systems. This causes some of the oddness we see between the different systems.  One of these days I'd like to review and redo that whole logic.

artisteacher:
>  One of these days I'd like to review and redo that whole logic.

That would be fantastic. I'll be happy to test on the mac end when you do.

PR #675, opened 2021-04-21
Support MOV formats through ffmpeg

manongjohn:
This PR provides support for importing and rendering MOV formats using ffmpeg.

As a test I imported an 1920x1080 30fps 900 frame MOV file and rendering it back out at 24fps. Quality wise, from a quick review, it looked the same or extremely close, but perhaps someone with a better eye for it can see differences and determine if it's ok or not.

I made it so MOV will render with a transparent background.

For rendering MOV, it currently uses the same ffmpeg command line settings as MP4, which isn't much (-pix_fmt yuv420p).  If someone feels this isn't good enough and can give me better ffmpeg parameters for rendering from PNG to MOV, let me know.

Even though 1.2 is in BETA mode, as MOV is still a popular format, I'm willing to include this in 1.2 if the output is good enough or can be improved by May 1st.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -709,7 +709,7 @@ TFrameId TFilePath::getFrame() const {
 
 bool TFilePath::isFfmpegType() const {
   QString type = QString::fromStdString(getType()).toLower();
-  if (type == "gif" || type == "mp4" || type == "webm")
+  if (type == "gif" || type == "mp4" || type == "webm" || type == "mov")
     return true;
   else
     return false;
--- a/toonz/sources/image/CMakeLists.txt
+++ b/toonz/sources/image/CMakeLists.txt
@@ -21,6 +21,7 @@ set(HEADERS
     ffmpeg/tiio_gif.h
     ffmpeg/tiio_webm.h
     ffmpeg/tiio_mp4.h
+    ffmpeg/tiio_mov.h
     ffmpeg/tiio_ffmpeg.h
     sprite/tiio_sprite.h
     mesh/tiio_mesh.h
@@ -50,6 +51,7 @@ set(SOURCES
     ffmpeg/tiio_gif.cpp
     ffmpeg/tiio_webm.cpp
     ffmpeg/tiio_mp4.cpp
+    ffmpeg/tiio_mov.cpp
     ffmpeg/tiio_ffmpeg.cpp
     sprite/tiio_sprite.cpp
     mesh/tiio_mesh.cpp
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -474,6 +474,11 @@ void Ffmpeg::getFramesFromMovie(int frame) {
     // frameArgs << "-accurate_seek";
     // frameArgs << "-ss";
     // frameArgs << "0" + QString::number(frameIndex / m_info->m_frameRate);
+    if (m_path.getType() == "webm") {
+      // To load in webm transparency
+      preIFrameArgs << "-vcodec";
+      preIFrameArgs << "libvpx";
+    }
     preIFrameArgs << "-i";
     preIFrameArgs << m_path.getQString();
     postIFrameArgs << "-y";
--- a/toonz/sources/image/ffmpeg/tiio_mov.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_mov.cpp
@@ -85,7 +85,9 @@ TLevelWriterMov::~TLevelWriterMov() {
   preIArgs << QString::number(m_frameRate);
 
   postIArgs << "-pix_fmt";
-  postIArgs << "yuv420p";
+  postIArgs << "rgb24";
+  postIArgs << "-c:v";
+  postIArgs << "png";
   postIArgs << "-s";
   postIArgs << QString::number(outLx) + "x" + QString::number(outLy);
   postIArgs << "-b";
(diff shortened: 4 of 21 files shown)


artisteacher:
I'm not seeing a transparent background on render in macOS. Transparent areas are rendering white. It must be taking the matte color under the saving preference. Otherwise loading and rendering works without major issue.

Loading an mov video and rendering it without making any changes back to mov does result in a larger file size than the original file - even though the Tahoma version doesn't include the original audio.

artisteacher:
Thanks for adding this!

Eventually, heic and avi would also be beneficial. I think avi is already available on Windows, but it would be helpful for anyone working in multiple environments to be able to use all the same formats.

manongjohn:
Oh. I probably incorrectly thought I was setting MOV to have a transparent background but wound up setting it to have a white background. I found a way to get transparency in MOV files through ffmpeg which i can easily add. Will try it out shortly.

As for MOV coming out bigger? Not sure why. Might be because of using PNG files as intermediary files. Don't know, at this time, if there is a way to shrink that through ffmpeg. Will poke around a little bit and see.

As for AVi, I didnt realize it wasn't an option on macs like it is on Windows. I wonder why that is.

Not sure about HEIC as that requires a specific GPL licensed library compiled into ffmpeg. Though I've replaced it with an LGPL version of it I still need to look more into using that as a format. Can look at it after 1.2 release.

DarrenTAnims:
This is great. It's something I see asked for a lot. So there's no need to install any extra libraries to your system to render to mov? ffmpeg does it all?
As an aside, but related, did you have time to look into rendering to WEBM with a transparent background?

manongjohn:
Yes. Ffmpeg will take care of MOV now. No more need to install an obsolete library anymore. Knoy thing Quicktime gave us was MOV and 3GP. Can add 3GP in the future if desired.

I am fixing transparency for both MOV and WEBM. Will be pushing that change to this PR soon. 🙂

artisteacher:
Webm is definitely transparent. Now I'm seeing a black background for MOV.

manongjohn:
What are you using to verifying transparency on macs?

I know when I render either WEBM or MOV with fully transparent colored (yellow) background, it shows black since that's what my application (OpenShot) shows on transparent backgrounds.  When I use a semi-transparent colored background, I can see the color with the black bleeding through.

manongjohn:
Hmm...I just double checked another way and neither WEBM or MOV are actually showing me a transparent background.  Showing as black as you are seeing.  Will continue to work on it more.

artisteacher:
I checked a few ways. Through Preview - where transparent gif/png show the window background color in transparent areas. I also tried the Quicktime player and the browser 2 ways. For both webm and mov, loading the video directly in a supported browser displays a black background. Adding the video to an HTML page results in the transparent background working for WebM but not mov.

manongjohn:
Ok. I think I figured out what was wrong with the MOV transparency (some copy/paste errors and typos).  I was able to correctly verify the MOV transparency in OpenShot this time, as well as loading the MOV back into Tahoma and checking the background in Preview mode.

Please check the artifact when you get a chance.

I'm having a hard time verifying WEBM transparency on Windows though.  It doesn't show as transparent when I load it in OpenShot or back into Tahoma after generating it.

artisteacher:
Now the Mov file won't play. It says it's not compatible with the Quicktime player.

manongjohn:
FYI, I was able to verify WEBM transparency by embedding it in a test webpage and fixed it to load into Tahoma with transparency.

artisteacher:
The mov file does load back into Tahoma with transparency - it's just not playable natively on mac.

manongjohn:
> Now the Mov file won't play. It says it's not compatible with the Quicktime player.

Ok..might be because I'm encoding it with a different library than before and it's not compatible with QuickTime (is that really still being used?).  Let me try switching it back to what it was using before and verify that transparency still works.

artisteacher:
Yeah, Quicktime still seems to be the native video player.

manongjohn:
I loaded QuickTime 7 on my windows machine and attempted to load the MOV file I created with Tahoma and it opened and played fine.  I also created an MOV file in my macOS VM and it also played fine in QuickTime without any issue, though it went through some conversion process just before playing it.  Not sure why it's having trouble on your machine.

I submitted a change to remove an ffmpeg parameter that might be causing the problem.  I still get transparency without it.  Let me know how it goes.

artisteacher:
It still won't open. It looks like this might be a problem for Catalina & up.

[https://forums.envato.com/t/mov-file-with-alpha-channel-transparency/289077](https://forums.envato.com/t/mov-file-with-alpha-channel-transparency/289077)

manongjohn:
Thanks that helps a bit.

I found parameters based on some searches on-line that allows it to open in Quicktime on Catalina and still appears to provide transparency, though on my VM, all I see is a white screen.  I think it's just my VM being goofy because I took the original PNG files w/ transparency and tried to make a MOV file in Quicktime directly using ProRes encoding and it also resulted in a white screen.

Let me know how it works for you.

artisteacher:
That works perfectly!

manongjohn:
Great! 

This particular codec doesn't seem to work in some of my media viewers in Windows. It does work in VLC and older Quicktime versions so I'm ok with it.

Thanks for helping me figure it out.

artisteacher:
I did notice that the mov and mp4 files rendered from Tahoma won't play in Chrome or Firefox for me. They both play without issue from Safari. Mp4 from other sources will play in Chrome. Mp4 and Mov from other sources will play in Firefox. It's still much better than we were getting before.

I wonder if a couple codecs should be offered in the format options?

manongjohn:
> I wonder if a couple codecs should be offered in the format options?

Yeah. That's a good idea. Crossed my mind after we got mov working. There are other codec that would work if you don't need transparency. Need to probably do some research and some experimentation to see what would be best to provide for any formats going through ffmpeg.

PR #677, opened 2021-04-23
Fix setting up CanonSDK for macOS

manongjohn:
Fixes an issue with macOS master builds with Canon support.


manongjohn:
Merging to see how it builds against master.

artisteacher:
Canon dslr's work without issue with the nightly build on macOS.

PR #679, opened 2021-04-24
Fix nightly release update and windows naming

manongjohn:
Nightlies working (yay!) but need to fix how updating the Release package name for windows.


manongjohn:
Canceled PR builds, going right to merging since the changes only impact master builds

PR #683, opened 2021-05-01
[WIP] Persian translations by @pikhosh

manongjohn:
This PR introduces the start of Persian translations. Translations provided by @pikhosh.

Though not complete, it is a good start and hopefully can be completed by the next version.  Thanks for what you have provided so far @pikhosh!


pikhosh:
There are still problems with the Persian language being RTL; Phrases and sentences are not compatible with the Tahoma2D environment...

manongjohn:
Out of curiosity, what is an example of something that is not compatible for translating? How would it look if you translated like it should be?

pikhosh:
> Out of curiosity, what is an example of something that is not compatible for translating? How would it look if you translated like it should be?

I translated the texts exactly like the English texts and this made them incompatible (If I used the right word) with the UI:
image
image
image

On the other hand, some parts of the UI, the texts are RTL:
image

And the text that is displayed in the status bar when hovering over the buttons; Due to having English words, its word order is wrong:
image

But the text that is displayed when a button is clicked is RTL:
image

I do not know if it is good to have Persian language in version 1.2 or not, because with the problems it has, it is almost useless.

manongjohn:
@pikhosh I just pushed a change that will switch the layout direction to RTL when you choose the Persian language.  Please check out the artifact from this PR once it is done building.

This will allow some stuff to switch to RTL like so

image

However there are some issues with this. Right off the bat, I noticed
1. The icon space in dropdown menus is wide for some reason.  Not sure why nor how to fix that.
2. Preferences layout looks terrible and needs to be reworked (well, for LTR languages, that's true too hehe)

I won't include this in 1.2 so there is more time to complete the translation, and address some of the UI issues.  I will likely need your help identifying where the UI needs to be fixed.

pikhosh:
> Please check out the artifact from this PR once it is done building.

First of all, thank you for your effort!

One of the current problems is that there is a problem in resizing the Xsheet and also the position of the column cells is slightly different from their headers:
Tahoma2D Xsheet_5

And in this section; The "Back" and "Forward" icons must be swapped; The arrows on the right of the image should also point to the left; The position of the folder icons should change; And the file list section should be RTL:
image

And in this section; The word "FPS" is in the wrong position:
image

And in this section; The number of frames is displayed instead of the scene title:
image

And the next problem is that some of the text in the "About Tahoma2D..." window is still LTR:
image

pikhosh:
Hi! Is there any plan to include Persian translation in v1.3? 🤔
Is there any chance that @shun-iwasawa can help with the UI problem?

pikhosh:
> @pikhosh I spent some time, recently, fixing the Xsheet/Timeline header offset you noted previously, but realized that even with this fix, the Xsheet/Timeline isn't good for RTL languages.
> 
> Unfortunately, it is a major effort to modify how the Xsheet/Timeline is drawn and to make it work properly for RTL languages. For this reason, I am sorry to say that I will not be able to support RTL languages easily. :(

No problem at all! Honestly, I really don't care about having a Persian translation anymore; I no longer believe that translating a professional and feature-rich program into a language that is not as up-to-date as English/the source language can be useful...

Anyway, thanks for even taking the time to consider these changes after all this time! And thanks for your hard work and for keeping Tahoma2D alive :)

PR #689, opened 2021-05-07
Fix loading plugins on Linux

manongjohn:
This PR fixes loading Dwango and WolfInABowl plugins on Linux systems.

Currently, even if you manage to build the plugins on a Linux system (in progress!), it will no load because the code is not passing the appropriate parameters to `dlopen()`  on linux environments.  Corrected the parameters and was able to load the plugins.


manongjohn:
Linux users:

I did manage to get the Dwango and WolfInABowl plugins compiled for linux which has been uploaded [here](https://github.com/tahoma2d/tahoma2d_plugins/releases/tag/v1.1).

You'll need this fix to load them.

PR #691, opened 2021-05-07
deb package structure

charliemartinez:
Files for the creation of the standardized deb package compatible with debian, devuan, ubuntu and derivatives, made for the Quirinux GNU / Linux distribution.


charliemartinez:
Hello!

You're right, I think it doesn't make much sense to add this folder to the source code.

Seeing that there is no official deb package, I generated the necessary files for this and organized the structure in such a way that it is enough to download the folder that I shared and enter the command **dpkg -b** followed by the name of the folder so that the .deb is generated.

Actually, the most interesting files here are those inside the "DEBIAN" folder, which can be considered the "source code" of the package deb to be generated with dpkg, however this package will not work correctly if the structure is not the one presented. here _(I shared it because I couldn't think of another way to show it to its full extent)._

There really is not much to modify, so if you are happy to share the .deb I made, I have also published it in this fork, as a release.
[https://github.com/charliemartinez/tahoma2d/releases/tag/v1.2](https://github.com/charliemartinez/tahoma2d/releases/tag/v1.2)

The reason why I made a deb package was to be able to incorporate Tahoma2D into the official Quirinux distribution repositories to be able to install it from software managers or with the apt command. Seeing that my package also works on Debian, Devuan, Ubuntu and Mint, it seemed like a good idea to share it. It is useful for those who make scripts that have the objective of installing several packages at the same time (installing at once with sudo apt-get install tahoma2d and all the programs that they need).

manongjohn:
On behalf of my self and all those that contributed to it, Thanks! :)

Regarding the deb package...I think I have a better understanding of what you were submitting.

So your deb package effectively system installs T2D rather than what I am currently providing, a non-installing portable AppImage.  At some point in the future, I was thinking of also providing installed versions of T2D along with the portable version.  Before I do that, I have some folder relocation work that I wanted to do.

I currently build the T2D linux releases on Ubuntu. I will look more into it, but I believe I can create deb packages on there.  The structure you provided will help me script the creation of the deb package when I am ready to do that.

charliemartinez:
The trick is that my deb doesn't actually "install" the program, but instead puts it in the / opt folder along with all its dependencies. In essence, it's still the same as your appimage, only unzipped. The first time I did this was with a software called StoryBoarder and I found that this greatly improved the performance of the application with respect to its use in appimage. I am building a repository for the Quirinux distribution with several of these debs that I am packing: https://quirinux.ga/main/buster/paquetes/

PR #700, opened 2021-05-20
Enhanced WAV support

manongjohn:
This PR natively enhances WAV support as follows:

- Added support for 32bit fixed and float WAV formats
- Fixes playback quality when loading 24bit fixed WAV formats
- Internally converts 8bit, 16bit, 24bit and 32bit float WAV files to 32bit fixed format
- Processes sound files as 32bit fixed PCM format where applicable.

Caveats:
- I was only able to test on Windows and Linux.  My macOS VM won't recognize my audio.
- I loaded an MP3 for regression testing and didn't have any apparent issues.  I didn't verify with AIF/AIFF or RAW files.
- Sound clipping is still an issue if the audio is still too loud
- Not 100% sure of my fixed <-> float conversions for lower bit formats, but shouldn't be an issue since everything is converted to 32bit fixed. (I didn't have issues converting 32bit float to 16 bit fixed so I think it's ok)

### DIFF
--- a/toonz/sources/common/tsound/tsop.cpp
+++ b/toonz/sources/common/tsound/tsop.cpp
@@ -483,6 +483,20 @@ public:
 
     return TSoundTrackP(dst);
   }
+
+  TSoundTrackP compute(const TSoundTrackMono32 &src) override {
+    TSoundTrackMono32 *dst = resampleT(const_cast<TSoundTrackMono32 &>(src),
+                                       m_sampleRate, m_filterType);
+
+    return TSoundTrackP(dst);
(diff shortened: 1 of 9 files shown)


artisteacher:
This will be great! I'll take a look on macOS as soon as I can. The next couple days are a bit hectic.

artisteacher:
I tested aiff, wav and mp4 with a few different encodings on macOS. Everything loaded without issue & sounded the same.

manongjohn:
@artisteacher Thanks for testing on macs!

PR #701, opened 2021-05-21
Fix macOS shortcut display

manongjohn:
This PR fixes how shortcuts for macOS displays in the following areas:

1. Shortcut popup - replaces Ctrl/Shift/Alt/Meta keys with the macOS key symbol equivalent
2. Status bar - removes the "+" between combo keys in keeping with how macOS displays multi-key shortcuts


artisteacher:
This works great on my end.

PR #706, opened 2021-05-31
Add Auto creation option commands/buttons

manongjohn:
This PR adds the following auto-creation toggle commands/buttons for their Preference counterpart as there have been a  number of requests for easy access to these Preference settings.

image

1. Toggle Auto-Creation (left)
2. Toggle Creation In Hold Cells (center)
3. Toggle Auto-Stretch (right)

You can assign shortcuts to these which are currently under `Misc`

I've added the buttons to the default Quick toolbar as well as Cell - Right-click Menu options.


artisteacher:
It's fantastic getting quick access to these commands!

I feel a little mixed about the placement. On my screen, the buttons are really small and make the timeline/xsheet feel a bit busier. On the other hand - it's very easy access.

If they are not kept directly in the timeline/xsheet, I would have them enabled in the command/quick toolbars by default.

On a semi-related note, I'd love to have the xsheet/timeline toggle brought back. Though for continuity, I'd probably use the same button the schematic & function editor use to toggle between states.

pikhosh:
I think the icons are good; Because it's really hard to design a perfect icon; But their size is really small! And I think it's better to have them in the command/quick toolbar than in that area of the timeline/xsheet.

manongjohn:
I was concerned about the size and placement of the new buttons as well.  I also felt it did make the xsheet/timeline busier, especially since those options are typically on and are colored blue.  They were a little distracting to me at first.

I think I will remove them from there and add them as defaults just to the Quicktoolbar, not the Command toolbar.  I'll also add them to the cell Right-click menu options for those that don't normally use the Quick toolbar but don't want to set up shortcuts for them.

PR #709, opened 2021-06-02
Paint Brush Tool enhancements

### LINKED ISSUE #708: Add Min & Max value sliders for the smart raster paint tool
To help paint areas on smart raster levels I sometimes use the smart raster paint tool, but it's hard to find the right size with only have 1 size slider. So I often revert to using the standard brush with a range set, so I can paint into the corners of my shapes with a lighter touch and fill larger areas by pressing more firmly.
It'd be a huge help to have 2 sliders for the smart raster paint tool to be able to choose a min and max size, so I could use this brush without having to change the size regularly.

manongjohn:
This PR enhances the Paint Brush Tool as follows:

- Adds Pressure Sensitivity and min/max brush size (fixes #708)
image
  - When using a mouse or if Pressure Sensitivity is off, the max brush size is used.

- Replicates the behavior of the old `Finger Tool` by holding `CTRL`, click on style in the drawing and dragging
  - This will pick and use the style you clicked on regardless if the style was used for AREA or LINE.
  - If you click in an empty area, the current style is used.
  - If you release `CTRL` while still dragging, the stroke will be ended and a new normal stroke using the currently style will be started

- Switch style by using `SHIFT` + click a style on the drawing

- Updates Status bar information with the `CTRL` and `SHIFT` options
image

### DIFF
--- a/toonz/sources/tnztools/paintbrushtool.cpp
+++ b/toonz/sources/tnztools/paintbrushtool.cpp
@@ -46,7 +46,10 @@ using namespace ToolUtils;
 
 TEnv::StringVar PaintBrushColorType("InknpaintPaintBrushColorType", "Areas");
 TEnv::IntVar PaintBrushSelective("InknpaintPaintBrushSelective", 0);
-TEnv::DoubleVar PaintBrushSize("InknpaintPaintBrushSize", 10);
+TEnv::DoubleVar PaintBrushSizeMax("InknpaintPaintBrushSizeMax", 10);
+TEnv::DoubleVar PaintBrushSizeMin("InknpaintPaintBrushSizeMin", 10);
+TEnv::IntVar PaintBrushPressureSensitivity("InknpaintBrushPressureSensitivity",
+                                           1);
 
 //-----------------------------------------------------------------------------
 
@@ -237,6 +240,16 @@ void drawEmptyCircle(int thick, const TPointD &mousePos, bool isPencil,
   }
 }
 
+//-------------------------------------------------------------------------------------------------------
+
+double computeThickness(double pressure, const TDoublePairProperty &property) {
+  double t                    = pressure * pressure * pressure;
+  double thick0               = property.getValue().first;
+  double thick1               = property.getValue().second;
+  if (thick1 < 0.0001) thick0 = thick1 = 0.0;
+  return (thick0 + (thick1 - thick0) * t) * 0.5;
+}
+
 }  // namespace
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


DarrenTAnims:
Mr @manongjohn. You're a genius sir. This will help immensely with painting.

artisteacher:
I haven't been able to test this yet - it'll have to wait for after the weekend when I have access to a tablet. However, I've noticed there's a typo in the screenshot. Leves should be levels.

> Paint areas in Smart Raster leves

manongjohn:
Going to merge this and Lock Alpha change as I have an additional enhancement I'm looking to make. Feel free to use the nightly, when done building, to test.

artisteacher:
Pressure sensitivity/brush sizes works great. I love the idea of using shift to pick the style. Though I wish it didn't make a mark at the same time.

Capitalization in the status bar is a bit inconsistent between tools - compare the brush & smart brush.

The finger tool capability works for what it is worth. I wish it was a little more helpful.

Not strictly related but I found an interesting behavior where the style switches to style 1 automatically:
1. Create a new scene
2. Add custom styles to the level palette
3. Select one of the new styles
4. Draw a stroke to create a level - style automatically switches to style 1 for that stroke

manongjohn:
> Though I wish it didn't make a mark at the same time.

Good idea. I will submit a fix for that.

> The finger tool capability works for what it is worth. I wish it was a little more helpful.

What would you suggest to make it more helpful?

As for the style switching, that's probably normal for any creation of a new level when using auto-creation since it's always a safe bet that a new level with a new palette will always have Style 1.

artisteacher:
I guess I just don't get why some pixels are ignored and others aren't. I can't seem to use it to fill in tiny gaps except for rare occasions. It seems to work best to fill in tiny translucent areas.

> As for the style switching, that's probably normal for any creation of a new level when using auto-creation since it's always a safe bet that a new level with a new palette will always have Style 1.

It's interesting though, that the new level has the altered palette.

manongjohn:
I had similar issues with the original Finger tool. This uses the same code. Not 100% sure what the criteria is for which pixels it can fill. Something i can look at in the future and see what can be done to enhance its use.

The switching to Style 1 on a brand new palette is more a legacy palette switching thing since originally new palettes only ever had the 2 styles. Now that you can define a default palette maybe we can change that a bit. When i have some time i can look into staying on the current style when moving from the default palette (on empty cell) to a newly created one.

artisteacher:
You are the best!

artisteacher:
@manongjohn I did get a crash when experimenting with this on a smart raster level. I was using an iPad as a tablet, some of the mypaint styles & the smart raster paint brush.
[crash.txt](https://github.com/tahoma2d/tahoma2d/files/6631936/crash.txt)

manongjohn:
Please open a new issue so it can be tracked separately.

Based on the crash log, it appears you might have been Ctrl-Dragging (Finger tool mode) and then released the Ctrl key while still dragging which lead to the crash?

If that is what you were doing, can you try again it to see if is repeatable?  I know when I tested it on Windows, I didn't have that problem.

artisteacher:
Yep, that was exactly it. It's repeatable without a tablet too. I'll make an issue now.

PR #712, opened 2021-06-11
Fix paintbrush modifier actions

### LINKED ISSUE #711: Smart paint brush crashes in finger tool mode
**Describe the bug**
Ctrl-Dragging with the smart raster brush (Finger tool mode) and then releasing the Ctrl key while still dragging leads to a crash.

**To Reproduce**
1. Create a smart raster level and draw a line with the brush
2. Switch to the smart paint brush and use ctrl+drag (cmd+drag on Mac) to enable finger tool capability
3. Release ctrl (cmd on Mac) while still dragging

[finger_crash.txt](https://github.com/tahoma2d/tahoma2d/files/6633350/finger_crash.txt)

**Device Information:**
 - OS: [macOS 10.15.6]
 - Tahoma2D Version: [e.g. Nightly (June 9th 2021)]
 - GPU: [Radeon Pro Vega 20]

manongjohn:
This PR does the following:

1. Fixes #711 where crash occurred when using `Finger Tool` mode and lifting CTRL while still dragging
  - I think this was caused by passing an incorrect value to the `finishBrush` procedure. I was passing line thickness instead of pressure and if the value was large enough, it could cause a crash.
 
2. Fixes issue reported in #709 where using the style selector mode would leave marks

### DIFF
--- a/toonz/sources/tnztools/paintbrushtool.cpp
+++ b/toonz/sources/tnztools/paintbrushtool.cpp
@@ -558,7 +558,8 @@ void PaintBrushTool::leftButtonDrag(const TPointD &pos, const TMouseEvent &e) {
       // If we were using FINGER mode before, but stopped mid drag, end previous
       // stroke and switch
       if (m_task == FINGER && !e.isCtrlPressed()) {
-        finishBrush(thickness);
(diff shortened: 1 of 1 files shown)


artisteacher:
This works great. Thanks!

PR #715, opened 2021-06-14
PR #3856 from OpenToonz: Fix toolbar resizable width

jeremybullock:
Co-authored-by: shun-iwasawa shun-iwasawa@users.noreply.github.com


manongjohn:
This isn't needed.  The bug this fixes appears to be specific to OT which was introduced by another OT bugfix which was purposefully not applied to T2D since we don't have that issue.

jeremybullock:
Good to know.  I was just going through PRs that hadn't been ported yet.

manongjohn:
Thanks.  I have a running list of fix and feature patches from OT that I want to bring over.  I was going to make it something I do monthly, but never got to it this month.  You beat me to most of the fix. :)

Issue #704, opened 2021-05-24
Memo visibility with Dark mode on macOS


artisteacher:
**Describe the bug**
If macOS is using dark mode, the text isn't very visible when typing a new memo.

**To Reproduce**
List steps to reproduce the behavior.
1. Set a Mac to use Dark mode in System Preferences -> General
2. In Tahoma, create a memo & start typing

**Screenshots, short videos or logs**
<img width="189" alt="Screen Shot 2021-05-24 at 10 42 57 AM" src="https://user-images.githubusercontent.com/21159116/119364736-2ef9f880-bc7d-11eb-8b26-67eb9d0749f2.png">

**Device Information:**
 - OS: macOS 10.14.6
 - Tahoma2D Version: 1.2

artisteacher:
If the memo window could take the background and text color from Tahoma's theme, it would fix the problem.

Issue #496, opened 2020-11-13
Black drawings hard to see with level strip active


artisteacher:
**Describe the bug**
With the Dark and Darker themes, if the level strip is active, it is hard to see black drawings.

**To Reproduce**
1. Use the the Dark or Darker themes
2. Make a vector drawing with just black lines
3. Select the drawing in the level strip panel

**Expected behavior**
I'd expect such a standard linework color to be more visible.

**Screenshots**
Dark theme:
<img width="585" alt="Screen Shot 2020-11-13 at 9 05 22 AM" src="https://user-images.githubusercontent.com/21159116/99080538-6e89ae80-258f-11eb-8d50-e073be2d5470.png">
Darker theme:
<img width="494" alt="Screen Shot 2020-11-13 at 8 40 09 AM" src="https://user-images.githubusercontent.com/21159116/99080141-de4b6980-258e-11eb-877b-53b777eb2cdc.png">

**Device Information:**
 - OS: macOS
 - Tahoma2D Version: 1.1

artisteacher:
Personally I wouldn't want to change any theme viewer background colors for the scene itself. However, it would be helpful if the viewer background lightened to something like #544e4e, just when vector levels are active in the level strip.
<img width="468" alt="Screen Shot 2020-11-13 at 9 15 54 AM" src="https://user-images.githubusercontent.com/21159116/99081477-ddb3d280-2590-11eb-8c99-81f98b538884.png">

jeremybullock:
It's interesting you bring this up.  I was thinking about this recently too.  I originally changed the viewer background for both aesthetic reasons and for stop motion work.  But I agree that it can interfere with drawing - either on vector levels or raster levels larger than the camera/canvas.

I'll take a look and see what can be done.

jeremybullock:
What do you think about lightening the viewer background color for the medium and dark theme.  Leaving darker as a good theme for stop motion?

artisteacher:
That's not a bad solution, though I like the dark background in general. Also, some animators might go back and forth between drawn & stop motion techniques. It would be nice to be able to change the viewer background value on the fly and as needed, without needing to get into preferences.

My first thought was that something like the camera preview color buttons could be great - except for viewer background ( whether preview is on or not). That way you could quickly change the background, depending on your workflow/need. So when doing stopmotion, you could leave the theme default or change to black. If you switch to drawing black vector linework - or anything that extends past the camera, you could switch to a light color.
[...]

PR #719, opened 2021-06-14
fix memo popup background color - OpenToonz 3910

jeremybullock:
Co-authored-by: shun-iwasawa shun-iwasawa@users.noreply.github.com


manongjohn:
Should check to see if this will fix #704

artisteacher:
It does!

manongjohn:
> It does!

Great. I thought as much.

Now need to fix a similar issue with the Viewer for working Vectors through the Level Strip.  Hard to see vector strokes in Dark mode since there is no "drawing box" with a lighter background.  I can't remember if there was an issue opened for it in OT or here, but that is a separate issue :)

artisteacher:
It's #496. Yeah, it would be great to have that fixed.

PR #721, opened 2021-06-16
Update builds to use Rhubarb Lip Sync 1.10.3

### LINKED ISSUE #720: Rhubarb no longer working in nightlies after 5-28
**Describe the bug**
Trying to use the apply audio tab using rhubarb fails on macOS on all builds starting on 5-29. Rhubarb works without issue on all older nightly builds.

**To Reproduce**
1. Setup an audio level and mouth (or etc.) level for lip syncing
2. In the apply audio to lip sync column window, click on apply

Screen Shot 2021-06-14 at 6 20 22 PM

**Device Information:**
 - OS: [macOS Big Sur]
 - Tahoma2D Version: [e.g. Nightly (5-29-21+)]

manongjohn:
This PR fixes #720 

Rhubarb apparently didn't support 32bit PCM WAV files.   In Tahoma, if you lip sync loading directly from the WAV file, it would work with 16bit PCM and 32bit Float WAV files, but not 32bit PCM. If lip sync using a WAV file loaded into a column, it would not work, regardless if 16 bit, 32bit or 32bit Float because internally Tahoma was using 32bit PCM format.

I modified and released Tahoma's version of Rhubarb Lip Sync (v1.10.3) to support 32bit PCM WAV.  This PR updates build scripts to use the newest version.

### DIFF
--- a/ci-scripts/linux/tahoma-get3rdpartyapps.sh
+++ b/ci-scripts/linux/tahoma-get3rdpartyapps.sh
@@ -23,6 +23,6 @@ if [ -d rhubarb ]
 then
    rm -rf rhubarb ]
 fi
-wget https://github.com/tahoma2d/rhubarb-lip-sync/releases/download/v1.10.2/rhubarb-lip-sync-tahoma2d-linux.zip
+wget https://github.com/tahoma2d/rhubarb-lip-sync/releases/download/v1.10.3/rhubarb-lip-sync-tahoma2d-linux.zip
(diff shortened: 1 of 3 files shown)


manongjohn:
@artisteacher When you have a moment, can you test out the artifact works now with the 32bit file you were having problems with? Should work from direct file and from any column loaded with a WAV file.

artisteacher:
All good on my end. Thanks for the swift fix!

PR #722, opened 2021-06-16
Fix crashing on Raw Image Capture

jeremybullock:
This updates the build script to include EdsImage.dll


manongjohn:
Going to merge this since Canon support is not part of PR builds.  Will confirm with nightly build.

PR #733, opened 2021-07-29
Add referenced fills

manongjohn:
This PR adds the ability to perform referenced fills using Smart Raster and standard Raster levels allowing users to create fill levels separate from linework levels.

To use this feature, you need to enable the `Refer Visible` option on the Fill Tool's toolbar option: image

When enabled, fills will factor in all visible levels in the frame when determining fill boundaries.

Smart Raster notes:
- This feature is only available for `Normal` type fills using `Areas` or `Lines & Areas` color modes.
- Referenced fills cannot tell where a referenced line came from so it will be treated as an `Area` for the purposes of filling
  - It will still treat normal `Lines` in the current Smart Raster level as before.

Other notes:
If you undo a referenced fill, change the visibility of columns and then redo, you will get a different result.

Implementation notes:
This feature builds and uses a temporary raster image of the current frame as reference, and essentially uses a raster fill logic.  As such, any Line/Area information from a referenced Smart Raster is lost in the reference image. Additionally, in regards to referenced Vector lines, the fill will be done against a rasterized Vector line.


manongjohn:
Quick update: I haven't merged this yet only because I'm trying to make it so you can use referenced fills on Plastic deformed frames, but have hit a roadblock.  Working on trying to get past the block.

manongjohn:
I'm planning on merging this tomorrow, probably, with a Known Issue that referenced fills on Plastic deformed levels doesn't work quite right.  

It fills it but uses the edge of the mesh, visible or not, instead of the actual line. I will continue to work on trying to fix it whenever I can, but feel it's working well enough overall that I didn't want to hold it back any longer.

Any issues with this, please let me know.

artisteacher:
Had a crash with "fill gaps" and "refer visible" checked at the same time. I had a layer for at least each level type with a few strokes which  combined to make closed shapes. Crashed on trying to fill on an empty level.

[fill-crash.txt](https://github.com/tahoma2d/tahoma2d/files/7140064/fill-crash.txt)

artisteacher:
Also getting some artifacting when filling regions next to each other but with the fills on different levels.
First fill:
Screen Shot 2021-09-09 at 6 36 36 PM
2nd fill:
Screen Shot 2021-09-09 at 6 35 24 PM

manongjohn:
> Also getting some artifacting when filling regions next to each other

That can happen now, which i believe is due to Fill depth = 0. See if it still does that when you change Fill Depth to 1. Sometimes 0.5 even helps stop that.

Will look into the crash issue. If you manage to save a test scene where you can reliably reproduce it, please attach.

artisteacher:
Yep, fill depth as 0 was the cause. Maybe 1 could be the default minimum value?

I'll certainly pass along a test scene if I can. The problem is I didn't save at all - was just working on an untitled scene in the Sandbox when the crash happened. I created and saved a scene the next time and of course it didn't crash.

manongjohn:
It has been suggested before, but I'm a little hesitant to set the minimum to 1. Btw, it does accept fractions as input and 0.5 sometimes helped in some cases I was testing.

Field depth helps to determine how much into an anti-aliased line it should go into and extremely thin lines that are heavily anti-aliased would produce these results with Field Depth 0, but in other cases, 0 is fine.

It would treat the Field Depth issue separately as it's not specific to this change.

manongjohn:
I think the Fill crash reported here is probably the same cause in #784 and not likely specific to this PR. 

There may be an underlying issue causing random Fill crashes on macOS, but I have a hard time producing it. Unfortunately the logs provided show they crash in the same procedure but there is not enough information to determine where it crashes within it.

manongjohn:
Merging this....any new issues with it or a result of it, please open a new issue.

Thanks to all that tested it out and provided feedback!

PR #750, opened 2021-08-15
Changed default sound column volume to 100%.

melieconiek:
Change the default volume for new sound columns to 100% since this is usually where users want it.


DarrenTAnims:
What was the default volume before this change? Did Tahoma always make sounds quieter before?

manongjohn:
LGTM.  Thanks for the improvement.

PR #751, opened 2021-08-15
Update scene sounds on save.

melieconiek:
Trigger a refresh of scene sounds when the user saves. This is a work around for sounds not being updated after some actions. Triggering an update after these actions would cause the application to lag too much, especially when there is a lot of audio.

I've been using this for about a year and have found it very helpful.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1685,6 +1685,8 @@ bool IoCmd::saveAll() {
   result = result && resources.save(scene->getScenePath());
   resources.updatePaths();
 
+  TApp::instance()->getCurrentXsheet()->notifyXsheetSoundChanged();
+
   // for update title bar
   app->getCurrentLevel()->notifyLevelTitleChange();
   app->getCurrentPalette()->notifyPaletteTitleChanged();
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1687,8 +1687,6 @@ bool IoCmd::saveAll() {
   result = result && resources.save(scene->getScenePath());
   resources.updatePaths();
 
-  
-
   // for update title bar
   app->getCurrentLevel()->notifyLevelTitleChange();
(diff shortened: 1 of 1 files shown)


manongjohn:
I know this is a workaround and most people will use Save All.  Should it also be added to Save Scene?

Autosave does use Save All or Save Scene, depending on what options you have enabled.

Side discussion: I haven't done much with sound other than to load it and play it. What actions don't cause sound to update?  Is there an Issue for that somewhere?

artisteacher:
There's been an open issue on the OT side for awhile: https://github.com/opentoonz/opentoonz/issues/1739

If you remove an audio track, or if you edit the track in T2D (ie. trimming or reordering parts of the track), the track playback doesn't update automatically. 

Currently, the best workaround when editing the track seems to be turning the audio track visibility on/off. For some reason that updates the playback.

melieconiek:
@manongjohn I totally missed your comment on this. Yes, save scene as well, probably.

I had tried adding the sound update to the places where it was missing, but the performance impact was staggering. Especially on longer scenes with lots of audio. If I had to guess, this is probably why the audio update doesn't trigger for these actions. I tried adding an update before playback as well. It was the same issue, it just takes too long for the audio update to complete. It's best to just trigger it manually when needed. At least until a better solution is put together.

manongjohn:
@melieconiek Sorry for keeping this open for so long. I wanted to take a look at refreshing sound at an individual cell for immediate sound updates in lieu of this but haven't really had a chance.  I just took a quick look and it will be a bit more involved than I expected.

Regardless if I manage to get sound to update per cell change or not, I realize this is still a good change to have to rebuild all sounds at once during a save.

Can you please update this to also work with `Save Scene` and then I'll merge it?

Thanks.

melieconiek:
I moved the sound update to the root save scene command. It triggers with `save scene`, `save scene as`, and `save all`.

melieconiek:
I left an extra carriage return in there, should I do something about that?

manongjohn:
LGTM.  Thanks for the update.

PR #762, opened 2021-08-22
Add View Check indicators to Viewer

manongjohn:
This PR makes it obvious that 1 or more View -> check menu options are enabled by putting indicators (text) in the upper left corner of the viewer when selected:

image

Updates:
Added Viewer right-click menu option to Hide/Show Check Indicators
image

Option controls Preference setting: `Interface` -> `Show Check Indicators` (Default is ON):

Command button/Shortcut is available for this setting as `Toggle Check Indicators`.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -235,8 +235,8 @@ public:
   void setColorCalibrationLutPath(QString monitorName, QString path);
   QString getColorCalibrationLutPath(QString &monitorName) const;
 
-  bool isCheckIndicatorEnabled() const {
-    return getBoolValue(checkIndicatorEnabled);
+  bool isViewerIndicatorEnabled() const {
+    return getBoolValue(viewerIndicatorEnabled);
   }
 
   // Visualization  tab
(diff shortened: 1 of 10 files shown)


beeheemooth:
That is, if these options are DISABLED, then the text will also be displayed?

beeheemooth:
hmm, it looks like no, as far as I understood from the code. Maybe instead of the word ENABLED put a red dot in front of the text so that there is less scribbling in the window?

manongjohn:
This better?
image

beeheemooth:
I think yes. More compact. It's difficult to change the color?

manongjohn:
No, not difficult to change color though i would prefer to keep the on screen colors for text as they are since it reads generally well. 

Still open to suggestion but would need to see how it reads.

beeheemooth:
The background color of the scene can be very diverse. On a white background, the red letters will be clearly visible, but in this case the text is poorly visible (at least for me). It would be great if the color of the caption changes depending on the background color, as it does with the color of the styles text on the palette when you change the color. The style name is easy to read no matter what color it is.

artisteacher:
> It would be great if the color of the caption changes depending on the background color, as it does with the color of the styles text on the palette when you change the color. The style name is easy to read no matter what color it is.

This would be helpful for any of the red text - ie. frozen.

artisteacher:
Yeah, I'd get rid of the asterisk. Moving frozen over there would also make since. Text is on the small side on macOS - though it's better than how tiny frozen displays. Moving frozen there to keep all indicators in the same place makes sense to me.

Screen Shot 2021-08-27 at 11 47 18 PM

PR #764, opened 2021-08-24
Fix selection tool for motion path editing

manongjohn:
This PR reenables the Selection Tool for Motion Path editing which was accidentally disabled, somehow, since 1.0.

This also makes Motion path enabled tools usable despite what cell you are in on the timeline/xsheet because it was reading the level type of the level you were in and wouldn't allow you to use the tool on a Motion Path unless you were on a blank frame or on a Vector Level to begin with

### DIFF
--- a/toonz/sources/tnztools/strokeselection.cpp
+++ b/toonz/sources/tnztools/strokeselection.cpp
@@ -774,8 +774,9 @@ bool StrokeSelection::isEditable() {
   }
 
   if (!filmstrip) {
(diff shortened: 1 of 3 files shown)


artisteacher:
Is there a reason we can't use the tools if the camera layer is active before selecting a path to edit?

manongjohn:
> Is there a reason we can't use the tools if the camera layer is active before selecting a path to edit?

Tools enable based on the level/column you are in so when you are on the Camera column, almost all tools are disabled by default.  I'll need to make an exception for Motion Path editing when you are in that mode.

PR #765, opened 2021-08-24
Fix keyframe icon positions

manongjohn:
This PR fixes the keyframe icons as follows

- Corrects the position of the Ease In/Out handles on the Xsheet
- Vertically centers all the Camera keyframe icons on the Timeline


artisteacher:
Positions are all improved. If it matters, this did make me notice that the keyframes and text in the timeline aren't centered vertically either.
Screen Shot 2021-08-28 at 8 41 35 PM

manongjohn:
Never noticed it before, but now that you've pointed it out, I will need to fix it or it will bother me to no end. LOL

PR #780, opened 2021-09-08
Allow renaming of revamped Pass Through Fx nodes

manongjohn:
This PR adds the ability to rename the revamped Pass Through Fx node that was added as part of #753.

By default, a newly added Pass Through node does not show it's name (usually PassThrough##).  When renamed by double-clicking the node, the name will appear immediately above it. If the name is blank, it will reset the name back to it's original name and will not be displayed.

Also sets the default text color when renaming stage or fx nodes by using the viewer text color specified by theme's stylesheet.


melieconiek:
Everything looks good here. Thank you for looking at this. :)

artisteacher:
Works great on my end. Thank you!!!

PR #781, opened 2021-09-08
Add Flip and Rotate Selection tool option buttons

### LINKED ISSUE #767: Feature request: Flip & rotate options for the selection tool
When using the selection tool, you can click and drag the handles to flip the image horizontally or vertically, but you can't guage it by eye very well. So you can use the H & V textboxes on the options bar. Entering -100% in them (replacing the 100%) flips in 1 of the 2 directions, but this is undiscoverable for new users. It's such a common question to such a common feature. It'd be so useful if there were 2 buttons for flipping the selection, in the same way as you have 2 buttons for flipping the viewer. Perhaps each button could be next to the appropriate textbox? I don't have strong feelings for where they should go.

Also, maybe left/right rotation buttons to rotate either 90 degrees; or maybe a 2nd set for every 15 degrees (as the brush line is setup for)

manongjohn:
This PR resolves #767 by adding the following 4 buttons to the Selection tool's option bar:

image

- `Flip Selection Horizontally`: Multiplies H value by -1
- `Flip Selection Vertically`: Multiplies V value by -1
- `Rotate Selection Left`: Adds 90 to rotation value 
- `Rotate Selection Right`: Subtracts 90 from rotation value


beeheemooth:
I imagined that the code should be much simpler, but apparently everything is not so simple here, haha.

DarrenTAnims:
Amazing! Thanks @manongjohn. This will help a lot of users.

artisteacher:
Works great on macOS.

Issue #784, opened 2021-09-10
Nightly Build crashes on macOS


artisteacher:
**Describe the bug**
We started using T2D with my classes today and ran into several issues.
1. If Tahoma2D was installed on another account, it crashes on open due to missing the canon SDK (but opens fine on the original account)
<img width="878" alt="Screen Shot 2021-09-10 at 1 59 57 PM" src="https://user-images.githubusercontent.com/21159116/132898074-ae701692-c53e-4db1-a2a5-4b70faf82c97.png">

2. A crash that might have been related to clicking on the edge of the toolbar while in full screen
[brush-crash.txt](https://github.com/tahoma2d/tahoma2d/files/7145598/brush-crash.txt)

3. A crash that seemed to relate to filling
[fill-crash-2.txt](https://github.com/tahoma2d/tahoma2d/files/7145606/fill-crash-2.txt)

4. A mystery crash
[crash-2.txt](https://github.com/tahoma2d/tahoma2d/files/7145612/crash-2.txt)

**Device Information:**
 - OS: macOS Catalina
 - Tahoma2D Version: [e.g. Nightly (2021-09-05)]

manongjohn:
The ESDK issue is a permissions issue. Not sure why it is completely closed off to the group/world.  I'll have to fix the build scripts to force the appropriate permissions on that folder.

To fix it manually in the meantime, please perform the following in Terminal

1. cd /Applications/Tahoma2D.app/Contents/Frameworks 
2. chmod -R 755 EDSDK.framework

manongjohn:
Reviewing the crash logs and attempting to duplicate on my macOS VM (Catalina)...

2. It appears when the user was just entering the canvas (just hovering into it I think), it crashed trying to get the current style meaning something was wrong with the palette or getting to the palette.  Not enough information to determine why.  I did experiment with maximizing the window and trying to move and click in between the canvas and bordering panes (tool option bar, style editor, etc) but could not reproduce the crash.

3. As I mentioned in #733, there may be something that is causing a crash in fills but I have a hard time reproducing it.  In this particular case, the user was either drag-filling or perhaps using Rectangular mode and creating the fill box when it crashed. I did try a number of different things (normal drag-fill, rectangular fill, gap closing, etc)....I managed to crash 1x but I can't be certain what sequence of events caused it.  I couldn't repeat it despite redoing similar actions.

[...]

PR #785, opened 2021-09-10
Set directory permissions on macOS/Linux builds

manongjohn:
This PR fixes one of the issues reported in #784 where other users failed to start Tahoma2D due to missing Canon SDK.

For some reason the Contents\EDSDK.frameworks directory only had owner access and completely blocked group/world access.

Modified macOS and Linux build scripts to change permissions to 755 (owner = full, group = read/execute, world read/execute) for anything the build scripts copies to the final directory such as ffmpeg, rhubarb and the Canon framework (macOS only).


artisteacher:
The action build opens without issue on another account. Thanks for the swift fix!

PR #789, opened 2021-09-22
Style set management

manongjohn:
This PR updates the Style Editor providing some style and style set management options.

style editor preview

### Style Sets

- Styles in the `Textures`, `Vector` and `Raster` tabs will now displayed in collapsable named groups (Style Sets)
- Style sets are named and based on the following root folders with addtional sets based on subfolders within them:
  - `tahomastuff/library/textures` (for `Textures` tab)
  - `tahomastuff/library/custom styles` (for `Vector` tab)
  - `tahomastuff/library/vector brushes` (for `Vector` tab)
  - `tahomastuff/library/mypaint brushes` (for `Raster` tab)
- If no styles are found at the root folder, no style set will be displayed for it, but subfolders will be displayed regardless if empty or not.
- Although you can have subfolders inside subfolders (not recommended), all are displayed on the same level in the Style Editor
- MyPaint application's Style sets that are visible to Tahoma2D will have an `(External)` label appeneded to the style set name
- Style sets can be hidden/collapsed as needed using the new Style Set button on the bottom of the Style Editor
- Style set visibility is saved between sessions

### My Favorites

- Right-clicking any style will provide options for adding/removing styles for the `My Favorites` style set
- My favorite styles are stored in `tahomastuff/profiles/users/<username>/favorites/library/` in one of the following root folders
  - `textures` for `Textures` tab
  - `vector styles` for `Vector` tab
  - `raster styles` for `Raster` tab
- `My Favorites` style set is only visible if there is at least 1 style is found
- You can have subfolders in the root folders mentioned above and display with `(Favorites)`
- When a style is added to `My Favorites`, the file(s) associated with will be copied from the original style set folders to the my favorite folders
- `Generated` styles have no files associated with them so a placeholder file is created in the favorites folder with a `.gen` extension

### Creating/Modifying/Removing style sets

- You can create new style sets either in the default library or as a favorite set by right clicking in any of the style tabs.
- Supported file types per set type:
   - `Texture`: BMP and other 32bit Raster image files that Tahoma2D can load normally
   - `Vector Brush`: PLI
   - `Custom Styles`: PLI, TIF, PNG, TGA, TIFF, SGI, RGB
   - `Raster`: MYB (MyPaint brush files)
- You can copy/move styles from 1 set to another. There are restrictions for Vector sets
- When files are moved/removed, the file(s) linked to the style are physically moved/deleted with no abilty to restore it

### Adding style to palette

- Pressing ALT key and clicking on a style will add it as a new style directly onto the current palette page 
- Right-clicking any style will also provide options for adding new styles to the current palette

### Mutliple style selection

- Press-HOLD CTRL key and clicking on 1 or more styles will add/remove the style from the selection
- You can select across multiple style sets
- Right-click anywhere in the style editor will bring up menu options that can be executed against the selected styles

### Known Issues:
- Palette update delay when multiple selected styles are added to the Palette.
- Existing delay and potential hanging when loading the `Vector` tab for the 1st time


artisteacher:
This is so helpful! I love the ability to favorite and make custom sets.

It might be nice to be able to rename a custom style set, in case you change your mind or notice a typo. Interestingly, renaming a custom style folder in the Browser creates a new folder and keeps the old one too. This might just be the default renaming behavior through the browser. I'd argue though that renaming shouldn't create a new version. That's more of a duplicate or save as function. The newly named directory doesn't display in the style editor until after a restart of T2D.

I did get a few crashes when going back and forth from this page in my browser and T2D. One time I tried to expand a style set after switching back to T2D. Way back in the day I used to regularly get a similar crash on OT but it hasn't happened in a really long time.
[crash.txt](https://github.com/tahoma2d/tahoma2d/files/7212930/crash.txt)

artisteacher:
If enabling/disabling multiple sets, you have to keep pressing the style set button which can be a little annoying. It would be better if that stayed open until you click outside of the dropdown.

manongjohn:
- Rename is a good option to have. 
- Keeping style set button open until either you close it or click outside the window is good too.

Great suggestions. I'll look to add those before I merge.  Maybe I should also add an option to rescan for new/removed sets?

As for the crash...that's an odd one.  Seems very similar to what DarrenT was experiencing but I could never duplicate it.  It seems to be crashing while building an icon for display on a titlebar.  I'll have to look into that more.

artisteacher:
Yes, an option to rescan would be great.

Loosely related, I've never been a huge fan of how new textures are added through the style editor. When the average user clicks on the "load texture" icon, they never think "ok, now I'll head over to the settings tab to actually load my texture". It's just not intuitive. A more expected behavior would be to press the "load texture" icon and then have a browser pop-up to choose an image texture.

Also, thanks so much for working on this. Brush organization has been needed for a long time.

manongjohn:
I agree about the `Custom Texture` style.  When I pressed it the 1st time while working on this, I had forgotten what it was and wasn't sure what it was supposed to do until I checked the `Settings` page.  The tooltip on it doesn't give enough information either.

Perhaps, if adding to the Palette (direct click or alt-click only) it should immediately put you onto the Settings tab?

artisteacher:
> Perhaps, if adding to the Palette (direct click or alt-click only) it should immediately put you onto the Settings tab?

That should work well - though I'd move the position of the `Load From File` field to the top in case the panel is too short for it to be immediately visible. That would have the added benefit of giving attention to the various settings for a texture at the same time. Of course then anyone adding a texture for a vector level will be confused by the texture settings being ignored on vector levels 🙃

artisteacher:
Here's the crash report from trying to expand a collection after returning to T2D from a web browser. I've also opened #790 for the first crash.

PR #795, opened 2021-09-27
Add Tahoma2D Stuff and Downloads folder access from Browser

manongjohn:
This PR adds access to `Downloads` folder and specific `tahomastuff` folders (under a `Tahoma2D` group) from the Browser pane:

This is especially useful for macOS users whose stuff folder isn't easily accessible through the system's file browser.

image

- `Library` - maps to `tahomastuff\library` (already existed but moved into Tahoma2D group)
- `Fx Macros` - maps to `tahomastuff\fxs\presets\macroFx`
- `Fx Plugins` - maps to `tahomastuff\plugins`
- `Studio Palettes` - maps to `tahomastuff\studiopalette`


artisteacher:
This will be so helpful. Thank you!

The other directory where quick access would be beneficial is Downloads.

manongjohn:
Added

artisteacher:
Thank you!!!

PR #798, opened 2021-10-04
Typo fixed (line 34)

kant:


manongjohn:
@kant I had left this open in hopes you would combine the various PRs you were creating single typo fixes, and possibly add more but it appears you have abandoned your effort.

I will merge this single typo fix and if you so decide to resume your efforts to help out the community, it would certainly be appreciated.

For future reference please combine as many typo fixes into a PRs as possible as it is both time and resource consuming for just a single character change. There are also storage cost implications involved when it comes to these PR builds.

Thanks

kant:
@manongjohn Roger that!

PR #805, opened 2021-10-05
Build using Qt 5.11.3

manongjohn:
This PR will drop back to building Windows builds using Qt 5.11.3, the last version before Qt added Windows Ink support to their libraries starting in 5.12.0.

I am doing this because 5.15.2 with/without WinTab support doesn't appear to fix the issues in T2D as it does in OT, despite all 5.15.2 changes being ported over.  In addition to that, the 5.15.2 w/ WinTab support didn't fix any of the reported 5.15.2 issues reported in OT on 3 separate machines I have, yet appears to work for others.

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   Windows:
-    runs-on: windows-2019
+    runs-on: windows-2016
     steps:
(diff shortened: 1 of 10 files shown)


manongjohn:
Canceling this as there are other issues with this version that cannot be fixed without going to later versions or custom compiling qt.

Going back to using 5.9.7 for now and will revisit moving to Qt 6.2 or later down the road when the time comes.

PR #807, opened 2021-10-05
Fix macOS brew error

manongjohn:
Homebrew on macOS builds is trying to update `curl` and running into issues updating dependencies.  This is a workaround to fix the issue.


manongjohn:
Issue fixed. Going to merge it now for future builds

PR #822, opened 2021-10-17
Update MyPaint style usage on various tools

manongjohn:
This PR updates various drawing tools to use MyPaint styles as follows

### Geometric tool - Smart Raster/Raster Levels

- When a MyPaint style is used, a guide will be used to create the desired shape. Once drawing is complete, the MyPaint style will be applied to the guide as if drawn manually.
- For closed shapes, you may see small gaps or imperfections in the line where the stroke starts and stops.  This primarily depends on the MyPaint style and how the start/end of that style is shaped.
- Shapes drawn with a mitered edge (vs rounded), are drawn using multiple strokes instead of 1 continuous stroke. As such, you may see imperfections in the line. (See previous comment)
  
### Geometric, Brush, Fill, Tape and Type tools - Vector Levels

- MyPaint styles are treated as a standard styles on Vector levels


manongjohn:
Did you try changing the Join type when using the MyPaint style? It looks as though it was drawn using the `Miter Join` (default) instead of `Rounded Join`.

PR #824, opened 2021-10-18
Enhance vanishing point management

manongjohn:
This PR enhances the management of vanishing points as follows:

- `Ctrl+Alt` modified to...
  - Allow moving existing vanishing points by click-holding an existing vanishing point and dragging to new location.
  - Allow click-hold-drag to create a new vanishing point and immediately move it to the desired location.
  - Remove a vanishing point after you stop pressing the button provided the vanishing points was not moved.
- Fixed `Shift + Alt` to clear all vanishing points

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -331,28 +331,27 @@ void FullColorBrushTool::leftButtonDown(const TPointD &pos,
   m_assistantPoints         = simLevel->getProperties()->getVanishingPoints();
 
   if (e.isAltPressed() && e.isCtrlPressed() && !e.isShiftPressed()) {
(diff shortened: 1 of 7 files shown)


manongjohn:
I'm a little surprised that the vanishing points are at specific levels rather than at a scene level where as the other guides are at a scene level.  You more likely would want multiple levels using the same vanishing points and would currently need to line them up perfectly for that to happen. 

Should we change that? That would lend itself to possibilities of saving as presets

artisteacher:
I can see the need for both - like onion skinning. However if this was scene level based and we could save/load presets, we could change the loaded guides as needed when going back and forth between levels. We wouldn't necessarily need anything specifically level based in that case.

manongjohn:
Canceling this in favor of the new Perspective Grid Tool PR #835

PR #835, opened 2021-11-03
Perspective Grid Tool

manongjohn:
This PR moves and revamps the Perspective Assistants Viewer options into a `Perspective Grid Tool`.

image

### Perspective Grid Tool image
(Thanks for the icon @konero!)

Supported Perspective types: `Vanishing Point` and `Line`
- Click in an empty space to add the selected `Type` using the current settings in the tool's option bar
   - Click+drag to create the perspective and move into position
- Both types have on-screen controls for rotation, spacing and moving the center point
  - `Vanishing Point` has additional controls for moving the center point into the distance
- Can customize `Opacity` and `Color` for each perspective
- `Horizon` modifies the type to have a horizon perspective
- `Parallel` shows parallel lines (`Line` perspectives only)
- `Presets` save current perspective set for later use

Keyboard options:
- `SHIFT` - Move perspective + perspective's controls
- `CTRL` click to select/deselect multiple perspective objects for multi-object updates
  - Can click+drag to create a selection box to select controls whose center point falls in the box

### Viewer update
In order to show/enable any Perspective Grids in the Viewer when you are not in the `Perspective Grid Tool`, you must enable the `Grid and Overlays` option of the Viewer and check the `Perspective Grids` option:
image

The perspective grids that display are whatever is currently loaded into the `Perspective Grid Tool`.  Select a `Preset` in the tool to change.

#### Brush Tool updates
Use `ALT` to snap to the Perspective Grids.
- Snapping to perspectives uses a guide to determine which perspective to use.
- Can snap to either `Vanishing Point` or `Line` depending on which perspective is closest to to the snapping guide's angle.
- If no Perspective Grids are active, ALT will snap to the Vanishing Point at the center of image


melieconiek:
Very nice. Can we add a setting to limit the draw distance to reduce the horizon haze? An adjustable point where it stops drawing the grid and just draws the horizon line.

It would be useful if holding down `ctrl` applied some angle snaps to the horizon rotation handle and locked the horizon position when adjusting the vanishing point handles -- so you could slide the vanishing point along the horizon, or force the other ray handles to adjust in order to maintain the current horizon position.

Awesome work on this!

image

artisteacher:
I agree that snapping to angles when adjusting line/horizon rotation would be super helpful.

The line mode works great - easy and intuitive.

Love the presets! Can they be stored in the library so they can be easily imported/exported?

I don't really use perspective tools in other digital programs so I'm not sure if this is standard or not, but the vanishing point widget feels less intuitive for me.  I'm just not sure what all the different handles are supposed to do.  I think having different shaped handles for different behaviors/controls would be helpful - like the animate tool's all widget does.

Part of this may be that I am seeing some weird behaviors with the vanishing point where the grid lines disappear. Ie. Dragging the right-most yellow handle further to the right or the left-most handle to the left results in something like this, where the lines are no longer visible:
Screen Shot 2021-11-03 at 5 42 55 PM

For some reason I'm having some deleting issues on Big Sur. Deleting vanishing points through the edit menu sometimes gets a crash. Once after deleting, one guide displayed as deselected and wouldn't select again, even though it should have been deleted. 

[delete-grid.txt](https://github.com/tahoma2d/tahoma2d/files/7470914/delete-grid.txt)

With the vanishing points, it's a little too easy to accidentally create a new vanishing point when meaning to adjust the widget for an existing one.

DarrenTAnims:
Having different shapes for the widgets on the ends of the lines would be really helpful to recognise which is which. Or maybe changing the cursor when you're near them? But this is looking great so far.

artisteacher:
As far as the disappearing grid lines go, the order that the handles are moved matters to a certain extent. Moving the leftmost or rightmost handles does cause the gridlines to disappear if they are the first handles moved, or if they are moved only after moving the bottom spacing and rotation handles. The direction or distance moved doesn't seem to matter. If the vanishing point itself, or the upper inner handles are moved first, then the outer handles can be moved as expected with the vanishing point lines remaining visible.

manongjohn:
> Moving the leftmost or rightmost handles does cause the gridlines to disappear if they are the first handles moved

Normally after you create a Vanishing Point and try to move the left/right most handles without moving the "pivots", it moves the center point over the opposite "pivot" point.  It's suspect there is some sort of calculation issue happening on a mac that doesn't happen in windows.

I'll see if I can duplicate it on my VM and confirm what is happening.

DarrenTAnims:
Firstly, let me say that I think having these lines as a tool on the toolbar and using the tool options bar fits so much better with the other tools and makes the options easier to see and edit. And the whole thing looks great; very professional . But there are a few things that I think could improve this and you might already have planned to look at them or not. But here goes...

Line Option:
1. When you first add a line, you get a green & yellow controller. From the outset, a user can't tell which is which.
  a) Could the yellow controller have a circular arrow around it, to signify rotation and the green one have a plus arrow in the middle, to signify movement.
  b) or could text appear near them as it does for the animate tool in All mode?
  c) or could the cursor change as you hover, like you get for the resizing of panels; with the cursor showing rotate/move style cursors
2. As I start to rotate, the yellow circle falls off the line. I think it should stick to the line
3. When I changed to the animate tool in All mode, clicking and dragging the widgets didn't work. In fact, trying again, it seemed to do that when I clicked into this window to type some text. Create a line. click to another app, then back into Tahoma and I couldn't select the line.
  a) I tried to repeat this, but couldn't a 2nd time, so might not be every time and might be a specific set of circumstances.
4. Then when I changed back to the grid tool, I could no longer select the lines I'd previously left on screen. So I couldn't edit or delete them.
5. After this went wrong, when I clicked and it added a new line, it added it "mirrored" to my location and when I moved left it moved right, when I move right it went left, when I moved up it went down, when I moved down it went up.
5. Could we add the Alt key to move slowly as we do for the animate tool?
6. When I add a line, then tick the Parallel checkbox another yellow widget appears. Again, could this be a different shape/show text/different cursor?
7. The parallel line option moves in the opposite way to how I'd expect. As it starts on the line, I'd expect to stick to the 2nd line. So I could move it away from the green marker to make the lines wider and closer to the green marker to make them closer. Again, it'd be useful to use the Alt key for more precise movements.
8. Just a further observation. Often while playing with the parallel lines option

Vanishing Point Option
1. When you choose the Vanishing Point option, the Horizon and Parallel checkbox options should either disappear or grey out (I'd prefer them to disappear personally)
2. The same point as for the line tool, I can't tell which yellow circle does what. We can't change them all, but I'm sure we could change some to make it clearer what they do.
3. To move the lines further apart/closer together you have to pull one of the markers away from/toward the centre. I would have thought that it'd work like the line's rotate option, that you rotate the marker around to move the 2nd line further away from/towards the 1st. And the rest would follow suit.
4. I'd expect the rotate marker should sit on a line, rather than just to the side. I think it'd help make more sense of this marker.
5. As I move either of the end 2 markers, I wouldn't expect the green marker to move anywhere but up and down on the same vertical line. But for me, it's flying off all over the place.

Others:
1. It'd be useful to have a Clear All button on the options bar

If any of this isn't clear, I'd be happy to explain more, to demonstrate or make a video of it.
And please don't take the length of this message as anything but suggestions to try to make this great tool even better.

Darren

artisteacher:
Crash is fixed and the lines stay visible!

> As I move either of the end 2 markers, I wouldn't expect the green marker to move anywhere but up and down on the same vertical line. But for me, it's flying off all over the place.

I'm also finding this a little hard to control/predict

manongjohn:
As far as vanishing point controls go, when you move the Left control handle, it will move the centerpoint along the line formed by the Right Handle and the Right Pivot point and vice versa.  So if you want it to move along a certain path, line up one handle+pivot along the line you want to follow, then use the opposite handle to move it up/down that line.

It took me a little bit to understand these controls as well. They are actually useful for moving the centerpoint when it's far off screen

These controls are the same in Krita's vanishing point assistant if you want to do any sort of comparison.

artisteacher:
When horizon is turned on, if the spacing is turned up too high, the straight horizon is lost:
Screen Shot 2021-11-05 at 9 54 23 AM
The other handle positions don't seem to matter. Personally, I'd expect the horizon to stay flat, no matter the other settings.

artisteacher:
Thanks for the explanation about the outer handles. I'm just seeing a really jerky, glitchy looking behavior when I move them - especially if I haven't adjusted the other handles:

https://user-images.githubusercontent.com/21159116/140538553-0a02d96e-0da4-43b3-aaf7-cd807566b45f.mov

You can also see an instance of an extra vanishing point randomly being created when I first tried to move the handle.

jeremybullock:
Are the the grid and orthographic grid still available?

artisteacher:
Horizontal or vertical grids can be added with the line mode with parallel turned on. There's not a direct orthographic option, though you can add two line widgets to end up with an orthographic grid.

manongjohn:
If you try moving the left/right vanishing point handles when they are all on the same line, as in when you first create a vanishing point, technically you shouldn't even be able to move the handles because the line of movement is exactly the same from either side.

Right now what happens is it pulls the center point over the moving handle's pivot point and may or may not be over it exactly to avoid divide by 0 errors.  I'll need to tinker with it more as when this happens, stuff can get buggy really fast.

artisteacher:
The additional icons and change in spacing direction are both really helpful!

I'm still finding the movement with the left/right handles hard to control.

With some spacings with horizon turned on, I'm seeing a gap between the guidelines and the vanishing point
Screen Shot 2021-11-07 at 8 41 40 PM
:

manongjohn:
> With some spacings with horizon turned on, I'm seeing a gap between the guidelines and the vanishing point

That gap is intentional, to help reduce the haze caused by so many lines going to the center point.  The center point is still the vanishing point.  I mimic'd this after the Morevna Editon's vanishing point.  It kind of gives it a plane-like effect too.

As for controlling the handles, you really only need to use those if you need to move your vanishing point off canvas at a far off distance.  You can move the center point directly if it's on screen. And if you try moving them when both handles and center point are all in a straight line, it will behave oddly.  Not much I can do about it.

In the interest of simplicity I am adding the toolbar option `Advance Controls` to show/hide (default) those Vanishing Point movement handles.

jeremybullock:
I tried this tonight, and there is some nice functionality.  I do miss the ability to do precise isometric work with this.

manongjohn:
> I tried this tonight, and there is some nice functionality. I do miss the ability to do precise isometric work with this.

Is this because you can't see the actual spacing/rotation angle values?  I debated adding controls for them in the tool options.

jeremybullock:
> Is this because you can't see the actual spacing/rotation angle values? I debated adding controls for them in the tool options.

It was a one click option before.  Now getting it set up perfectly is impossible.  
I also don't see a way to ensure a perfectly vertical or horizontal grid.

manongjohn:
1 way to do that is to add a Line Perspective which automatically makes it at 0% angle so perfect horizontal is easily achieved.  Add a second Line Perspective then hold ALT to rotate (x2) at 45 degree increments so getting to perfect vertical isn't entirely difficult either.  Save and reuse for later or in other scenes.

If people think it would be helpful, I can expose controls for spacing/angle values on the toolbar so you can enter it directly to jump to specific values.  Plus it could be a good way to see what the angle is when rotating/spacing using the on-screen controls.

artisteacher:
Exposing the spacing/angle values would be great. 

I think angle snapping is every 15deg for the brush. Should they be consistent?

manongjohn:
If we want to include default Presets, I am certainly open to that idea.  I would just need someone to create/name them and send me the `perspective_grids.txt` file that is generated in your the profile directory and and I can add logic to initialize on first time use.

artisteacher:
Also, can we use the same shortcut for snapping to angles in both the vector brush and the grid? It's cmd/ctrl for the brush...

I think it would be helpful to add the grid presets to the library instead of profiles, to allow for easier sharing. Either way, I don't mind making a few defaults to share. Just let me know which ones.

manongjohn:
> I think angle snapping is every 15deg for the brush. Should they be consistent?

I think in some other tools, like the geometric tool, snapping was at 45deg.  If 15deg is a better choice, I can easily change it.

manongjohn:
> It's cmd/ctrl for the brush...

We are unfortunately inconsistent when it comes to CTRL modifier key usage.  In one tool angle snapping is done with CTRL, in another it's SHIFT, I believe.

CTRL is currently used for selecting multiple objects to multi-object operations, which I chose to emulate how selective add/remove items is done in Windows file browsers.  I'd need to switch modifier keys around if that's the case.

artisteacher:
> We are unfortunately inconsistent when it comes to CTRL modifier key usage. In one tool angle snapping is done with CTRL, in another it's SHIFT, I believe.

Yeah, looks like polyline is shift for snapping to angles while the vector brush is ctrl. It doesn't look like either one should change, since the other modifiers for the tools are standard behaviors. Oh well. As long as the commands end up in the status bar, people will figure it out.

DarrenTAnims:
The changes made recently, since I last added my comments, are superb. This is so much easier to use; so much easier. And the help text, at the bottom of the screen helps too. But I am confused about the wording of one of them and what it means though.

Holding Alt does allow snapping to 45 degree intervals. I've seen some conversation above about making the 15 degrees. I think I'd prefer 15 degrees too (just to add to that camp). But if you've already started rotating, it doesn't snap at 0 degrees, then 45 degrees, etc. It moves 45 degrees from where it is, which isn't what I'd expect. I'd hope to use the snapping to snap AT 0, 45, 90 degrees, etc. Without it, it's hard to snap back to a horizontal or vertical position.

The 2nd thing is, what does "& space/lock center movement to horizon" mean? I thought that by also holding the space bar, the snapping would snap at 45 degree angles to the horizon, so that I could rotate to 0, 45, 90 degrees, but it doesn't. Is it just me that misunderstands this?

manongjohn:
Heh...no worries...trying to condense how something works into very few words to use as a hint is always difficult.

Currently, the 45deg snapping, is based on current angle rotation, not from 0.  I think that's how some of the other angle snapping works as well?  I'd need to confirm.  If other angle snapping is done to actual 0, 45, 90, etc...then I'll make this match to be consistent.

The "& lock" was meant for the "Snapping " where as "Lock center movement to horizon is another thing.

As for `Lock center movement to horizon`....typically if you move the center point directly, you can move it in any direction.  The ALT modifier limits the movement to the horizon line that it is on.

I think in need to find a better way to word the ALT modifier behavior. Please feel free to suggest.

ALT does 3 things depending on the control you are moving:
1. Locks angle rotation changes to 45deg increments
2. Locks spacing changes to 10 point increments
3. Limits the center point to the horizon line

artisteacher:
I agree with @DarrenTAnims about the snapping behavior.

manongjohn:
Was looking at the other rotation snapping and I see how it snaps to 15deg positions, I'll be changing the rotation snapping for perspective tool to match.

DarrenTAnims:
Finding short descriptive text is really hard. And isn't one of my specialities :-)
But after playing with it for a while, I now understand what they all do. Here's my best attempt so far:

   **Alt - Snap rotation. Snap spacing. Lock line movement to horizon.**

DarrenTAnims:
The snap rotation surprised me though. I expected the rotation to change only as the widget passed each 45 degree point. For instance, if I was rotating a line around 360 degree, I have to move the rotation widget around the centre point through 360 degrees. But if I hold Alt, the line moves through 360 degrees as I move through 90 degrees on screen.

Also, the rotation widget starts on the line, but soon falls off it as soon as you move it. I'd expect it to be locked to the original line. The same with the line's spacing widget. It drops off the line that you're on. I think it'd feel more intuitive if it remained connected to the initial line.

I guess you've made the Vanishing Point's spacing widget work the same as the line widget does, where you move it away from the centre widget to change the spacing. However, I feel like it'd be more intuitive to have it locked to the initial line that it's on and then you rotate it to change the rotational spacing.

Small things on the options bar:
When you've got a line selected, can the Horizon tickbox be greyed out until you tick the Parallel tickbox? You don't see any change until you press it anyway.
When you've got a VP selected, can the Parallel tickbox be greyed out. I don't think it does anything.

Now that's out of the way, can I say, I love the idea of hiding some of the perspective widgets, so you only see them when you tick the Advanced Controls checkbox. I love the new widget markers. It's much easier to see at a glance which is which, especially after looking away and back. I hadn't noticed the status bar tips until this release, so that really helped me. I love the Shift key to move all of the widgets together. I'm using that a lot.

artisteacher:
Maybe the status bar could be a little more context sensitive in this case? If horizon isn't ticked, just: 
Alt - Snap rotation/spacing
If horizon is active: 
Alt - Snap rotation/spacing; Move center along horizon

Alternatively, center could be point or vp...

I think that "⌘Click/Drag - Select multiple Perspective Objects" could be just: "⌘Click - Select multiple"

manongjohn:
`Parallel` and `Advanced Controls`  will become enabled/disabled depending on if you have `Line` or `Vanishing Point` selected in the toolbar.  They will also be enabled if you select a perspective object on screen that the control is relevant to.  For example: If you had `Vanishing Point` selected in the tool bar, normally `Parallel` would be disabled, but if you select an existing Line perspective object on the screen, it will enable the control.

artisteacher:
I've noticed that when zoomed out (below 75% zoom), clicking with the tip of the mouse within this area:
Screen Shot 2021-11-09 at 10 37 27 AM
more often results in unintended vanishing points or lines. The lower the zoom, the worse this gets and the target area (where you can click without making a new guide) seems to shrink. This is the same for all of the handles.

The zoom level doesn't seem to matter for the other extra guide issue. As previously reported, if the first action after creating a vanishing point or line is to click on the vanishing point/center to adjust the location, a 2nd vanishing point or line is made. This is the case even when clicking directly on the center of the vanishing point.

manongjohn:
Currently rotation/spacing movement is based on the difference in movement of the control and therefore is not tied to any line or any particular angle. Depending on how far away you are from the center point it will change how quickly you will fall off any line it appears to be on or how/when it will snap. Controls usually just happen to be on lines when the object is created because it's easy to just initialize controls x or y distance away from center point.

I didn't originally want to tie controls to grid lines as I would need to factor distance from center point, angles from some reference point and other things to calculate rotation/spacing movements, plus keeping things on lines. (Chalk it up to lazy development and/or all this math was giving me a headache to begin with *smirks*) 

With that in mind, if we tie controls to lines, 1 of 2 things can happen if you consider moving the center point:
1. As lines move, controls move with it:  This would allow angle and spacing to be maintained, but can cause controls to go off screen and be come hard to access especially for vanishing points moved far off screen.
2. As lines move, controls stay in position:  Controls stay on screen, but angles/spacing will adjust relative to where the center is moved to and where the controls currently are.  You'd have to readjust angles/spacing whenever you move the center point.

2 is probably what needs to happen, but I'll be honest and say I find it frustrating dealing with setting up perspectives in the Morevna Assistant which behaves this way.  I like knowing that I set my desired angle/spacing and can move/relocate the center as needed without having to worry about fixing things because I moved it. The only saving grace is, I can use the SHIFT modifier to move the center and controls at the same time.

Anyway...will work on changing it.

manongjohn:
> The lower the zoom, the worse this gets and the target area (where you can click without making a new guide) seems to shrink.

I see the same thing. Will look to fix.

> As previously reported, if the first action after creating a vanishing point or line is to click on the vanishing point/center to adjust the location, a 2nd vanishing point or line is made.

Just to verify something, when you create the 1st vanishing point, is your mouse going off canvas at all or moving around before trying to move the center point directly?  Or do you immediately click again to try and drag the center point right away?

artisteacher:
This is when I immediately click.

manongjohn:
FYI.  Currently working on changing how the Spacing control works.  Just working out 1 last issue with ALT spacing.

Still debating adding edit boxes for rotation/spacing values.

DarrenTAnims:
This is looking so nice.
I like the idea of seeing the values change as I rotate, and maybe adding the rotate button on the options bar (as you have for the selection & animate tools) for consistency. But I think it'd be handy for setting the line rotation to a simple angle: 45, 90 or 180 degrees.

artisteacher:
I agree that edit boxes in the tool options bar would be helpful. 

Targeting issues are both fixed on my end!

manongjohn:
So I've modified the Spacing controls as follows:
- `Line`: Spacing control stays on grid line and can move along the grid line without changing spacing
- `Vanishing Point`: Angular rotation of Spacing control relative to Rotation control alters the spacing. Spacing control will **not** stay on line due to how rays are spaced as the angle increases/decreases.

Other notes:
- When you Rotate a perspective, the Spacing control will rotate to maintain spacing.
- When the center point is moved directly, the spacing and rotation controls do not move. This will cause the space/rotation to adjust as you move the center point.  Exception: with Shift or with locked horizon movement, spacing/rotation controls will move to maintain rotation/spacing.
- Known issue: When spacing multiple objects, especially if adjusting line and vanishing point simultaneously, the spacing control of the other objects may get "glitchy" at certain angles as you move the main control.  Regardless of where the control ends up, the position-angle should still be correct for it's spacing.

manongjohn:
I've added spacing/rotation options on the tool bar.

I believe I've addressed most of the concerns and suggestions.  Unless bugs or serious issues are found, I think I'm done with this.  Will leave this PR open for a few more days to give anyone who wants to test it out before it's merged.

artisteacher:
It's great having the additional tool options. If the tool option bar doesn't have enough room, the values for spacing and rotation end up just displaying the end of the numbers - ie. .12 rather than 100.12. I'd probably increase the minimum box size a little.

On Mac laptops, our delete key is the equivalent of a backspace? on non-apple keyboards. So delete doesn't work without using the fn key. This is the only program I use where I need the fn key to delete, so I tend to use the cut command instead. Unless there's a fix to make the Mac laptop delete key work, it would be helpful to add either backspace or cut as other ways to delete the guides.

Last thing, the .grid files are not visible in the Browser Room. It would be nice for visibility to be on in the Browser, though the grid files should obviously be blocked from loading into the Timeline - just like the macros.

I'm looking forward to sharing this with my students! I really prefer this implementation over the proposed OT version. Thanks so much for your hard work!

manongjohn:
@artisteacher I'll expand the boxes.  I originally had them sized the same as similar fields on other toolbars but shortened them because it seemed like a lot of wasted space.  I forget about the font sizes on other systems.  Might be why some of these fields are so wide. :)

I didn't know that about the Mac Delete key.  Will try and keep that in mind in the future.  I'll see about switching the use of the actual DEL key with whatever T2D's shortcut is for `Delete`. I can also look to support for the `Cut` shortcut at the same time.

I'll also look to make the .grid files visible in the Browser as I know it's difficult for mac Users to get to the stuff folder outside of T2D.

artisteacher:
> shortened them because it seemed like a lot of wasted space.

There's lot's of wasted space on my end too. I assumed it was because you can have ridiculously big values for some of the fields. Even if you double the field size, that will still be a lot smaller than the fields for the Animate Tool.

Thanks for looking at delete and the .grid visibility!

manongjohn:
For certain fields like in the Animate tool, we just set the max size which is calculated to be based on the font size of a template string + 10 (pixels?).  That's mostly why there is so much extra space.

I just realized, I set the ability to delete perspective objects to use the Delete Shortcut already.  The default is the DEL key. Does that not work because it shouldn't work for any other tool if that is the case.

DarrenTAnims:
I've taken a look at this today. Every time I download a new version, I love it even more :-) Great job. The rotation, moving & spacing are all easier to do and more intuitive too. I love the recent changes, but as always, I've a few comments for you to decide if it's just Darren or if you want to change anything ;-) 

Having the rotation in the options bar is really useful. And I can see what it's doing. It adds 90 degrees to the value each time you click it. However, I was hoping that it would be a shortcut to the rotation snapping, where it would move at 15 degree angles, to the next 15 degree spot. So if the line was angled at 4degrees, then clicking the left arrow would take it back to 0 or clicking the right arrow would take it to 15degrees. Otherwise, after moving the angle by dragging it, the only way to hit 0, 45 or 90 degrees is to type in the box, which seems to defeat the point of the button.

And whatever figure you settle on for this rotation step, it should be added to the tooltip. "Rotation selection right 15 degrees" or whatever. Currently is says, "Rotate Perspective Right". It's not necessarily a perspective line.

Also, if you click and drag your line, it rotates from 0 to -180, then it jumps to 180 and then moves back to 0. But if you click the rotate buttons, it keeps adding or deducting 90, leading to very large figures. I would hope it would stay within the same range as adjusting it by hand.

Clicking and dragging on the labels doesn't adjust the values as it does for the selection, animate and other tools. I use that feature a lot, so it'd be nice to have the grid label do the same, for usefulness as well as consistency.

As I said before, I'm not great with tooltip or status bar text, but I think it might be useful to change:
"Alt - Snap rotation/space; Move along horizon"
to:
"Alt - Snap rotation/spacing; Move along horizon (when Horizon ticked)"
I thought this meant to use the space bar, rather than it changing spacing.

It might even be worth splitting off the "move along horizon" part to not work with Alt along, but instead work with Shift+Alt together, being as it has the same effect as using shift (moving all controls together). Then you could have it as a separate tip: "Shift+Alt - Move along horizon (when Horizon ticked)", which again, simplifies the tips.

The final thing, that you've already covered in an earlier chat, is the gap between a VP with horizon ticked and the actual horizon line. As you move the lines wider, they are drawn nearer to the horizon, till they actually touch it. But having that gap seems a little strange. Is there anything that can be done to draw the lines up to the horizon line?

manongjohn:
Hehe. It's all good.

The rotation buttons rotates 90deg from current position.  That's how it was implemented in the other tools so I kept it consistent.  I was thinking 90 might be too much and will likely change them to 15deg increments in all tools in a separate PR, so I will leave it as is for now. Making it snap to nearest 15deg increment and tooltip wording will be considered/address then as well.

I'll see what I can do about the angle being displayed.

I'll see about adding the label dragging.

Someone's starting to know my trigger word..."consistency". Careful not to abuse it :)

As for statusbar descriptions, I'm trying to use less words, not more. :)  Wanted to avoid having multi-key modifier combinations hence overloading the ALT key.  I think what I have suffices to explain that something different happens depending on the control you are on. However, Shift+Alt is probably a better choice for locked horizon movement since it would work for any control, not just center point

I can draw to the center point and keep the haze or stop at an artificial horizon line and reduce the haze to a certain degree. I don't particularly want to make that a setting in any fashion.  I may be able to reduce the opacity as it gets closer to the horizon so the overlap of line isn't so pronounced and reduce the haze that way.  Have to play with that.

I personally prefer it drawn to center point because I can then overlap the center points of VP and a Line to form a soft "group".  This way, I can rotate them together on the same center and won't need to reposition them to line them up again.

artisteacher:
> Does that not work because it shouldn't work for any other tool if that is the case

Correct - my delete key doesn't work anywhere in the program. Which is why I usually use the cut shortcut.

manongjohn:
Have you tried changing the Delete shortcut to the `Backspace`?  If that works, does it make sense to default the Delete shortcut to Backspace for Mac builds?

I think there is only 1-2 places in T2D where Delete vs Backspace is actually coded to do different things.

artisteacher:
Backspace is the standard delete key on macOS. All Mac laptops, and the default iMac keyboards just have the backspace delete key. You can upgrade the iMacs to get an extended keyboard. In that case both backspace and forward delete are labeled as delete - though forward delete has an additional icon. It's not unusual for my students to use the wrong key first when using T2D.

While we can change the shortcut, I think the standard should really work. I do have to confess that since I update/test out action builds too often, I've given up changing incompatible shortcuts (zoom, delete, insert). 

That said, there probably is a subset of users that use non-apple keyboards that would probably end up confused if their delete key stopped deleting. Like how the majority of us are confused now 😅

DarrenTAnims:
One thing I've just noticed - When I click the camera view button or the preview button on the viewer, you can still see the guides that go outside of the camera box. I'd expect them to be constrained to the camera box. Is this possible? It might speed up the view by not drawing as much. Not that it's slow atm.

artisteacher:
They don't render so should we even see the guides when preview is on? I'm not sure about the sub-camera preview though, since I don't tend to use it. 

We can actually add/edit guides in the camera view - so it is helpful to be able to see the guides outside of the camera. Though for the brush or etc. maybe it would make sense for guide visibility outside of the camera to be tied to the background opacity value?

On a similar note, the guides can look interesting in the 3D view.

artisteacher:
New updates all work as expected. 

It might be nice for grid visibility to save with the scene too. If I left grid visibility on last time I accessed the scene, I'll most likely still need it on in a new session.  Along similar lines, rule of thirds turns on by default when grid visibility is turned on. I get that it's a visual cue that grid visibility is on - but I don't tend to need it myself. If visibility isn't going to be sticky, I'd rather have rule of thirds off until I choose to turn it on.

manongjohn:
The `Rule of Thirds` and `Perspective Grids` are checked by default on initial setup, so it will show something when the `Grid and Overlays` button on the viewer is enabled.  I don't want to turn off Rule of Thirds so when it is used for the 1st time and no grids have been created, something will show.

The button and checked settings are saved when you quit normally out of T2D and restored on startup so a grid, the last <custom> that was saved when quitting, should show immediately even before you select a project, unless I'm misunderstanding something.

artisteacher:
The presets and <custom> all work perfectly. 

> I don't want to turn off Rule of Thirds so when it is used for the 1st time and no grids have been created, something will show.

That's what I assumed. It's just that grid visibility is always off when I restart T2D - no matter how I left it last time I was working. So every time I start T2D and turn grid visibility back on, I also have to take the extra step to turn off rule of thirds. 

I don't mind it being there the very first time. I'd just rather it stay off once I've turned it off, until I choose to turn it back on. It's just a small annoyance.

manongjohn:
Odd. It works fine on Windows for me.

Is the Grid and Overlay Button staying on and just the Perpective Grid doesn't stay checked?

artisteacher:
No, the Grid and Overlay button is always off upon restarting. The perspective grid does stay checked.

manongjohn:
I'll have to see if I can duplicate in my mac VM as that button stays on upon restarting for me in Windows.

Last question as it could be related...if you uncheck Rule of Thirds and restart, does it stay unchecked or is it rechecked again?

artisteacher:
It is rechecked again.

manongjohn:
Sounds like it's failing to write environment settings to the `tahomastuf\profiles\users\xxx\env.ini` file, though that would also mean it may be failing to do that for other things as well.  I'll have to investigate that separately.  Might be a mac specific issue.

artisteacher:
Just checked and the button state/grid visibility does save correctly in the latest nightly build.

manongjohn:
I can reproduce the issue with the button state/grid not sticking on macs but it works fine when I try to debug it through xcode, go figure.

It is related to this PR, somehow, even though I didn't directly modify anything that would cause it not to save.

Technically it is saving the statuses to the file when it is saved, but appears to be resetting on startup, and it's impacting other tools as well so I'll need to identify and fix it before I can merge this.

artisteacher:
The grid visibility settings are loading properly now. Thanks!

manongjohn:
@artisteacher Thanks for verifying.  

I'm going to merge this now.

I am actively working on creating grids at the Level...um..level...and will create a new PR for it once it is ready.

PR #413, opened 2020-10-25
Add Shader Fx: HSL Blend GPU

justburner:
This PR adds a shader Fx to perform component blending modes such as: Hue, Saturation, Color and Luminosity.
Those exists in programs like Photoshop, Krita, CSP... but does not natively exists in OpenToonz or Tahoma2D.
It will be found under "Shaders" as `GPU HSL Blend`.

Settings to perform each blending mode:

Hue Blending:
☑️ Hue
⬜ Saturation
⬜ Luminosity

Saturation Blending:
⬜ Hue
☑️ Saturation
⬜ Luminosity

Color Blending:
☑️ Hue
☑️ Saturation
⬜ Luminosity

Luminosity / Brightness Blending:
⬜ Hue
⬜ Saturation
☑️ Luminosity

PR #841, opened 2021-11-29
Update Shader Fx: HSL Blend GPU

justburner:
This PR updates `HSL Blend GPU` shader (#413) to match the one merged recently in [OpenToonz #4146](https://github.com/opentoonz/opentoonz/pull/4146) by adding Clipping Mask feature.


manongjohn:
LGTM.  Thanks for porting the update.

PR #843, opened 2021-12-03
Update Actions build scripts

manongjohn:
This PR does the following:

- The build scripts zip up the builds and posts them as artifacts.  When downloading a github artifact, github automatically zips it again (on Windows at least), resulting in a double-zipped download.

I've modified the Windows workflow to uploaded the unzipped Tahoma2D folder as the artifact rather than the zipped build.  The zipped build will still be used when for updating the nightly releases.

- Currently when this repo is forked, the action scripts are copied to the fork and executed when the owner pushes to their master.  The fork's actions fail in SDK steps because it does not have access to the private SDK repo.

This adds checks around steps for SDK and Nightlies updates so those steps only work when the repository_owner = tahoma2d


manongjohn:
All checks passed and look good to me.  Windows artifacts are no longer double zipped.

Going to merge this now since it only impacts the build scripts so I can verify the nightly builds are still working correctly.

PR #847, opened 2021-12-19
Implicit Frame Hold

manongjohn:
Normally for an image to display in consecutive frames, you would have to  use an Explicit Hold, exposing the image in the timeline over multuple frames for as long as you want the image to be rendered.

This PR adds Implicit Holds, where you expose the image once and the image will be rendered on all subsequent frames until either a new frame is encountered, a stop hold frame is encountered or you've reached the end of the scene.

image

Notes regarding this change:

- Implicit Holds work only for all level types except sound and note levels
- Implicitly held frames show as a tinted cell
- Implicit Holds works for normal view, preview mode and output rendering
- New Preference: `Enable Implicit Hold` (Enabled by default)
image
- Preference `Enable Auto-stretch Frame` is disabled by default
- New button on Quick Toolbar
image
- An implicit hold of a frame can be stopped by using Cell menu/context menu option `Stop Frame Hold`  to add a stop frame. on the timeline.
  - Stop frames are denoted with an `X` through the cell.
  - Stop frames have drag bars to move them
- Tools and most Cell menu/context menu commands work on implicit cells as if they were exposed
- The `Insert Frame` command now inserts frames before the current Scene frame instead of after


melieconiek:
Awesome stuff! This will making working on the xsheet a lot easier.

I would prefer that drawing in an implicit exposure would behave like drawing in an explicit exposure. I.E. it would target the currently exposed drawing and wouldn't trigger autocreation if it was disabled. Right now it behaves as if there's no drawing exposed there.

manongjohn:
> I would prefer that drawing in an implicit exposure would behave like drawing in an explicit exposure. I.E. it would target the currently exposed drawing and wouldn't trigger autocreation if it was disabled. Right now it behaves as if there's no drawing exposed there.

This did kind of cross my mind as I was working on this. I suppose there are really 3 options that can be considered when drawing on an implicitly held frame:
1. Update the original explicitly exposed frame (CSP does this)
2. Create a new duplicate frame of the original frame (Krita does this)
3. Create a new blank frame

Not sure what Adobe/Flash or other animation software might be doing.

Thoughts?

melieconiek:
If you changed to work this way it would work the same as if you were using explicit holds.

If you wanted a new drawing to be created in an implicit hold you would keep `Enable Creation in Hold Cells` enabled. This way the user would still have the option.

manongjohn:
True. Option 1 would be when Autocreation is disabled.  Option 2 or 3 would be when Autocreation is enabled.

Let me see what I can do about Option 1.

manongjohn:
Sorry for the force push. I wanted to rebase this PR against the latest master due to the recently applied OT patches that impacted the Timeline/Xsheet.

With this last set of commits, I've updated the tools and almost all Cell menu/context menu commands to work with implicitly held cells as if they were explicitly held.  Let me know if you find an action/undo/redo that doesn't quite work the same way with explicitly held cells.

Please note that I've changed the behavior of `Insert Frame` to insert Scene frames before the current indicator instead of after.  Instead of extending the cells under the current frame indicator, the cells are now shifted, and the prior cells will be extended to meet the shifted frames.

As usual, any questions/issues/concerns, let me know.

melieconiek:
Everything looks awesome, can't wait!

melieconiek:
Oh, one thing, zerary effects auto fill when created, which isn't necessary for implicit holds.
image

melieconiek:
A stop marker at the end of a scene shouldn't be included in the auto end frame for preview/render.
image

melieconiek:
If you enabled explicit holds it renders the last drawing in the stop frame.
image

manongjohn:
> If you enabled explicit holds it renders the last drawing in the stop frame. 

I can't seem to reproduce this one.  I have fixes for the other 2.

melieconiek:
Replicated in the most recent build.

1. Disable `Enable Implicit Hold`.
2. Create a new level on frame 1, draw something in it.
3. Move the viewer to frame 2.

The drawing shouldn't be rendered in this frame, but it is.
image

manongjohn:
Thanks. I was able to duplicate.  Will look into.

manongjohn:
Should be fixed now.  Let me know if there is still an issue with it or if you (or anyone else for that matter) find anything else that needs to be fixed.

melieconiek:
That's all I've got. Thank you for the fixes!

melieconiek:
The auto size doesn't update properly when a stop frame is first added. It will set correctly after you drag the frame around or makes some other kind of edit.
image

manongjohn:
Unable to duplicate it despite how simple it sounds like to duplicate it.

manongjohn:
Ignore my previous/deleted reply. It took me a minute to understand what was wrong with the auto stop marker. Should be fixed now.

Unless a significant issue/concern is brought up, I will likely be merging this in the next day or so.

beeheemooth:
How to select all frames in this case? Clicking on the column heading selects all frames, but they cannot be moved all together.

ezgif-7-25d05afef4

Suppose a column has a thousand frames. I only want to keep the first three. How do I delete the rest?

ezgif-7-650cb56186

melieconiek:
@beeheemooth If you click and drag a region inside the xsheet you can drag the exposures about. You can click, hold shift, and click again to create a region as well. This can be used to select a range for deletion as well and isn't any different than what you'd need to do for explicit holds.

Dragging exposures after selecting the column doesn't work, implicit holds or not, and should probably change. You can certainly delete the entire column that way, though.

selection

beeheemooth:
@melieconiek I understand that frames can be selected and moved, but this is not the essence of the problem I described. Look at the video: a thousand frames linked by exposure can be removed in two steps (1.delete one frame 2.delete the rest of frames). How can you do this in this case as easy? How to select all frames (to the top or, to the bottom) without stretching the kilometer selection frames? 
ezgif-2-a1244384bd

melieconiek:
@beeheemooth 
> How to select all frames (to the top or, to the bottom) without stretching the kilometer selection frames?

I understand you're used to doing it a particular way, but making a region selection with shift click isn't any slower.

selection

manongjohn:
To @beeheemooth 's point, with implicit cells, what is lost is the ability to click on drag bar and it auto selects all frames that are connected.

I can look at using the implicit cells as connectors and selecting all those that are connected up to a stop frame or last populated frame. May take me a bit

beeheemooth:
Little by little, implicit cells become more and more explicit ;)

melieconiek:
I really don't like the drag bar change, the extra step of having to select the exposure is a little irritating. Maybe if it only did this if you clicked on the drag bar region outside an exposure?

Otherwise, can this be optional?

manongjohn:
> I really don't like the drag bar change, the extra step of having to select the exposure is a little irritating. Maybe if it only did this if you clicked on the drag bar region outside an exposure?
> 
> Otherwise, can this be optional?
> 
> EDIT: I don't think this is the behavior that a new user would expect.

I am not fond of it either and also agree about new users.

Perhaps i can assign it a Modifier key to do this type of selection.

melieconiek:
> I am not fond of it either and also agree about new users.
> 
> Perhaps i can assign it a Modifier key to do this type of selection.

That sounds like a good compromise. It's not that I would _never_ want to do this, I just don't think it should be the default behavior.

manongjohn:
So SHIFT or ALT can be used for implicit range cell selection. Any particular preference?

Keep in mind that CTRL is currently used for including keys, so CTRL+SHIFT/ALT will be needed also for implicit range selection+keys.

melieconiek:
SHIFT would be my preference

manongjohn:
I can't use ALT or SHIFT to control implicit range selection since that alters Cell dragging behavior once triggered.

Unfortunately, for now at least, I will be reverting that particular change and will revisit it later.

manongjohn:
FYI, I plan on merging this in the new few hours unless something else is found or needs to be discussed.

PR #852, opened 2022-01-02
Change Cell Mark apperance

manongjohn:
The original implementation of the Cell Mark feature from OT used a small circle to mark cells.

This PR changes the appearance of the Cell Mark to fill the cell instead, similar to how TV Paint marks cell.  The drag bar, if there is one, retains the normal color for the level type.

image


manongjohn:
Rebased and updated to account for Implicit Frame Hold changes

PR #853, opened 2022-01-02
Fix Note Text cell issues

manongjohn:
The Xsheet cell resizing and Note Text changes ported from OT introduced some issues relating to how Note Text cells were drawn.  This mostly impacted the Timeline.

This PR fixes the following issues:
- Fixed text no longer displaying when zoomed out all the way in the timeline.
- Fixed text written too high in the timeline cell because it didn't account for drag bar
- Fixed the timeline indicator in Frame 1 being offset by 1 pixel
- Fixed the Note text cell height in the timeline being 1 pixel too tall
- Fixed the frame separator between cells changing colors as timeline/xsheet was zoomed in/out
  - Also fixed this for Sound column in the Xsheet
- Fixed the drag bar not being continuous when adjacent cells are populated.
- Fixed cell color gap in Frame 1

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -2436,6 +2436,7 @@ void CellArea::drawSoundTextColumn(QPainter &p, int r0, int r1, int col) {
     QPoint xy;
     int markId;
     QRect nameRect;
+    bool isEndOfRange;
   };
 
   auto getCellInfo = [&](int r) {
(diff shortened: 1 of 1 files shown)


artisteacher:
Frame 1 still has a 1 pixel? gap below the dragbar, at the left of the frame:
Screen Shot 2022-01-11 at 8 52 16 PM

Otherwise, this looks much better.

manongjohn:
I have a fix for the start of the dragbar overlapping the cell's frame that happens on all levels. It's in the Various Bug Fixes PR, though it doesn't quite fix Frame 1.

PR #854, opened 2022-01-03
Various bug fixes

manongjohn:
This PR will fix a few minor bugs I've run across.  The following have been fixed:

- Fixed default Shortcut Preference preset still pointing to `defopentoonz` when it should be `deftahoms2d`
- Fixed Level Strip selection box referencing `- No Current Level -` in some places when it should be `- No Level -`
- Fixed undoing `Cut` of selection of only empty cells. Cut would shifts cells up but when you `Undo`, it doesn't shift cells back down
- Fixed undoing a `New Level` creation in an empty frame in middle of existing level. `Undo` shifts cells up incorrectly instead of restoring prior empty cells
- Fixed undoing `Each 2/3/4` when you select frames less than the `Each #` (example: select 2 cells, but choose Each 3). Nothing happens, but if you `Undo` the action, it adds a new cell incorrectly.
- Fixed start of drag bars starting over frame
- Fixed undoing column loaded PSD file leaving empty columns behind when it should remove the added columns
- Fixed how the X for end of levels and stop frames were drawn incorrectly

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -111,12 +111,14 @@ void copyCellsWithoutUndo(int r0, int c0, int r1, int c1) {
 
 bool pasteCellsWithoutUndo(const TCellData *cellData, int &r0, int &c0, int &r1,
                            int &c1, bool insert = true,
-                           bool doZeraryClone = true) {
+                           bool doZeraryClone  = true,
+                           bool skipEmptyCells = true) {
   TXsheet *xsh = TApp::instance()->getCurrentXsheet()->getXsheet();
   if (!cellData) return false;
   if (r0 < 0 || c0 < 0) return false;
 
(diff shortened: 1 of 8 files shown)


artisteacher:
On a related note, when importing a PSD with layers as columns (no sub-scene), using undo leaves empty layers behind.

artisteacher:
The PSD import undo fix partially works - the empty layers go away. However there's still a problem:

1. Import a PSD with layers as a column (no sub-xsheet)
2. Undo the import
3. Immediately draw in the viewer to create a new smart raster level (don't click in the Timeline)

The new level ends up in a new layer that is placed after the PSD layers would have been, effectively adding the empty layers back in

manongjohn:
When the columns are created, the current cell focus is in the last column created. Seems when a column is physically removed, the cell focus remains in the last column created. It's more obvious on the Xsheet due to the infinite columns.

That behavior apparently happens whenever columns are deleted, whether that is by Undo or even the Delete command. 

I won't be addressing it at this time.

PR #855, opened 2022-01-04
Multi-Thread FFMPEG and responsive finalizing window

justburner:
Port of opentoonz/opentoonz#4205 into Tahoma2D.


manongjohn:
Although I understand the desire for the warning, the `(UNSTABLE)` label is a bit off-putting, not that rendering previously was 100% stable as it could randomly just lock up and never finish or crash for no apparent reason.

I think I would have preferred this checkbox in the Output settings dialog under Advanced settings, enabled only for types using ffmpeg.

As this is essentially an OT port, I'll merge it as is.

Thanks for this enhancement!

PR #863, opened 2022-01-13
Allow multiple key stroke shortcuts

manongjohn:
This PR enhances shortcuts by allowing you to use up to 4 key strokes to trigger an action.

For example, you could set `Play` to `P,L,A,Y` if you wanted to.  Only when you hit those 4 keys in that order will the Play action be triggered.

With this change, keys like arrow keys, numbers, home, and others that are normally blocked for shortcuts are only blocked for the 1st key. These keys can now be used in 2nd-4th keys in a sequence.

Example:
	- `P,Right (arrow)` = `Play`
	- `S,Left (arrow)` = `Short Play`

**Limitation:**

The beginning part of the sequence (1st key, 1st 2 keys, 1st 3keys) cannot match another shortcut using the same number of keys since the actions with the fewer key strokes will take priority and trigger.

Examples of bad assignments:
- 1-key matches
  - `P` = `Play`
  - `P,L` = `Short Play` => would never be reached because 1st `P` would trigger Play
- 2-key matches
  - `P,L` = `Play`
  - `P,L,A` = `Short Play` => would never be reached because `P,L` would trigger Play
- 3-key matches
  - `P,L,A` = `Play`
  - `P,L,A,Y` = `Short Play` => would never be reached because `P,L,A` would trigger Play

A warning will be provided if you try to do this.
  
You could instead do:
- 2-key
  - `P` = Not a shortcut
  - `P,P` = `Play`
  - `P,S` = `Short Play`
- 3-key 
  - `P` = Not a shortcut
  - `P,L` = Not a shortcut
  - `P,L,A` = `Play`
  - `P,L,S` = `Short Play`
- 4-key
  - `P` = Not a shortcut
  - `P,L` = Not a shortcut
  - `P,L,A` = Not a shortcut
  - `P,L,A,Y` = `Play`
  - `P,L,A,S` = `Short Play`

### DIFF
--- a/toonz/sources/toonz/shortcutpopup.cpp
+++ b/toonz/sources/toonz/shortcutpopup.cpp
@@ -80,7 +80,7 @@ public:
 //-----------------------------------------------------------------------------
 
 ShortcutViewer::ShortcutViewer(QWidget *parent)
-    : QKeySequenceEdit(parent), m_action(0) {
+    : QKeySequenceEdit(parent), m_action(0), m_keysPressed(0) {
   setObjectName("ShortcutViewer");
   setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
   connect(this, SIGNAL(editingFinished()), this, SLOT(onEditingFinished()));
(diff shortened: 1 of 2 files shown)


artisteacher:
Setting the shortcuts works well. If you have a shortcut - say E for the eraser but set "E, E" for the normal eraser and "E, S" for the segment eraser, there's a warning message about "E" being used already and has priority. It might be helpful to have an option to remove "E" when confirming the new shortcut.

I can definitely see how this approach makes it easier to set and remember a larger number of shortcuts.

However multi key stroke shortcuts don't seem super compatible with temporary tool switching. With the "E,E" example, if I try to use temporary tool switching for the eraser (just holding down E), it partially works but the eraser doesn't erase freely. It'll erase where I click, but brushing with the eraser doesn't work - too much resistance. I was honestly surprised it worked at all...

It's too bad that single stroke and multi-stroke shortcuts aren't compatible with the same starting keys. Pressing "E" once would switch to the the last state of the eraser and work with temporary switching. But pressing "E,E", "E,S" or etc. would trigger that state of the eraser, as long as the keys were pressed within a certain amount of time (something fairly quick). Not sure if that's possible at all without introducing a delay. And no one is likely to want a delay...

manongjohn:
> It might be helpful to have an option to remove "E" when confirming the new shortcut.

I didn't want to bombard people with multiple questions trying to set up a new shortcut.  Figured the warning was enough to either cancel the assignment and switch the new shortcut to something else or act as a warning that the other shortcut needs to be changed now.

> However multi key stroke shortcuts don't seem super compatible with temporary tool switching.

I'm also a little surprised it works, though in my opinion, temp tool switches should really be single key.  It is probably acting oddly because when you hold down "E", the key is being repeated constantly so it's processing every other E and breaking in between.

> It's too bad that single stroke and multi-stroke shortcuts aren't compatible with the same starting keys.

Yeah.  Shortcut keys work on immediate key press and don't wait for a key release.  I'd have to change how shortcuts are handled, essentially wait for a key release then wait some number of milliseconds to see if another key is pressed. All these delays, no matter how miniscule, would make it frustrating to use especially if you only use 1-key shortcuts.

Thanks for testing and the feedback. I'll be merging this soon.

PR #864, opened 2022-01-13
Change UI highlight color

manongjohn:
Inspired by the Tahoma theme created by @andeon found [here](https://github.com/andeon/opentoonz-themes), this PR will change the UI's highlight color to a color that is similar to colors in the "Tahoma2D" logo.

image


artisteacher:
I'm a fan!

andeon:
Nice!  Now the one there makes more sense, because now it officially mimic the Tahoma :)

PR #866, opened 2022-01-14
Implicit hold fixes

manongjohn:
This PR fixes the following issues related to Implicit Holds:

- Implicit frames that are mesh deformed do not render/preview
- Mesh Level stop frames prevent rendering/preview of deformed level's frame
- Mesh Level stop frames load in as missing frames instead of stop frames
- Cannot drawing on deformed frame where the Mesh Level stop frame is
- Insert Multiple Keys does not work on Implicit Cells and incorrectly works on stop frames
- Palette Level name shows in Stop Frame when not in Implicit mode
- End frame in Preview and Output settings cannot be set beyond last actual frame when in Implicit mode
- Cannot Render beyond last actual frame when in Implicit mode


artisteacher:
Looks good on macOS. Thanks for the swift fixes!

PR #873, opened 2022-01-18
Convert scene to/from using Implicit Holds

manongjohn:
This PR adds 2 new `Scene` menu commands to convert the current scene to use Implicit or Explicit Holds. 

image

- There is no Undo action for this.  To revert the change, you much use `Revert Scene` to restore scene from the last save.
- After conversion, the Implicit Hold setting will be toggled on or off depending on which hold type was chosen.
- When converting to Explicit Holds, if a column does not have a final Stop Frame (implicit frames are held forever) the explicit hold will end at the last populated scene frame.

### DIFF
--- a/toonz/sources/include/toonz/txsheet.h
+++ b/toonz/sources/include/toonz/txsheet.h
@@ -588,6 +588,9 @@ in TXsheetImp.
   void notifyStageObjectAdded(const TStageObjectId id);
   bool isReferenceManagementIgnored(TDoubleParam *);
 
+  void convertToImplicitHolds();
+  void convertToExplicitHolds();
+
 protected:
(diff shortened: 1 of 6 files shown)


artisteacher:
Great idea!

DarrenTAnims:
Can you convert to implicit holds, then save the project, then convert to explicit holds, then save the project. And continue doing that?

manongjohn:
> Can you convert to implicit holds, then save the project, then convert to explicit holds, then save the project. And continue doing that?

Sure. You can even switch multiple times without saving in between if you need to for some reason.

I made these in part for scene compatibility with OT and older versions of T2D. This way you can work in whichever mode you prefer regardless of how the scene was originally saved.

PR #875, opened 2022-01-23
Implicit hold fixes 2

manongjohn:
This PR fixes the following Implicit Hold issues:

- Refactored Implicit Holds
- Fix Fx preview on implicit cells with no stop marks
- Prevent stop frames from autostretching
- Add Insert/Remove Multiple Stop Frames commands to Scene Frames context menu
- Fix freezing when Shift-down to next frame skipping infinite holds
- Fix freezing with Autoexpose on infinite holds
- Fix level name missing from cell if preceeded by stop frame
- Fix editability of level under mesh stop frame

### DIFF
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -348,11 +348,13 @@ Return true if cell in \b row is empty.
 */
   bool isCellEmpty(int row) const override;
 
+  bool isCellImplicit(int row) const;
+
   /*!
 Return cell in \b row.
 \sa getCells and setCell()
 */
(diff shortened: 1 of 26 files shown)


RodneyBaker:
I did a little testing on this build.
Thanks for the global stop frame add/remove.
It seems to be working as expected.

Initially I did have trouble where stop frames didn't seem to want to remove themselves.
I am now convinced that I must have accidentally been selecting one of the other options near the one intended.
These are very similar and might be confusing to hit via right click.

image

Stop Markers, Multiple Keys, Multiple Stop Frames, etc.
Can be a little confusing.

I assigned shortcut keys and that was working very nicely!.

Will continue testing other parts of this but
All in all looking very good.

PR #878, opened 2022-01-23
Change Default Level Type to Raster

manongjohn:
This PR will change the following:

- Default Level Type to Raster.
- Reorder the New level buttons on the Command and Quick toolbars, Level menu options and New Level popup dropdown to be: Raster, Smart Raster, Vector.
- Reorder style tabs on the Style Editor to be: Raster, Vector, Texture

Why change the default?

Most users coming from other software are used to Raster and Vector images and the properties/functions associated with them.  Smart Raster is unique to OT/T2D, as far as I know, and new users are not aware that colors do not blend without the aid of an Fx.  This may cause confusion/frustration on the onset and give users a negative impression of the software.

Note: Changes to the Command/Quick toolbars will not be seen if you import prior settings into an upgraded version of T2D.

### DIFF
--- a/stuff/profiles/layouts/settings/commandbar.xml
+++ b/stuff/profiles/layouts/settings/commandbar.xml
@@ -1,8 +1,8 @@
   <commandbar>
 
-    <command>MI_NewVectorLevel</command>
-    <command>MI_NewToonzRasterLevel</command>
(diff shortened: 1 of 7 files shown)


RodneyBaker:
A bold but very useful change.

PR #882, opened 2022-01-25
Add polyline point handle Shift/Alt modifiers

manongjohn:
This PR adds the following keyboard modifiers to adjust the cusp point handles of Polyline.

- `Shift` - Snap to angles
- `Alt` - Change angle of handle

polyline_controls


artisteacher:
I am so happy!!! You are the best!

artisteacher:
Works great! I would also add the alt/opt shortcut to the status bar when polyline is active. Right now it has the others:
Screen Shot 2022-01-25 at 6 23 01 PM

PR #885, opened 2022-01-27
Current Cell Indicator color preference

manongjohn:
This PR moves the current cell indicator color setting out of the stylesheets and into `Preferences` -> `Scene` - `Current Cell Color` so users can change it to a color they prefer since the current color, black, is hard to see on darker backgrounds.  I've left the default color as black, for now.


manongjohn:
Closing and resubmitting under a different branch

PR #886, opened 2022-01-27
Change Current Column/Cell Indicator color

manongjohn:
This PR moves the current cell indicator color setting out of the stylesheets and into `Preferences` -> `Scene` - `Current Cell Color` so users can change it to a color they prefer since the current color, black, is hard to see on darker backgrounds. The default color is Cyan.

Additionally, disabled the Current Column color indicator and hid the preference setting.  The change in color text made it difficult to read the column name and seemed too subtle to  be of any use.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -410,6 +410,9 @@ public:
   void getCurrentColumnData(TPixel &color) const {
     color = getColorValue(currentColumnColor);
   }
+  void getCurrentCellData(TPixel &color) const {
+    color = getColorValue(currentCellColor);
+  }
 
   LevelNameDisplayType getLevelNameDisplayType() const {
     return LevelNameDisplayType(getIntValue(levelNameDisplayType));
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -133,6 +133,7 @@ enum PreferencesItemId {
   syncLevelRenumberWithXsheet,
   currentTimelineEnabled,
   currentColumnColor,
+  currentCellColor,
   levelNameDisplayType,
   showFrameNumberWithLetters,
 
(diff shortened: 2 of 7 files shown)


artisteacher:
I always think it's kinda funny that there's a colors tab but not all of the colors are actually there. I get it - onion skin colors should be with the onion skin settings and etc. However I wonder if the colors tab in preferences should have a more specific name. It seems that the colors are all viewer specific?

Also, would it make sense for the current column color and current cell color to be the same?

manongjohn:
The `Colors` tab appears to only deal with the viewer, I think. It is a bit misleading as I'd go there for colors first before looking elsewhere. Not sure `Viewer` is an appropriate as I think there might be some viewer preferences are located in other tabs.

I was thinking of setting them to the same color, though yellow is my least favorite, or were you thinking only 1 color preference setting for both?

artisteacher:
Maybe  `Viewer Colors`?

I like using a cyan myself (ie. 100,250,250). I thought perhaps the new UI highlight color could work but the contrast might not be enough in some cases.

It could go either way. 1 color preference, or a checkbox for them to use the same color would make sense. Though it's not a huge deal having them separate.

manongjohn:
Doing some additional reflection on this, in particular the current column indicator.

I never did like how the current column indicator would just change the column name's text color.  Depending on the color and level type you were on it was either too subtle or it made the column name hard to read.  Seems even worse using `Light` theme.

Thinking of changing it and only having 1 current color indicator.

Ideas for changing it:
- Tint column name area similar to being selected when it's the focus.
- Highlight box around the column name area like the highlight box for current cell focus, using a common color indicator.
- Remove the current column indicator.

Thoughts/suggestions for or against the change are welcome.

artisteacher:
I don't have strong feelings about the approach, beyond the fact that there's plenty of room for improvement now.

It feels a bit inconsistent across the UI. The active palette swatch and nodes in the schematic both have very obvious borders that show the selection. Contrastingly the active column and active cell are pretty subtle.

PR #888, opened 2022-01-27
Type Tool for Raster levels

manongjohn:
This PR enables the Type Tool for Raster levels.


artisteacher:
Screen Shot 2022-01-27 at 6 35 13 PM
I did notice that vertical text behaves interestingly:
Screen Shot 2022-01-27 at 6 35 33 PM
I would expect it to just continue vertically - or go to the right if there's not room. There's no way to control the text box size to control that either. This isn't new - I just hadn't noticed.

PR #891, opened 2022-01-28
Hex color names editor

justburner:
Port of opentoonz/opentoonz#4232

-----

Continuation of a feature that was left incomplete.
Now the user can modify the hex editbox color names inside the program itself.

- Refactored code, now all Hex implementation have its own C++ file.
- Hex editboxes now supports auto-complete of color names.
- Editor can be started by clicking the hamburger and choosing `Hex color names...` on the very bottom.
- Only personal entries can be managed, default colornames.txt can be viewed but can't be modified inside OT.
- All personal entries can be imported and exported, when importing the user can choose to replace or merge existing entries.
- Some characters in name aren't allowed: `\` `#` `<` `>` `"` `'`

### DIFF
--- a/toonz/sources/toonzqt/hexcolornames.cpp
+++ b/toonz/sources/toonzqt/hexcolornames.cpp
@@ -1,0 +1,953 @@
+#include "toonzqt/hexcolornames.h"
+
+// TnzLib includes
+#include "toonz/toonzfolders.h"
+
+// TnzCore includes
+#include "tconvert.h"
+#include "tfiletype.h"
+#include "tsystem.h"
+#include "tcolorstyles.h"
+#include "tpalette.h"
+#include "tpixel.h"
+#include "tvectorimage.h"
+#include "trasterimage.h"
+#include "tlevel_io.h"
+
+// Qt includes
+#include <QCoreApplication>
+#include <QPainter>
+#include <QMouseEvent>
+#include <QLabel>
+#include <QToolTip>
+#include <QTabWidget>
+#include <QCheckBox>
+#include <QFileDialog>
+#include <QMessageBox>
+
+using namespace DVGui;
+
+//-----------------------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


artisteacher:
Love the autocomplete!

Is there a reason that the highlight has a larger margin on the left but not the right?
Screen Shot 2022-02-03 at 5 15 41 PM

justburner:
> Love the autocomplete!
> 
> Is there a reason that the highlight has a larger margin on the left but not the right? Screen Shot 2022-02-03 at 5 15 41 PM

I have no clue, i guess Qt reserves that space for an icon even if one isn't supplied. 🤷

toonz/sources/toonzqt/hexcolornames.cpp
```diff
@@ -0,0 +1,953 @@
...
+  m_hexLineEdit    = new HexLineEdit("", this);
+  m_hexLineEdit->setObjectName("HexLineEdit");
+  m_hexLineEdit->setFixedWidth(75);
+
+  // Main default color names
+  QGridLayout *mainLay = new QGridLayout();
+  QWidget *mainTab = new QWidget();
+  mainTab->setLayout(mainLay);
+
+  m_mainTreeWidget = new QTreeWidget();
```
manongjohn, line 468: To get rid of the space left of the name in the `Name` column, set the following:

```
 m_mainTreeWidget->setRootIsDecorated(false);
```
Same for `m_userTreeWidget`

manongjohn: @justburner Would you be able to make this simple change to fix the spacing?  Once it's fixed, I will merge this.

justburner: Ops sorry... got distracted with other projects.
I'll fix it asap.

manongjohn:
LGTM.  Thanks for porting it over and applying the fix.

PR #892, opened 2022-01-28
Stop motion play sound on capture and deciseconds resolution

justburner:
Port of opentoonz/opentoonz#4233

-----

This PR improves the Stop Motion a bit by the following:

- Added option to play a click sound when an image is being captured
- Time lapse interval setting is now saved
- Interval and review times are in deciseconds resolution (1/10 of a second)
- Added `Use Numpad Shortcuts When Active` tooltip to show which and what the shortcuts do
- Now `Use Time Lapse` checkbox is a checkable groupbox.

Warning: There's a chance of ears bleed if click sound is played on headphones ;)


artisteacher:
I'm getting a crash on open in the action build on macOS.
[sm-crash.txt](https://github.com/tahoma2d/tahoma2d/files/7998585/sm-crash.txt)

justburner:
That's strange, crash seems inside Qt itself in QWidget::hide() for no apparent reason, since there's no crash on Windows and i don't have access to MacOS platform is impossible for me to figure out why it really crashes.

Major changes there were the checkable QGroup, each QCheckbox having their own label text and using DoubleField for better precision.

Nevertheless the report was valuable to point a clue of why it crashed... worse case scenario (if we can't find the real cause) is reverting back the UI redesign.

artisteacher:
OT's latest nightly opens just fine. However I do get a crash when opening the stop motion controls window. Crash report looks the same.

I would try to build this myself and troubleshoot on this end, but I haven't been able to get a build environment working yet on my new M1 mac. I'm not sure if it's the new mac itself or Monterey that's the problem. Either way, the building directions and current scripts aren't compatible with my new setup and I don't know enough to get it working on my own.

toonz/sources/stopmotion/stopmotioncontroller.cpp
```diff
@@ -1374,32 +1375,50 @@ StopMotionController::StopMotionController(QWidget *parent) : QWidget(parent) {
...
 
     m_subsamplingFld = new DVGui::IntField(this);
     m_subsamplingFld->setRange(1, 30);
     m_subsamplingFld->setDisabled(true);
 
-    m_placeOnXSheetCB = new QCheckBox(this);
+    m_placeOnXSheetCB = new QCheckBox(tr("Place on XSheet"), this);
     m_placeOnXSheetCB->setToolTip(tr("Place the frame in the Scene"));
 
-    m_directShowLabel = new QLabel(tr("Use Direct Show Webcam Drivers"), this);
```
manongjohn, line 1396: @justburner There is reference to this variable on line 2052 which is only used on non Windows builds. 

```
#ifndef _WIN32
  m_directShowCB->hide();
  m_directShowLabel->hide();
#endif
```

Since you don't create/initialize this label, it crashes when it tries to hide it.

justburner:
Nice finding, that small hidden piece of code completely escaped my radar o.o
Of course Visual Studio graying out the code didn't help either... oof.

I now see that OpenToonz also have that sneaky code in there, i'll proceed to patch it right away in both software.

artisteacher:
No longer crashing. Seems to work just fine on macOS.

PR #894, opened 2022-01-28
Fix implicit scene saving as explicit

manongjohn:
Accidentally broke scene saving using implicit holds as part of PR #875.  Causes it to save using explicit holds.

This PR will fix it. 

I will merge this once this PR builds successfully as a quick fix.


manongjohn:
Tested artifact and seems to be working properly now.

PR #897, opened 2022-01-30
Update ffmpeg to v5.0

manongjohn:
This PR will update all builds to include ffmpeg 5.0 (LGPL)

Windows and Linux versions will use version n5.0 builds from https://github.com/btBN/FFmpeg-builds.
MacOS was built locally using the 5.0 source from https://github.com/FFmpeg/FFmpeg

NOTE: Builds of OpenCV used for building macOS/Linux versions will continue to still use ffmpeg 4.3.1 for now.

### DIFF
--- a/ci-scripts/linux/tahoma-buildffmpeg.sh
+++ b/ci-scripts/linux/tahoma-buildffmpeg.sh
@@ -16,7 +16,7 @@ sudo make install
 cd ..
 
 echo ">>> Cloning ffmpeg"
-git clone https://github.com/tahoma2d/ffmpeg ffmpeg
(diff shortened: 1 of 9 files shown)


manongjohn:
I've added some logic for mp4 rendering to use `libx264` or `libopen264` codec if the ffmpeg being used supports it.  This should fix the issue with mp4s not playing correctly in Discord (Windows app) and maybe some other places as well.

For now it's hard coded to whichever it finds, but I intend on adding a `Codecs` option to the `Options` dialog box for any types rendered via ffmpeg.

Due to updating the `tahom2d\FFmpeg` repo with source for ffmpeg 5.0, I will need to merge this before I can push any more code changes or merges otherwise macOS/Linux builds will fail.

PR #902, opened 2022-02-11
Restore ability to change root pin

manongjohn:
In light of recent discussions regarding changing the IK root pin, this PR will restore the ability to change the root pin.


manongjohn:
Canceling this in favor of the same restoration being introduced in OT but done with some changes.

Issue #900, opened 2022-02-06
Tahoma Crash(And OT)-New knowledge


Tragustauro:
@[manongjohn] and [artisteacher]
Hello.

Thank you for putting in time to do this. 
I have both Opentoonz and Tahoma as I am trying them both out for my productions. 

Both have issues.

I am actually not 100 sure the crashes is caused by the imported sounds. Both mp3 or wav.

I want to use the softwares primarely as Pencil line tests. And I have been trying many many things and run into issues.
It seems like I have a conflict with OT and Tahoma when connecting to my HD webcam.
I am using the Stopmotion function. 
On the first "call for connecting to the web cam it all seems fine. I used it close the "live view" and move to the drawing desk.
I then come back and wants to turn on the "live view again" and everything freezes up and crashes.

I am on Win7 pro. (Wich I love) and Am using logitech 9000 webcam.

Thanks in advance.
T.

_Originally posted by @Tragustauro in https://github.com/tahoma2d/tahoma2d/issues/868#issuecomment-1030891189_

Tragustauro:
Hello. Sorry about my ignorance. First time using Github.

I think I found out about the crash. For both OT and Tahoma.

webcam logitech 9000 
Newest driver updated. 
Webcam software works like a charm. Crisp picture .
Tahoma and OT installed.
Everything works like a charm.
On start up I disregard the startup menu where you can name your scene. I close it.
Activate "Stopmotion" control/function.
Webcam activated and no problem.
I test out my scene. (Paper-pencil linetest)
I do either of two things.

1: I save the scene
2: I close down with out saving.

On the next startup I have the option of starting all over again from fresh. Or I can choose 
"Recent Scene"
When I choose start all over agian in a new scene= Noproblem.
When I choose "RecentScene"  Tahoma starts to work on opening the "recent Scene" I chose. And Tahoma freezes/Crashes. For some reason I might get lucky that the scene actually starts up. But as soon as I touch "Stop-motion" it crashes.!???

So to work with out crashes I need not to save anything but start over.

Btw. Once the crash has been done I have no more connection to the webcam eventhoug I see it is still installed. Even in The webcams software I can not reach it!!..

I will try tonight to make a video of he problem

Tragustauro:
Hello. Sorry about my ignorance. First time using Github.

I think I found out about the crash. For both OT and Tahoma.

webcam logitech 9000
Newest driver updated.
Webcam software works like a charm. Crisp picture .
[...]

PR #904, opened 2022-02-13
Fix reloading stop motion scene crashing application

manongjohn:
This PR attempts to fix the crash reported in #900.

When opening a stop motion scene, T2D attempts to connect to the camera and immediately turns on `Live View` mode, however the camera may not be connected.  On some systems, this may cause a crash at startup.  

This change will not automatically turn on `Live View` mode at startup.

### DIFF
--- a/toonz/sources/stopmotion/stopmotion.cpp
+++ b/toonz/sources/stopmotion/stopmotion.cpp
@@ -2429,6 +2429,7 @@ bool StopMotion::loadXmlFile() {
     }
     xmlReader.readNext();
   }
+  emit(updateStopMotionControls(webcam));
   return true;
 }
 
@@ -2976,7 +2977,8 @@ void StopMotion::changeCameras(int index) {
 #endif
 
   // close live view if open
-  if (m_liveViewStatus > LiveViewClosed) {
+  int liveViewStatus = m_liveViewStatus;
+  if (liveViewStatus > LiveViewClosed) {
     toggleLiveView();
   }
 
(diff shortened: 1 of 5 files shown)


manongjohn:
@Tragustauro I've pushed a change that will connect to the camera on a scene reload and keep it on instead of rapidly connecting/disconnecting.  I'm hoping this will help your issue.

When it finishes building, please download the artifact from here and let me know if it helps at all https://github.com/tahoma2d/tahoma2d/actions/runs/1853526828

Tragustauro:
@manongjohn 

Hello sir.
I actually think you did it!!!!.


At a first teset opening T2D->Selecting cam->Turn live view on-> Shoot a couple of frames-> Scroll thru frames just shot->Save all->Close downT2D->Open T2D-> Select last open scene->Cam->Shoot......Etc etc..

Seems to work like a knife thru butter. I will work with it the coming days and I´ll let you know if anything. This is so cool.

May I ask? Do you also support OpenToonz? Because it is tha same issue and that could be cool to make that work also.


T2D aslo feels more smooth now and correct. (If a software can be smooth)


Awesome!!!
I will return with feedback.

Thanks!! 
Ps: I added a picture of the file I downloaded and ran. I assume it is correct!?

CaptureT2D402mb

Tragustauro:
By the way I noticed the aspect ratio has changed!?.. Before the live view frame was filling out the entire cam frame

CaptureT2D402Amb
.

Tragustauro:
CaptureT2D402Bmb

manongjohn:
If it isn't crashing, then I'd say you downloaded the correct one. Good to hear the crashing has stopped. :)

I am focused primarily on developing Tahoma2D.  While I bring OpenToonz changes into Tahoma2D, I very rarely port my T2D changes to OT as I don't have time to maintain both applications.  My changes can be ported back to OpenToonz by another developer if one wishes to.

As for the aspect ratio, I'm not sure.  I didn't intentionally change anything related to it.  Double check the resolution is correct in the Stop Motion settings.  Keep in mind, the  default camera/scene size is 1920x1080 but your camera appears to support only 1600x1200 according to your screenshots you provided on the issue.  I don't see a difference between those and what you just posted here.

manongjohn:
I will be merging this in the next day or 2 so it will be in any nightly after that.  Once it makes it to the nightly, it will be in the next release.

You can continue keep using the working one. Switch to a nightly later when you feel like it to get any new changes since then.

Not sure when the next official release will be, yet, but will likely happen in a couple months.

Tragustauro:
Thank you.

PR #908, opened 2022-02-19
Update Windows builds to use MSVC 2019 with Qt5.9.7

manongjohn:
On March 15, 2022, Github is deprecating the `windows-2016` build environment that uses MSVC 2017. As such, we need to move the Windows build environment to `windows-2019` which uses MSVC 2019

Unfortunately Qt 5.9 is not officially supported on this version and moving to Qt 5.15 caused significant problems for random users that I could not duplicate. 

As I am not ready to deal with Qt 5.15 issues for the T2D 1.3 release, I've had to compile Qt 5.9.7 from source and package it in the `tahoma2d\qt` repo for Github builds to build with on MSVC 2019.

This will buy us a bit more time until I am ready to switch to 5.15 (or later) and fix the issues that were being reported.

I have confirmed locally that I could build, package and run T2D with this configuration.

Additional verification by other users would be extremely welcomed.

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   Windows:
-    runs-on: windows-2016
+    runs-on: windows-2019
     steps:
       - uses: actions/checkout@v2
         with:
--- a/ci-scripts/windows/tahoma-build.bat
+++ b/ci-scripts/windows/tahoma-build.bat
@@ -10,16 +10,16 @@ IF NOT EXIST build mkdir build
 cd build
 
 REM Setup for local builds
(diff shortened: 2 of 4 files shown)


manongjohn:
Appears to be an issue with windeployqt not finding the QT DLLs to include in the package.  Working on fixing.

manongjohn:
Ok. Problem fixed. 

For some reason my anti-virus software (Avast) keep blocking the artifact download. Keeps thinking it is infected by "Win64:Evo-gen [Susp]" but it isn't.

I don't think it likes the `lzocompress.exe` what was compiled and packaged with the release. It would sometimes quarantine it when I was testing Tahoma2D that was compiled in MSVC 2019 (usually when I deleted it).

PR #909, opened 2022-02-20
Revert enhanced flipbook playback

manongjohn:
Due to slow Xsheet playback issues (and to some degree the Timeline) that was introduced when the Enhanced flipbook playback OT patch added in PR #874, this PR will revert those changes and related changes, for now.

Should OT ever fix the issue, I'll revert this revert and apply the fix.


melieconiek:
Seems to work, thank you. Hopefully this gets properly sorted later.

manongjohn:
Canceling this as Shun fixed the issue in OT by reverting the changes that impacted the viewers, but keeping the flipbook change. It seemed to fix the problem.

Will port over the OT fix.

PR #912, opened 2022-02-22
Stop Motion Camera Calibration

manongjohn:
This PR is based on OpenToonz's Camera calibration for the Camera Capture feature by @shun-iwasawa  (opentoonz/opentoonz#4250)

The Camera Capture feature was removed from Tahoma2D so I applied the changes from the OT PR to the Stop Motion Controller feature instead.  The `Calibration` setting can be found on the Stop Motion Controller's `Settings` tab.

image

It would be appreciated if someone with a Canon DSLR can verify this change does not show the new control under Canon's settings as well as does not impact operations with it.

-----

### How to calibrate camera

1. Prepare the checkerboard
    - Print out `checkerboard.tif` found in `tahomastuff/library/camera calibration`.
    - It would be better to paste the checkerboard pattern on some flat panel like cardboard.

2. Capture the pattern
    - In the Stop Motion Controller's `Settings` tab, enable `Calibration` box
    - Click `Start calibration`.
    - Take 10 snapshots of the checkerboard pattern in various positions and angles.
calib_checkers
    - Once the capturing is done, calibration will be automatically applied whenever the `Calibration` box is checked.
   

3. Export & load calibration settings
    - Calibration settings will be saved under `tahomastuff/library/camera calibration` with name "[MACHINENAME]\_[CAMERANAME]\_[RESOLUTION].xml"
    - The settings file will be overwritten after finishing the new calibration.
    - You can export the file for backup, or load it afterwards.


artisteacher:
So I'm not seeing my canon dslr recognized at all with this action build. I did copy the EDSK over, since it wasn't there by default. Is canon turned off when the action build is building?

I don't have this issue with the latest nightly build.

manongjohn:
Oops. I completely forgot that PR builds do not have Canon support. Sorry about that.

I will build and package locally and post here for you to test with as soon as possible.

manongjohn:
@artisteacher Here is a link to download it from my personal OneDrive account: https://1drv.ms/u/s!Au-akGfpWFmGi_cLTmcD2fRWMOB3Aw?e=JddCp3

Let me know if you have trouble accessing it.

artisteacher:
My camera is now recognized. Got a crash when clicking on the capture button after trying to start the calibration.
[calibration.txt](https://github.com/tahoma2d/tahoma2d/files/8181277/calibration.txt)

manongjohn:
> My camera is now recognized. Got a crash when clicking on the capture button after trying to start the calibration. [calibration.txt](https://github.com/tahoma2d/tahoma2d/files/8181277/calibration.txt)

Of course it's crashing somewhere in opencv, but not telling me how it got there through T2D.

Just to verify something...can you capture normally without calibration?

artisteacher:
Isn't that always how it goes!

Yes, normal capture works just fine.

manongjohn:
Ok..I've discovered the issue.  I put in logic for Canon captures that isn't really being hit after capturing an image so when it tries to process an image to build calibration information, it crashes because my logic actually never got the image.  The logic that I put in would only work when it's downloading images from the camera under some condition that I am not aware of.

Canon captures images and immediately converts it into a Raster image, never going through OpenCV, the way webcams capture and process images.   I'd have to convert from Raster to OpenCV format, process it for calibration or apply calibration, then back to Raster in order for this to work.

Unfortunately, I think it's going to be too difficult for me to make this work without a Canon camera at my disposal.

artisteacher:
Yeah, that sounds a little complicated. It sounds like this should be kept as a webcam only feature for now. It can always be added later, when you hopefully have Canon access.

artisteacher:
I would recommend adding the capture sound when capturing for calibration.

manongjohn:
Actually, there is I thing I can try even though i don't have a canon camera.  I'll take the Raster output that is generated after webcam processing and try and convert as if it was generated by a canon and see if I can get the logic to work.  If it works, I'll apply the same logic to canon capture and maybe, just maybe I can get it working.

manongjohn:
@artisteacher Ok...hopefully this one works.  Please try it out when you get a chance: https://1drv.ms/u/s!Au-akGfpWFmGi_cLTmcD2fRWMOB3Aw?e=FNja0O

artisteacher:
It works!

artisteacher:
Export doesn't work:
Screen Shot 2022-03-03 at 6 43 51 PM
It looks like the xml file is actually saving in the wrong location:
Screen Shot 2022-03-04 at 10 10 22 PM
If I move the xml file to the camera calibration folder, it's no longer recognized by T2D. If I try to load it from the new location, clicking on "ok" to load does nothing.

Btw, thanks so much for figuring out the calibration for Canons!

manongjohn:
@artisteacher Thanks for verifying it works! 

I've added the capture sound (if enabled in Options), fixed path where calibration data is saved as well as added a few hints for calibration usage.

The latest changes don't require a Canon camera to test, but if you want me to generate a build with Canon support for you to do more testing, let me know.  I would be more than happy to create it.

artisteacher:
Changes all work just fine on the built-in camera!

PR #930, opened 2022-03-16
Fix incorrect QT_VERSION checks

manongjohn:
This PR fixes the checks for QT_VERSION 5.10 and later to use the correct format.

I was using an older syntax for QT version 5.10 and later which was incorrect.  When compiled on those versions or later, the code would never become active since the check always failed.  This replaces those checks to use the proper method so the code will be enabled when compiled with the right version or later

This will only impact macOS/Linux builds as they are currently building on 5.15 whereas Windows builds still [stubbornly] use Qt 5.9.

With this change, the following will now be activated for macOS/Linux builds:
- 30bit Display Support 
- Spacing of key when exporting xsheet to PDF
- Font spacing on Type tool
- Different error messages in Record Audio

### DIFF
--- a/toonz/sources/common/tvrender/tfont_qt.cpp
+++ b/toonz/sources/common/tvrender/tfont_qt.cpp
@@ -179,7 +179,7 @@ TPoint TFont::drawChar(QImage &outImage, TPoint &unused, wchar_t charcode,
   // (21/1/2022) Use this workaround for all platforms as the crash also occured
   // in windows when the display is scaled up.
   if (chars[0].isSpace()) {
-#if QT_VERSION >= 0x051100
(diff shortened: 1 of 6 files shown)


artisteacher:
Sorry, I hadn't gotten around to checking this out before it was merged. Is there supposed to be a font spacing option in the tool option bar?

manongjohn:
No problem. As far as the font spacing goes, no.. its just how it calculates the spacing of the letters. Nothing the user can control. Just may see slight differences in spacing of certain fonts in comparison to prior version.

artisteacher:
Ah, ok. I was excited at first that we might be getting some form of kerning.

PR #933, opened 2022-03-19
Update labels to Guided Tweening

manongjohn:
This PR relabels `Guided Drawing Controls` to `Guided Tweening Controls` in an attempt to make this feature more discoverable for those looking to use Vector inbetweening.

Also updated the label in the popup to indicate it uses Onion Skin.

Aside: Will work on documenting this feature in the manual since it isn't quite intuitive to use....yet


artisteacher:
Definitely a helpful change.

PR #934, opened 2022-03-21
Tahoma branding design update

keyschlaff:
Updating Tahoma logo design listed below:
- Windows icon
- macOS icon
- Linux icon
- Tahoma splash screen,
- Tahoma startup panel logo

**New design:**
tahoma2d


jeremybullock:
There is some nice work here.  I'm biased, but I miss the stronger orange.  The original colors were inspired by the old Zune logo  (Yes, I said Zune.)  I really liked how the magenta and orange played together.

manongjohn:
Well that certainly explains the color choices then. Lol.  I always wondered how you got that color scheme.

I probably would have used cooler colors, though I have to say, the T2D colors eventually kinda grew on me a little.

artisteacher:
I like the slightly more subtle colors myself! Though I would adjust the kerning between the T and a - move them just a bit closer together.

keyschlaff:
> There is some nice work here. I'm biased, but I miss the stronger orange. The original colors were inspired by the old Zune logo (Yes, I said Zune.) I really liked how the magenta and orange played together.

My first thought with the color palette reminds me with Zune as well
I thought Tahoma favors the magenta more over orange

> I like the slightly more subtle colors myself! Though I would adjust the kerning between the T and a - move them just a bit closer together.

I can fix that if you want

melieconiek:
I'd recommend keeping the right underside of the 'T' down like it is now -- instead of trying to make it level with the other side.

I know it's not 'right,' but there's too much white space on the right side under the 'T' otherwise. It's unbalanced.

manongjohn:
Although I am ready to merge this as is, I am a bit curious what it would look like if the T was "unfixed" but still maintaining the new style.

manongjohn:
@keyschlaff  Thanks for updating the brand.

PR #936, opened 2022-03-22
Changes in stopmotion room

charliemartinez:
I suggest this layout and panels for the **stopmotion room**. I understand that Path has its tab, however we animators are used to having it visible along with the photo shooting panel. In addition, we use the vertical **x-sheet** a lot _(in fact, for me, the horizontal one could not be)._ 

Demo image


manongjohn:
Seems your PR to change the Stop Motion room includes some Debian packaging stuff which I believe we previously discussed would not be storing with the source. I cannot accept this PR because of it.

In your fork of T2D, I would recommend not making changes directly to the master. Create a branch from a clean master and make your specific changes there, then make PRs from that branch to T2D.

PR #937, opened 2022-03-22
Changes in stopmotion room

charliemartinez:
I suggest this layout and panels for the **stopmotion room**. I understand that Path has its tab, however we animators are used to having it visible along with the photo shooting panel. In addition, we use the vertical **x-sheet** a lot _(in fact, for me, the horizontal one could not be)._ 

Demo image


manongjohn:
Is there a particular need to have both the xsheet and timeline in the same room? Having both can cause some scrolling issues as they always try to keep the same frames visible.

charliemartinez:
@manongjohn  You are right and I think the same, in fact most stop motion animators are more used to the x-sheet than to the timeline. I'll fix it myself.

manongjohn:
Just curious, from a Stop motion point of view, i think this room was originally designed for capturing images, I think. I dont mind the changes if it makes sense for Stop Motion work. How is the schematic useful in this context?

PR #938, opened 2022-03-22
Stop motion room changes

charliemartinez:
I suggest this layout and panels for the **stopmotion room**. I understand that Path has its tab, however we animators are used to having it visible along with the photo shooting panel. In addition, we use the vertical **x-sheet** a lot (instead of the timeline).

demo stop motion room image


jeremybullock:
Is that on a 4K monitor?  I think some settings won't be visible with this layout on a 1080 screen.

charliemartinez:
> Is that on a 4K monitor? I think some settings won't be visible with this layout on a 1080 screen.

It is a VGA 24'' 1920x1200 16:10 screen.

This is how it looks on my 2021 Fujitsu laptop, 12.5' ​​1366x768 16:9

demo stop motion laptop

The open modules are equivalent to the ones that are open by default in DragonFrame. However, I'm happy to say that in Tahoma2D the space looks much neater. Take a look at what Dragonframe looks like out of the box on the same laptop:

demo stop motion dragonframe laptp

jeremybullock:
With the paths being one click away and with customizable rooms, I don't see a huge benefit from this change, especially since the settings get cut off with this change.  I'm also not sure why the schematic view is in there.

artisteacher:
I can't test this out directly on my 13in MacBook, since it runs on Catalina. The display is set to the equivalent of 1440x900 - the default.  I manually remodeled the stop motion room to look similar:
smallscreen
It's just way too crowded to be usable on my screen. I can barely see any of the stop motion controls. I can't fix that with this configuration due to the minimum height of the motion path panel - unless I make the schematic too short to be useful. Even then, it'll only help a little.

I don't think we should change to a layout that just doesn't fit on smaller screens. Anyone with the room, can still manually make the adjustment.

There are probably ways to make a change like this more feasible in the future. Ie. if we could collapse panels out of the way when we aren't using them...

charliemartinez:
> With the paths being one click away and with customizable rooms, I don't see a huge benefit from this change, especially since the settings get cut off with this change. I'm also not sure why the schematic view is in there.

As a computer scientist I agree with you, as an animator I think that the same applies in reverse: whoever does not need these tools can hide them. It would be rare to find a professional animator who doesn't require them. I recently opened an ISSUE asking for a feature that already existed. So I wondered why I didn't find it? how could i improve that?

The motion path is logical to be called like this and not "increment editor", as in dragonframe. Although I would have had a hard time finding it on my own, because I would not have known what name to look for it. Seeing it is easier. And it is a basic tool. Imagine hiding the pencil in the 2D drawing panel...

For a few years I have been developing a specific distribution for animators and it is not the first time that it has happened to me that some things, of very common use, go unnoticed because they are not visible from the first moment. If the intention of Tahoma2D is to be adopted by professional animators, we have been using Dragonframe as a standard and these details make the transition much easier for us, not having to adjust everything manually.

charliemartinez:
> It's just way too crowded to be usable on my screen. I can barely see any of the stop motion controls. I can't fix that with this configuration due to the minimum height of the motion path panel - unless I make the schematic too short to be useful. Even then, it'll only help a little.
> 
> I don't think we should change to a layout that just doesn't fit on smaller screens. Anyone with enough space, can still manually make the adjustment.
> 
> There are probably ways to make a change like this more feasible in the future. Ie. if we could collapse panels out of the way when we aren't using them...



> I can't test this out directly on my 13in MacBook, since it runs on Catalina. The display is set to the equivalent of 1440x900 - the default. I manually remodeled the stop motion room to look similar: smallscreen It's just way too crowded to be usable on my screen. I can barely see any of the stop motion controls. I can't fix that with this configuration due to the minimum height of the motion path panel - unless I make the schematic too short to be useful. Even then, it'll only help a little.
> 
> I don't think we should change to a layout that just doesn't fit on smaller screens. Anyone with enough space, can still manually make the adjustment.
> 
> There are probably ways to make a change like this more feasible in the future. Ie. if we could collapse panels out of the way when we aren't using them...

In the distribution of the image that you have remodeled I see everything. What would be the controls that you are not being able to visualize?

I insist, if we compare the current industry standard (of stopmotion) with Tahoma2d, Tahoma2d panels are still more comfortable and usable thanks to their vertical sliders (see image of DragonFrame that I shared above).

**Ideally,** it would not be necessary to add the FX motion path: we could draw directly on the screen and eliminate the need to use the Stage Schematic panel.

By the way, the timeline is very uncomfortable to work with stop motion, making cycles with the x-sheet is more usual.

manongjohn:
Still not sure you need both Schematic Viewer and Motion Path panel in this room.

From the Schematic viewer, if you add a Path or select an existing Path, then either hit the shortcut 'B' for Brush Tool or use the Toolbar, you can start drawing without the need for the Motion Path panel..  Having the schematic only, you lose quick access to change the steps and step spacing. Perhaps double-clicking the Path in the Schematic should open a Motion Path panel directly to that path, like a properties window? 

On the flip side, just having the Motion Path panel should have access to all the tools you need for creating and managing the Path (selection, brush, control point editing, pinch, geometric tools).  For this reason, I would think the Motion Path panel would suffice.  You can reclaim some space by doing away with the Schematic and the Toolbar.

I'm more concerned about users on smaller screens as this will likely look extremely cramped. I know we can't adjust for every screen size, though we should definitely aim towards something that is more common, and/or maybe 1 resolution lower.

Something to keep in mind, since we removed the Camera Capture feature, regular animators could also be using this room for Pencil tests.  However, since the room is labeled "Stop Motion", I don't mind making it more Stop Motion-centric...we are touting T2D as 2D Animation and Stop Motion software afterall. :)

artisteacher:
> Perhaps double-clicking the Path in the Schematic should open a Motion Path panel directly to that path, like a properties window?

Great idea! Or even the reverse - double-clicking a path in the motion path panel could open a floating schematic with the path node active.

manongjohn:
Another suggestion, if I may....

If you insist on having a Toolbar in that room, you can probably do away with the actual Toolbar by replacing the Viewer with a ComboViewer and toggle showing the Toolbar in it.  In the proposed layout It can save a little more space on the screen and make use of dead space in the viewer above the fit-to-window camera image.

artisteacher:
@manongjohn - should we be able to toggle showing the toolbar in the combo viewer? I thought it just always displayed. We wouldn't even need both combo viewer and viewer options if the tool option bar and toolbar could be toggled on and off.

manongjohn:
@artisteacher Can't remember if the toolbar is on by default or not in a ComboViewer.  You can toggle it in the comboviewer's context menu under `GUI Show / Hide` sub-menu.

manongjohn:
> Why do we have both combo viewer and viewer panels available then? Is there a performance difference?

Honestly I wondered that myself for the longest time.

I remember there was a slight difference when I was comparing them also thinking ComboViewer would probably be a better choice in many cases. I forget exactly what, but users wouldn't necessarily care or even notice that difference.  The only interesting thing about the ComboViewer, beyond the built in toolbar and options bar, is that it provided a horizontal toolbar option. I think the ComboViewer was used in one of the Studio Ghibli rooms which was probably used to make the room layout more optimized for whatever workflow it was.  

Even more Off topic: I could have sworn when OT was first released, there was separate options for a Vertical or Horizontal Toolbars but it disappeared in the next version.  I don't even see the code for it.  Maybe I was imagining it. :)

charliemartinez:
@manongjohn 

> Still not sure you need both Schematic Viewer and Motion Path panel in this room...

Maybe we could have a room for **stopmotion** and another lighter **pencil test** room.

Regarding the need for the **schematic viewer**: I have noticed that if I don't add a motion path module, the tool only draws a raster level:

whithout

 Not only do I need to create a motion path module on the schematic, but I also have to remember to click on it (select it).

width

There are quite a few steps and if they are not visible, it is worse.

Can this be simplified? I would also like to reduce the elements on the screen.

charliemartinez:
@artisteacher 

It depends on what we mean by "worse".

I showed above a screenshot of the 12' screen of my laptop, with which I usually work, and I don't find the use of vertical sliders uncomfortable at all. I do find it uncomfortable to have to open a panel every time I change the scale: each movement we photograph has its own scale. For that we use the motion path: there is no automatic interpolation in stop motion, we use it as a guide to take the pictures and we use it all the time. 

The progress has been great. It is wonderful that we can already draw these scales and that we can already draw more than one. It would be great to even be able to see more than one at the same time (I haven't gotten that yet).

I studied animation film at a state school [IDAC](https://es.wikipedia.org/wiki/Instituto_de_arte_cinematogr%C3%A1fico_de_Avellaneda), I was teacher there,  where I am very interested in the implementation of free software, I also took workshops with animators like [Juan Pablo Zaramella](https://es.wikipedia.org/wiki/Juan_Pablo_Zaramella) and I can assure you that this tool is more necessary, and capture of course.

Each animation technique has its characteristics, its particular workflow. It is not my intention to open a discussion, mine are just suggestions. I am only here in case it helps you to know how to work in stop motion: with scales and with verticals x-sheet: this is how we are trained and this is how we can work outdoors when we don't have a computer (with printed x-sheets) and then continue with our computer. What's more, we only use one level, although that level is very necessary and not so much the number of frames that we can see, but the way of reading them: they teach us to read them in this way, there are codes like in the scores. Stop motion is still the most traditional technique of all, rearranging the head to more suitable flows for post-production is not so easy. We use timeline in post-production, although not when animating (stopmotion).

I am not against innovation, although there are already mature tools that present a different workflow, such as **qstopmotion**, or other lightweight ones that are very useful for pencil testing, **boats-animator**, with a beautiful UI on an aesthetic level, such as boats-animator. Tahoma2D opens up hope for professional animators who are already used to a very extended way of working and even without my modifications this is possible, which makes me very happy, although it will cost a lot to transmit it to others, it will require more documentation and tutorials... What if we put everything in sight. They are only suggestions.

@manongjohn 

Maybe it's a bug or I'm not knowing how to use the tool (most likely) but I'm not getting it. If I don't create a module in the schematic, the motion path is just a trace. It is not activated neither by clicking nor selecting it

demo stop motion room image

charliemartinez:
> > Perhaps double-clicking the Path in the Schematic should open a Motion Path panel directly to that path, like a properties window?
> 
> Great idea! Or even the reverse - double-clicking a path in the motion path panel could open a floating schematic with the path node active.

Great !!! Those kinds of ideas are wonderful, I'm a favor 100%. Understanding that it requires a lot of programming work, more complex than rearranging a configuration file. For now, try to make the most of what you have. Very good job guys, I am very grateful indeed.

manongjohn:
Quick usage.  Hope this helps:

motionpaths


It may not show it is selected in the Schematic, but paths are visible on the screen unless you hide them.

If this doesn't help, you'll need to record what you are doing and share it so I can better understand what problems you are having.

charliemartinez:
@manongjohn 

WOW I never realized that "+" button existed. Thousands of thanks! It's great to be able to work in this simple way, without the need for the schematic panel. My workflow now looks like this (remember that in stopmotion we only use one level of the x-sheet):

demo stop motion room image

artisteacher:
@charliemartinez - I'm just saying that going from seeing the whole panel to seeing only one or 2 lines of the panel is going backwards in usability. I can't even see the first 2 options fully! If enough was visible to start with, the vertical scrolling wouldn't be a problem. It's just that with the schematic window too, the vertical scrolling is crazy excessive. I'd personally take switching panels over that any day.

The latest screenshot looks much more reasonable.

charliemartinez (commented):
I remove the schematic module

manongjohn:
LGTM.  Thanks for the improvement!

PR #939, opened 2022-03-24
Default column names to 1st level name

### LINKED ISSUE #114: Change Column Names
Tahoma is Timeline oriented over the Xsheet. Col1, Col2 & etc. make a lot of sense when the Xsheet is the priority - but not in the Timeline.

My favorite option is to name columns/layers with the name of their first level - just like how the thumbnail images are determined. Multiple levels can end up in a column/layer but it is less likely now that additional noncontiguous frames are part of the same level. This could also make it easier to identify where specific levels are located if they look similar or have thumbnails that are too small to identify.

manongjohn:
This PR will default an empty column's name to the name of the 1st level added to the column. (Resolves #114)

This should occur when adding a level to an empty column via...
- New Level popup (OK & Apply)
- Load Level popup
- Drag/drop levels from Scene Cast
- Drag/drop files from File Browser
- Drag/drop files from outside T2D (i.e. Windows File Explorer)
- Copy/paste of external image from another software
- Copy/paste of existing cell in the timeline
- Autocreation (brush tool, geometric tool, type tool)
- Create Blank Drawing
- Adding Zerary Fx (existing behavior)
- Adding Note level (existing behavior)
- Dragging existing cell (existing behavior)

If a column already has a level and a new level is added to it, the column name will not change
If a column has multiple levels and the 1st level is removed, the column name is not change

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -28,6 +28,8 @@
 #include "toonz/mypaintbrushstyle.h"
 #include "toonz/preferences.h"
 #include "toonz/toonzfolders.h"
+#include "toonz/tcolumnhandle.h"
+#include "toonz/tstageobjectcmd.h"
 
 // TnzCore includes
 #include "tgl.h"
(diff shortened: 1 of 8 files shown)


artisteacher:
Thank you! You are amazing. I've been waiting for this since I first used OT!

DarrenTAnims:
At last!!! I've been asking for this since I first started using OpenToonz. Great addition. Thanks @manongjohn

artisteacher:
I just noticed that the layer name still displays as Col1 or etc. when in the act of parenting with the Skeleton Tool in Build Skeleton mode. It would be helpful if the layer name displayed here too:
Screen Shot 2022-03-29 at 6 25 38 PM

manongjohn:
> I just noticed that the layer name still displays as Col1 or etc. when in the act of parenting with the Skeleton Tool in Build Skeleton mode.

I wouldn't be surprised if there are a few other places like that. It will likely be more obvious now that this change is merged in. Probably should open a new issue so we have a place to add any more we might find.

PR #940, opened 2022-03-25
Fix small text on high res monitors

manongjohn:
This PR scales the size of the text in the Viewer for high res monitors otherway they appear too small and are hard to read.

The following will be fixed:
- Viewer check indicators (i.e. Transparency Check, etc) 
- Tracker tool object label
- Skeleton tool frame indicator

### DIFF
--- a/toonz/sources/tnztools/skeletontool.cpp
+++ b/toonz/sources/tnztools/skeletontool.cpp
@@ -30,6 +30,7 @@
 // TnzQt includes
 #include "toonzqt/selection.h"
 #include "toonzqt/selectioncommandids.h"
+#include "toonzqt/gutil.h"
 
 // TnzTools includes
 #include "tools/tool.h"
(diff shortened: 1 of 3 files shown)


artisteacher:
This was on the small size before:
Screen Shot 2022-03-29 at 6 14 35 PM
Though I'm not sure it needs to be quite so big now:
Screen Shot 2022-03-29 at 6 14 08 PM

manongjohn:
On my UHD screen it was extremely tiny without this change. This kind of put it back to where it was on my HD screen. I thought I checked it again on an HD screen. I'll double check it again and compare to current version.

artisteacher:
On a related note, there are a couple other areas that are excessively small on my HD monitor.

- Skeleton tool's animate mode for the active bone - both the text and the position icon are tiny (the text only displays when there are multiple drawings in the level)
Screen Shot 2022-03-29 at 8 31 37 PM

-  The guides (that you pull out from the ruler) are especially faint
Screen Shot 2022-03-29 at 8 35 42 PM

 - The Tracker Tool's letter is also on the small side:
Screen Shot 2022-03-29 at 8 42 44 PM

manongjohn:
Regarding the check indicators, I did compare the results from this PR vs a more recent nightly. On both my HD and UHD screens, they are the same size I originally had them.  I always intended them to be on the larger side so it's very obvious that check is on.

Regarding the other small text you've pointed out, I'll work on updating those. I might open a new PR for that though.

manongjohn:
I'm going to use this PR to fix the small text in viewer only.  I've relabeled this PR to reflect that.

As for the thin/faint lines on high res monitors, I actually need to go through all tools and fix any viewer objects (lines, text, gadgets) that aren't scaling properly.  I will handle that in a separate PR.

artisteacher:
Tracker tool object label & Skeleton tool frame indicator sizes are both improved on my end.

manongjohn:
Thanks for verifying.

PR #942, opened 2022-03-27
Timeline/Xsheet navigation tags

manongjohn:
This PR adds Navigation Tags to the Timeline/Xsheet

image

Command/Quicktool bar buttons:
image
(`Toggle Navigation Tag`, `Previous Tag`, `Next Tag`....If someone feels like making better icons, feel free.)

Navigation Tag features:
- You can tag the currently selected scene frame with the `Toggle Navigation Tag` right-click menu option or command
- Move quickly to a tag using the `Next Tag` and `Previous Tag` right-click menu options or commands
- A `Tags` list is maintained which allows you to see all tag and jump to any of them quickly 
- The default tag label is `Frame #`, but you can `Edit Tag` and add additional message to it
- The default tag color is Magenta, but you can `Edit Tag` and change it
- Hovering over the tag will show the tag's label as a tooltip
- Can click+drag an existing tag to move it
- Tags will shift with `Insert Frame` and `Remove Frame` accordingly

WARNING: Tag information is saved in the scene file.  If you save a scene with tags, the scene will not open in prior versions of T2D or OT that do not support Navigation Tags.  If you need to share with older versions or OT, use the `Remove Tags` options in the `Tags` list.

Suggestions welcome.


F-Burning:
This is very helful, thank you!

Maybe I would do the pins more bigger

image

manongjohn:
The icon has to fit when timeline/xsheet are zoomed all the way out, which doesn't leave much room. The largest size without needing to resize the image is i think 10x10 px.

F-Burning:
Could the icon auto-resize when do that? or is too much complicated? 

Only curiosity.

manongjohn:
Although the image is SVG, it sometimes doesn't look good when it is scaled down. I'd rather have a smaller consistant image in that case. It's a consideration though. 

Other options, like using a different type of marker or marking different areas of the frames area and such has also been considered, but almost every option has its own drawbacks. Just a matter of which is the least detrimental.

F-Burning:
For me this is the best option, the only problem is the size, but if there is no other solution, for me it's fine like this 👍

jeremybullock:
I don't mind the pin icon, but as an alternative, Premiere uses markers which look like this: https://i.ytimg.com/vi/FUU6nt6LwqY/maxresdefault.jpg

jeremybullock:
Nice work on this @manongjohn

manongjohn:
Thanks. 🙂

Those icons look like stickies..which i think are a better fit for the look and feel. Will play with it.

manongjohn:
I just pushed changes to replace the pin with a tag similar to Premier and also renamed everything from `pin` to `tag`.  Added new option to change the tag's color as well.

jeremybullock:
Nice work.

manongjohn:
Thanks for the feedback and suggestions!

artisteacher:
One enhancement for the future might be an option to add a tag "every nth frame".

F-Burning:
When a try to change the color to white in a tag, this goes pink again. The other colors works fine.

By the way, if it's possible, I think the option to create your own color would be great, because full saturated colors hurt my eyes :eyes: 

Thank you again Mr!

manongjohn:
> When a try to change the color to white in a tag, this goes pink again. The other colors works fine.

@F-Burning Thanks for reporting it.  I'll fix it a part of the beta testing cycle.

PR #947, opened 2022-04-02
Implicit Hold fixes 4

manongjohn:
This PR fixes additional Implicit Hold Issues

- Fix implicit hold particle texture rendering

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -600,7 +600,7 @@ int TFx::getReferenceColumnIndex() const {
 
 //--------------------------------------------------
 
-TFxTimeRegion TFx::getTimeRegion() const {
+TFxTimeRegion TFx::getTimeRegion(bool ignoreImplicit) const {
   if (m_imp->m_portTable.empty()) return TFxTimeRegion::createUnlimited();
 
   TFxTimeRegion tr((std::numeric_limits<double>::max)(),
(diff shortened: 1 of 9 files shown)


artisteacher:
Definitely better. I haven't run into the closing issue so far.

 I'm still not seeing the particles texture display in the swatch viewer straight away. I either need to click back on the particles node - even though it's active, or switch to another frame. I believe that's a pre-existing behavior though. 

The viewer's preview itself is still very slow, but does show up eventually. Explicit holds is the same.

manongjohn:
I have noted some odd issues with the fx swatches that I do need to look at separately.

Even in the current 1.2 release, I notice when particle fx use a Raster level for texture, preview rending is slow.  However, if using a Smart Raster for the texture, the preview render is much faster.  Not entirely sure why that is, but the behavior is consistent, even with OT.

artisteacher:
It just might be an issue people run into more freely, now that raster is the default level type. It is strange that raster would be slower - there should be less data to keep track of.

F-Burning:
I still don't quite understand how implicit hold works in this case.

Is there an explanation as a tutorial in some thread? Thank you!

manongjohn:
@F-Burning In terms of textures for particles, if you attach a level to a particle Fx, it will use the drawings from the level that are exposed in the timeline/xsheet.

Particle fx will use implicit frames as if they were explicitly exposed.  There was an issue because there was no real "end" so it would try to use a significant number of frames and take a long time rendering just 1 frame.  I put in some logic to somewhat ignore the implcit frames, though thinking through it while writing this, I think i need to revisit what I put in as it might not render exactly like explicit frames would.

PR #948, opened 2022-04-04
Rename Tahoma2D branding image files

manongjohn:
This PR fixes an issue with the updated branding not being applied to macOS builds.

In order to make the branding files easier to find, I've renamed the ones we use and removed old ones that are no longer needed.


artisteacher:
That fixed it - the icon is updated now. 

I do want to mention that the updated icon makes it more obvious than before that the app icon is oversized in the dock in comparison to all the other app icons:
Screen Shot 2022-04-03 at 11 20 04 PM

artisteacher:
It was oversized before:
Screen Shot 2022-04-04 at 10 11 11 AM
Since it wasn't the same shape, I think it was a little less obvious.

manongjohn:
Ok.  I've fixed the OSX icons.  I redid the `Tahoma2D.icns` file using the SVG version of the logo and creating PNGs of several sizes based on some Apple docs that explained how to create the `.icns` file.  

The docs suggested leaving a border around the icon image. Explains why T2D's icon looked larger since our icon filled the entire size of the image. I shrunk the icons a bit to leave the border per recommendation so now the Tahoma2D icon looks generally the same size as others of similar shape on Macs.

artisteacher:
Perfect - thank you!

PR #949, opened 2022-04-05
Copy missing libopencv dylib needed for plugin support

manongjohn:
This PR only impacts macOS build packaging to correct a missing dylib needed by Tahoma2D plugins.

Currently modifying the Tahoma2D plugin builds for OSX so that it uses the opencv library included with T2D rather than looking for opencv installed on the system.  This would eliminate the need for users to install opencv manually.  

It was discovered not all opencv libraries used by the plugins are included with T2D so I've updated the build scripts to manually include the missing library.

### DIFF
--- a/ci-scripts/osx/tahoma-buildpkg.sh
+++ b/ci-scripts/osx/tahoma-buildpkg.sh
@@ -70,6 +70,9 @@ $QTDIR/bin/macdeployqt $TOONZDIR/Tahoma2D.app -verbose=0 -always-overwrite \
    -executable=$TOONZDIR/Tahoma2D.app/Contents/MacOS/tfarmcontroller \
    -executable=$TOONZDIR/Tahoma2D.app/Contents/MacOS/tfarmserver 
 
+echo ">>> Manually copying missing dylib needed for plugins"
+cp /usr/local/lib/libopencv_dnn.4.5.dylib $TOONZDIR/Tahoma2D.app/Contents/Frameworks
+
 echo ">>> Correcting library paths"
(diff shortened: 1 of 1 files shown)


manongjohn:
Please test with the following plugin bundle: [Tahoma2D_plugins_osx.zip](https://github.com/tahoma2d/tahoma2d/files/8421904/Tahoma2D_plugins_osx.zip)

Once confirmed, I will merge this and update the bundle in the current  `tahoma2d_plugins` repo.

artisteacher:
It's still not working - for some reason I'm still not seeing libopencv_dnn.4.5.ddylib in Frameworks...

Taking a look at the steps from the action build:
Screen Shot 2022-04-05 at 7 38 28 PM

manongjohn:
Seems my local build environment somehow has libopencv_dnn, but the github build environment doesn't.  Not entirely sure where the libopencv_dnn came from but it's a part of my local builds.  I'll need to see how to get it built in github builds.

manongjohn:
I've decided to go the opposite route.  Instead of trying to get libopencv_dnn to build into the github environment, I'm rebuilding opencv in my local environment to exclude libopencv_dnn and will rebuild accordingly.

I've recompiled the plugins without libopencv_dnn.  Please try this on any T2D installation: [Tahoma2D_plugins_osx.zip](https://github.com/tahoma2d/tahoma2d/files/8422745/Tahoma2D_plugins_osx.zip)

I'll be canceling this PR since it won't be needed, but feel free to reply on this if the bundle works.

artisteacher:
It works! Now I need to read up on how to use them...

manongjohn:
Thanks for verifying.  I've just updated the tahoma2d_plugins repo so Mac users should no longer be require to install OpenCV.

PR #951, opened 2022-04-06
Fix Navigation Tag color

manongjohn:
This PR does the following for Navigation Tag colors

- Fixes a bug where selecting `White` would color the tag with magenta instead 
- Added additional colors: Dark Magenta, Dark Red, Dark Green, Dark Blue, Dark Yellow, Dark Cyan, Dark Gray


manongjohn:
> Could the last color used be a sticky setting?

Added

Issue #626, opened 2021-03-23
Xsheet & Timeline resize on Tahoma2D start.


melieconiek:
**Describe the bug**
The xsheet and timeline bins resize when Tahoma2D starts. This only affects the layout that is open on startup.

**To Reproduce**
List steps to reproduce the behavior.
1. Set the xsheet or timeline to a certain size.
2. Restart Tahoma2D.

**Device Information:**
 - OS: Windows
 - Tahoma2D Version: https://github.com/tahoma2d/tahoma2d/commit/0d7afaf98a9c2407425128866304ef15e89ecf41

artisteacher:
I'm not seeing this on macOS.

artisteacher:
Are you using the actions artifact?

manongjohn:
Might be just a Windows thing? I see it with my local builds.

PR #954, opened 2022-04-07
Fix starting panel resizing issues

manongjohn:
This PR fixes a couple of reported panel sizing/display issues at startup.

Due to the logic implemented in #630...
- When restarting after closing T2D with a maximized viewer, the maximized viewer was not restored and the panels in the room did not have any controls.
- When starting a new installation of T2D, some monitor resolution/scale configurations would cause the panels to appear really small and not resize to fit the larger screen.

I reverted the logic that was blocking the panel resizing at startup to fix the issues noted above.

I found and fixed the root cause of #626: The Palette viewer's `Move Palette` widget is only visible when there is a palette. During startup the widget's visibility would toggle from OFF (no palette) to ON (default palette) resulting in panel redistribution with different sets of geometry ultimately causing the timeline to resize.  Forced visibility to ON when the panel is hidden.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -807,10 +807,6 @@ int main(int argc, char *argv[]) {
     w.checkForUpdates();
   DvDirModel::instance()->forceRefresh();
 
-  // Disable the layout temporarily to avoid redistribution of panes that is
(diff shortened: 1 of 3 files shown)


artisteacher:
This works well!

manongjohn:
Thanks for testing!

PR #964, opened 2022-04-12
Improve keyframe icon targeting

### LINKED ISSUE #772: Cursor pointing on wrong place in timeline tab- Mac
Hello, my problem for today is:

the keyframe dragging point of cursor is in different place.
Not sure I'm describing this bug correctly so I give screen recording (sorry for poor quality).

I have problem with NVIDIA Drivers, maybe thats the case, but in other hand all tools like control point editor works fine.
I think it's something with timeline

https://user-images.githubusercontent.com/45039631/131728540-55d29033-50c0-47f9-bb1a-06fa61c81150.mp4

Tahoma2D 1.2
Build: May 12 2021

Mac (27-inch, Late 2012)
Procesor 3,4 GHZ Quad Core Inter Core i7. 32 GB ram 1600 MHZ DDR 3.
Grafika: NVIDIA GeForce GTX 680 MX 2GB
System: macOS Catalina Version 10.15.7

Cheers!

manongjohn:
This PR helps to improve the targeting of keyframe and ease icons on the Timeline/Xsheet as follows

- Fixes #772 by increasing the activation area of the keyframe and ease icons, especially when zoomed out. Primarily impacted Timeline.
- Make ease icons slightly larger
- Enlarges the keyframe or ease icon when within the activation area
keyframeiconactivation
- Fixes an issue with moving an ease icon into the frame next to a keyframe or jumping away
- Shows ease icons when there are at least 3 cells between keyframes. Used to require 4

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -162,15 +162,18 @@ enum class PredefinedDimension {
   SCALE_THRESHOLD        //! scale threshold to simplify the view
 };
 enum class PredefinedPath {
-  DRAG_HANDLE_CORNER,   //! triangle corner at drag sidebar
-  BEGIN_EASE_TRIANGLE,  //! triangle marking beginning of ease range
-  END_EASE_TRIANGLE,    //! triangle marking end of ease range
-  BEGIN_PLAY_RANGE,     //! play range markers
+  DRAG_HANDLE_CORNER,         //! triangle corner at drag sidebar
+  BEGIN_EASE_TRIANGLE,        //! triangle marking beginning of ease range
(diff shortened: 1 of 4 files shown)


DarrenTAnims:
This looks really useful if you're targetting animation keys, but what if you're trying to click the drawing cell behind it. Does this fix mean that'll be harder to do?

manongjohn:
> This looks really useful if you're targetting animation keys, but what if you're trying to click the drawing cell behind it. Does this fix mean that'll be harder to do?

It might be a little harder to do, especially when zoomed out all the way. Trying to find a good balance between the 2 is difficult on the Timeline since there isn't a lot of vertical space when zoomed out.

That's one reason I added the visual indicator for the icons so you know if you'll hit the icon or the cell behind it.

PR #969, opened 2022-04-13
Update to use CrashRpt built in MSVC 2019

manongjohn:
The CrashRpt currently being used was built with MSVC 2017 and doesn't appear to be working correctly since T2D now builds with MSVC 2019.

I've added a version of CrashRpt built using MSVC 2019 to the [tahoma2d\crashrpt2](https://github.com/tahoma2d/crashrpt2) repo.  Modifying the build packing scripts to use it.


manongjohn:
I confirmed this is working again.  Merging now so they start working in the nightlies again.

PR #987, opened 2022-04-25
Draw keyframe instead of using image

manongjohn:
Since we use a diamond for the keyframe markers, there really is no need to use an SVG image of the marker when we can draw it.

The large diamond shown when over a keyframe won't appear as a stretched image and aligns better with the other keyframe indicators.


artisteacher:
This looks so much better!

I do wish that clicking and dragging on a keyframe moved the keyframe instead of starting a selection of keyframes. As it is now, you have to click and release before clicking & dragging. If this changed, we would still be able to select multiple keyframes with the ctrl|cmd modifier.

manongjohn:
> I do wish that clicking and dragging on a keyframe moved the keyframe instead of starting a selection of keyframes

I kind of want to change that too, but would like to discuss that change and see how that would be received and also discuss how to enable that form of selection.

artisteacher:
Maybe with alt?

manongjohn:
> Maybe with alt?

Seems Alt is really the only choice available.  Shift and Ctrl have other behaviors when it comes to keyframe activation.

slacksmatt:
or When user _Click-Hold_ it long enough (e.g 200ms), it will activate the _Click and Drag_

PR #994, opened 2022-05-09
Replace DEL with Backspace for OSX

manongjohn:
This addresses an issue for MacOS users where the `Delete` key on mac keyboard does nothing because it is internally identified as `Backspace` and actions are not typically defined against it.

The following changes were made **for macOS builds only**:

- Replaced the default `Delete` keyboard shortcut `DEL` with `Backspace`
  - This generally impacts wherever a menu shortcut command or action uses the T2D delete command option.

- Anywhere `DEL` is hardcoded for a particular action and there is no `Backspace` action defined, the logic has been switched to work with `Backspace` instead
  - Hex Color Name Editor - Removing entries
  - Motion Path Panel  - Graph widget - Removing curve points
  - CurvesFx - Removing curve point

- Anywhere the `Backspace` is hardcoded for a particular action and there is no `DEL` action defined, the logic will remain the same
  - Script console - removing text
  - Stop Motion Panel - Numpad shortcuts active option switches `Backspace` shortcut to Stop Motion Remove Frame command

- Anywhere a `DEL` action and a `Backspace` action is hardcoded with different actions, the logic will remain the same
  - Type Tool - text edit box

### DIFF
--- a/toonz/sources/toonz/graphwidget.cpp
+++ b/toonz/sources/toonz/graphwidget.cpp
@@ -828,7 +828,11 @@ void GraphWidget::mouseReleaseEvent(QMouseEvent* e) {
 void GraphWidget::keyPressEvent(QKeyEvent* e) {
   if (m_currentControlPointIndex == -1) return;
 
+#ifdef MACOSX
+  if (e->key() == Qt::Key_Backspace) {
+#else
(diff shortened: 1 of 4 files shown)


artisteacher:
This works beautifully! This makes life so much easier on macOS!

Issue #235, opened 2020-09-27
Easier means to set playback in/out points


ladlon:
Rather than having to scroll around the timeline/xsheet to find each in/out marker, and then dragging it (sometimes all the way across the entire duration of the track) to their intended locations, it would be great if there were easier methods to set playback in/out points.

-Set in/out points to fit entire duration of longest column (First and last frame)
-Set in (or out) point to current selected frame/cell
-Set in and out points at start and end of selected frame range
-Move current in/out marker range to current frame/cell (putting in marker on selected frame, and keeping the previous range/distance between the markers before the move)
-And any other useful ones you or other users here may think of

I'm just finding with the current 'grab and drag only' method, there's a lot of timeline scrolling and looking around to find the markers... and then dragging them back to where you want them can be a chore (if the intended location is far offscreen).

artisteacher:
You'll find a couple helpful options for moving the markers in the context menu for the frame numbers (top edge of the timeline or left of the xsheet). You don't necessarily need to drag the markers to move them. Setting the markers based on a selection could be a helpful addition though. 

On a related note, you can also turn on shortplay in preferences (& set a shortplay shortcut) to play a shorter range without needing to change the markers.

ladlon:
You are correct!  Yep, those work for me.

My bad...  I'm new to this, and ideally, I should be giving it more time before I post suggestions, as I'm aware that I might simply not have noticed some things.  I'm just kind of in a crunch to learn this in the next few days, as I have a full broadcast project to do... and made the questionable decision to use the project as an opportunity to properly learn OT/Tahoma!

Thanks again, artisteacher.... and my appologies to the forum staff for me not spotting that before my post.

DarrenTAnims:
It might be worth joining the OpenToonz Discord to discuss if your requirements are already available in the software before you add issues to this github. And you'd get more immediate feedback about them and be able to discuss different approaches. Then log issues only for missing features.

PR #995, opened 2022-05-09
Set Start/Stop Markers enhancement

manongjohn:
This PR adds the following:

- A new option to set the start and stop markers on a selected frame range by selecting a range of cells in the timeline/xsheet, right-click the start/stop marker area and then choosing `Set Markers to Selected Range`.
- A tool tip for setting start/stop markers using CTRL/ALT + click when hovering over frames area


artisteacher:
I think this might resolve #235.

melieconiek:
Might want to make the CTRL/ALT modifiers for setting/resetting the frame range more apparent somehow as well. It's a lot faster than using a dropdown.

manongjohn:
Didn't realize you could use CTRL/ALT for setting markers.  I added a tooltip when hovering over that area for it.

melieconiek:
CTRL + ALT click resets markers as well.

PR #998, opened 2022-05-15
Add Default Projects Folder to Browser

### LINKED ISSUE #996: Add your Default Project Path into the list of root folders
It would make using Tahoma so much easier if my Default Project path (that I setup in the Preferences dialog) was added to the list of root folders in the Load Scene (and other dialogs)?

Root Folders

manongjohn:
This PR fixes #996 by adding an entry for `Projects` in the `File Browser` if a `Default Project Path` is set.  In case of multiple paths (separated by `;`), multiple entries will be created with the base path name appended to `Projects`.

image

image

### DIFF
--- a/toonz/sources/toonz/filebrowsermodel.cpp
+++ b/toonz/sources/toonz/filebrowsermodel.cpp
@@ -126,6 +126,13 @@ void DvDirModelNode::addChild(DvDirModelNode *child) {
 
 //-----------------------------------------------------------------------------
 
+void DvDirModelNode::insertChild(int row, DvDirModelNode *child) {
+  child->setRow(row);
+  m_children.insert(m_children.begin() + row, child);
+}
+
+//-----------------------------------------------------------------------------
+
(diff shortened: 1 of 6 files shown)


DarrenTAnims:
Amazing!!
Thanks MJ

artisteacher:
Works well with macOS!

I did find that if I add a second path later on in a different session, three project directories temporarily display in the Browser, until T2D is restarted:

- Projects
- Projects (1st name)
- Projects (2nd name)

manongjohn:
> I did find that if I add a second path later on in a different session, three project directories temporarily display in the Browser, until T2D is restarted:

I see the problem.  It's replacing the wrong entry when you update root Project folder list.   In Windows, there is an entry in the folders list for `Network` that doesn't exist in macOS/Linux versions so it's replacing whatever normally follows the Projects folder (recent open projects lists).  Because of this, if you click on the wrong Project folder, it will crash T2D.

I'll submit a fix shortly.

PR #1009, opened 2022-06-07
Fix ffmpeg builds on macOS Catalina

manongjohn:
Ffmpeg builds on macOS needed for opencv builds for Tahoma2D are failing because Homebrew upgraded libdav1d from 0.9.2 to 1.0.

Cannot compile a higher version of ffmpeg that allows for the newer version of libdav1d because that would also force a higher version of opencv which I cannot get to compile on Catalina.

To fix this issue and resume with the current versions of ffmpeg/opencv, this PR will build libdav1d 0.9.2 from source instead of using the brew version.

### DIFF
--- a/ci-scripts/osx/tahoma-buildffmpeg.sh
+++ b/ci-scripts/osx/tahoma-buildffmpeg.sh
@@ -25,6 +25,30 @@ sudo make install
 
 cd ../..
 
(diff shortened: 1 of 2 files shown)


manongjohn:
Seems to have fixed it. Going to merge so I can start merging other stuff...

PR #1031, opened 2022-08-21
Fix parent column indicators

manongjohn:
This PR fixes some minor things with the Xsheet's Parent Column dropdown

- Removes Sound and Note columns from the list since you cannot parent columns to these in the Schematic
- Shows parent for Palette columns since it is changable in the Schematic and from the Xsheet.
- Corrects the Palette column color. It was using the default column color (gray-ish) instead of the color assigned to palette columns.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -2903,9 +2903,10 @@ void CellArea::drawPaletteCell(QPainter &p, int row, int col,
                              : m_viewer->getReferenceColumnColor();
     sideColor = m_viewer->getReferenceColumnBorderColor();
   } else {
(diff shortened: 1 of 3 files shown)


adamearle:
Yeah this is really interesting. I have to ask "Shows parent for Palette columns since it is changeable in the Schematic and from the Xsheet"  What is this for? Curiosity kill the cat?

manongjohn:
> Yeah this is really interesting. I have to ask "Shows parent for Palette columns since it is changeable in the Schematic and from the Xsheet" What is this for? Curiosity killed the cat?

No idea. Just know that a Palette column, primarily used for Fx i think, can be a parent to other columns in the Schematic and it can be a child of other columns also. You could always change it's parent on the xsheet, but it never showed on the xsheet, so I just made it to show.

PR #1034, opened 2022-08-25
tahoma2d/tahoma2d#1030 : Change grass preset default

ABelliqueux:
As mentionned here: https://github.com/tahoma2d/tahoma2d/discussions/1030

This resolves the issue by setting the default value to -1 .


JPUser1:
Thanks for the heads up I will add fix to OpenToonz!

manongjohn:
@ABelliqueux Thanks for attempting to fix this, however, after further review of the issue, the root cause potentially still impacts all Particle Fxs, not just Grass.

I have found the bug in the code that causes the issue and will be submitting a fix shortly so will close this.

PR #1058, opened 2022-11-21
Fix pen issues

### LINKED ISSUE #1007: Drawing delay on vector and smart raster levels
### Bug Description

- No problem with drawing with the mouse.
- No problem with drawing with the pen on the raster levels.
- No problem in v1.2.

### Steps to reproduce

1. Open Tahoma2D.
2. Select the brush tool.
3. Create a vector or smart raster level.
4. Start drawing.

### Relevant logs/screenshots

https://user-images.githubusercontent.com/73311467/171819316-597077e5-7350-487b-baa1-a9a9c94e32c4.mp4

### Software Version

1.3

_No response_

### Operating System

Fedora 36 and KDE neon

### Graphics Card

AMD RX 570

### Drawing Tablet

XP-PEN G640

manongjohn:
This PR fixes the following pen issues on Linux

- Fixes #1007 (Linux issue)

When the pen is hovering while using the Brush or Eraser tool on Smart Raster and Vector levels, there was constant `QPaint()` calls while the pen cursor was moving to the new location causing a significant amount of lag.  Oddly, this was not an issue while drawing if you pressed down, waited for the cursor to catch up then started drawing.

This issue appears to be caused by some unknown QT 5.15 bug on Linux systems with calls to `processEvents(QEventLoop::ExcludeUserInputEvents)` that was not an issue when T2D was being built using Qt 5.9.7.

Not sure what QT version this might have started with so I only I blocked execution of this statement for Linux builds using QT 5.15 or later.

- Pen doesn't work in Schematic Viewer. (macOS/Linux issue)

The Schematic Viewer expects pen events to create both Tablet and Mouse events.  Mouse events are actually what drives pen interaction with the schematic viewer.

`QGraphicsView::tabletEvent()` is shadowed in the schematic viewer which forces acceptance of the Tablet event instead of calling the base class function to finish handling it. For Windows this does not prevent the creation of the Mouse event, but it does block it for other builds. To fix, I simply replaced the forced event acceptance with a call to the base class function so the Tablet event is handled normally and creates the Mouse event.

### DIFF
--- a/toonz/sources/tnztools/rastererasertool.cpp
+++ b/toonz/sources/tnztools/rastererasertool.cpp
@@ -1740,7 +1740,9 @@ bool EraserTool::onPropertyChanged(std::string propertyName) {
 //----------------------------------------------------------------------
 
 void EraserTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
+#if (!defined(LINUX) && !defined(FREEBSD)) || QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
   qApp->processEvents(QEventLoop::ExcludeUserInputEvents);
+#endif
 
   struct Locals {
     EraserTool *m_this;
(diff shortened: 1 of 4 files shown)


artisteacher:
That's so much better in the schematic! 

I think the only area where stylus input is still glitchy is the Geometry Tool's arc/multi-arc. They both work with the stylus, however with a mouse you can click & drag to position the top of the arc. With the stylus, you can't click & drag to set the top of the arc. The height gets set immediately with that third click.

PR #1064, opened 2022-11-27
Fix MeshImage compression crash

manongjohn:
This fixes a crash I stumbled upon when I was playing around with the Plastic tool and ran low on available memory.

When T2D starts running low on memory, it will attempt to compress any uncompressed raster images in memory.  T2D crashes when it tries to do this on Mesh levels because it fails to get a raster image that doesn't exist.

Modified logic to skip Mesh Images like it does with Vector Images.

### DIFF
--- a/toonz/sources/common/tcache/timagecache.cpp
+++ b/toonz/sources/common/tcache/timagecache.cpp
@@ -19,6 +19,7 @@
 
 #include "timagecache.h"
 #include "trasterimage.h"
+#include "tmeshimage.h"
 #ifndef TNZCORE_LIGHT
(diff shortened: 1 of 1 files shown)


artisteacher:
Oh, I wonder if that's the cause of some random render crashes. Good catch!

manongjohn:
> Oh, I wonder if that's the cause of some random render crashes. Good catch!

Yeah, been wondering about that myself.

beeheemooth:
Judging by the description, with a high probability this solves problem # 1024. I can confirm this when I have the light. :(

PR #1131, opened 2023-02-04
Symmetry Tool + some tool bug fixes

manongjohn:
This PR adds a new Symmetry Tool as well as fixes a few tool related bugs found while integrating.

image

### **Symmetry Tool** 
--------

The Symmetry Tool provides 1 symmetrical guide which can be moved about the canvas and rotated as needed.  The guide can be configured with the following options:

- `Lines` - Set the number of lines of symmetry (2 - 16) radiating from the center point.
- `Opacity` - Adjust the visibility of the symmetry guides
- `Rotation` - Adjust the rotation of the symmetry guide
- `Color` - Set the color of the symmetry guide
- `Line Symmetry` - Use guide lines as mirror lines. This is only available with an even number of lines.
- `Presets` - Allows you to save and reuse guide settings
- `Reset Position` - Re-centers the guide to the middle of the canvas

Additional notes:
- Settings are remembered between sessions.
- Moving the center doesn't move the rotation handle. Hold `Shift` while moving center to shift both
- `Alt` snaps rotation to 15° increments

In order to show/enable the Symmetry Guide in the Viewer, you must enable the `Grid and Overlays` option of the Viewer and check the `Symmetry Guide` option:

The following tools have been configured to work with the Symmetry guide:
- Brush Tool
- Fill tool
- Paint tool
- Eraser tool (Does not work with `Invert` option)
- Tape tool
- Geometry Tool
- Selection Tool

Symmetry guide does work with `Frame Range` and `Onion` tool options as well as Guided Tweening with the Vector Brush and Geometric tools.  Results may not be consistant on all symmetrical regions due to various factors.

Known issues while using Symmetry Guide:
- Brush tool on standard Raster layers has some drawing lag when not using a MyPaint brush style. The lag becomes worse with more lines of symmetry and high `Hardness` setting. Not entirely sure why yet considering symmetry works fine on other levels. 
- Tape tool on Vector layers can take a lot of time to work if using `Rectangle` mode with high `Distance` setting and many lines of symmetry.

### **Tool Bug fixes**
--------
The following tool bugs were fixed
- Fill Tool
  - Fixed bug with changing Onion-fill setting not taking effect until you change another setting
  - Fixed bug where sometimes switching to Normal Onion-fill mode doesn't work until you switch frames
  - Fixed crash caused when Freehand/Polyline Onion-fill mode stroke is completely outside of Smart Raster canvas
- Eraser Tool (Vector)
  - Fixed bug with vector frame range undo incorrectly applying all undo actions to the wrong frames. Sometimes resulted in crashes.
- Tape Tool (Smart Raster)
  - Fixed bug with smart raster frame range undo incorrectly applying all undo actions to the last frame.


Dixxi0509:
This is amazing can't wait!

PR #1146, opened 2023-03-21
Fix autodetect 3rd party

manongjohn:
This PR fixes T2D's autodetection of 3rd party software (ffmpeg/rhubarb) that I inadvertently broke for Linux builds when I ported over an OT patch related to autodetection.  Not sure if macs were impacted also but this should fix them if it was. Windows was ok.


artisteacher:
Yes, macs needed the fix too.

artisteacher:
This works perfectly on macOS. Thanks for the fix!

PR #1150, opened 2023-03-30
Fix mixing multiple audio + rendering 32bit float sound

manongjohn:
This PR fixes the following: 
- Updated ffmpeg sound rendering arguments to account for 32bit float to fix garbled/noisy rending.
- Fixed issue with multiple audio not always mixing


artisteacher:
So much better!

PR #1060, opened 2022-11-25
Fix Implicit Hold issues

manongjohn:
This fixes the following Implicit Hold issues

1. Fixes #1011 
2. Fixes #1023
3. Fixes #1041
4. Fixes a reported issue of New Level creation with Step > 1 creating explicit cells
5. Fixes a Subscene issue mentioned in #1055
       - In this case, the initial frames exposed when Collapsing levels into a Subscene is based on the last keyframe or the  last stop frame, whichever is first. If neither are present, then it's at the last drawing..

PR #1167, opened 2023-05-31
Fix more implicit issues

### LINKED ISSUE #1055: Implicit Hold Issues
### Bug Description

I'm loving implicit holds but have noticed a couple issues:

1. Playback markers & output range don't update automatically. Not sure if this part is a bug or a deliberate decision. Either way, I think it would be more helpful for actions like reframing & adding keyframes to automatically adjust the playback range & output range - unless they've been manually set.
2. Sub-scenes ignore implicit hold frames and only generate the correct number of frames if a drawing exists on the last frame

### Steps to reproduce

For 1:
1. Create a level & draw on a couple frames
2. Use the reframe command set to 2's (or etc.) -> playback markers & output range don't update

or
1. Create a level with a single drawing
2. Add keyframes with the animate tool on several different frames (ie. 1, 10, 50 or etc.) -> playback markers & output range don't update 

For 2:
1. Collapse a couple layers with single drawings & multiple keyframes into a sub-scene 

### Relevant logs/screenshots

Inside a sub-scene:
Screen Shot 2022-11-01 at 10 24 57 PM
The result of collapsing the layers:
Screen Shot 2022-11-01 at 10 25 22 PM

### Software Version

Nightly (Specify below)

June 21st 2022

### Operating System

macOS Monterey

### Graphics Card

_No response_

### Drawing Tablet

_No response_

### LINKED ISSUE #1143: Crash when reframing.
### Bug Description

This crash happens in the Xsheet and in the timeline.

### Steps to reproduce

1. Create a new level (of any type) and move the frame a few cells lower (you can also create a level directly in the cell you need).
2. Highlight the frames so that the cells above the level you created are also highlighted.
3. Apply any "Reframe" option.
4. Crash.

ezgif-5-f0639abde7

### Software Version

Nightly (Specify below)

Built: Feb 28 2023

### Operating System

windows 8.1

### Graphics Card

NVIDIA GeForce GT 1030

### Drawing Tablet

Huion KAMVAS PRO 20 (2019)

### LINKED ISSUE #1163: Reframe when Camera Track Selected Causes Crash
### Bug Description

Not technically a bug since (obviously) Camera tracks can't be (or aren't supposed to be) retimed, but still...

If the user has a camera track selected and hits one of the Reframe buttons, this results in a catastrophic crash and the loss of all work.

A warning dialog would be nice... as well as—needless to say—not allowing the crash.

### Steps to reproduce

To reproduce:

- create a series of frames,
- select them all as well as any frame on the Camera track (or just the frames on the Camera track),
- click on one of the reframe buttons in the timeline toolbar.

### Relevant logs/screenshots

[CrashRpt-Log-20230525-075505-{ae8b21f2-91d2-4abb-a069-6543ab995a48}.txt](https://github.com/tahoma2d/tahoma2d/files/11564468/CrashRpt-Log-20230525-075505-.ae8b21f2-91d2-4abb-a069-6543ab995a48.txt)

### Software Version

1.3

v 1.3.1 (Dec 8, 2022)

### Operating System

Windows 10 21H2 19044.2965

### Graphics Card

GTX 980M

### Drawing Tablet

none

manongjohn:
This PR fixes the following Implicit Hold Issues

- Fixes calculating an Implicit Scene frame count
   - The change made in #1060 to calculate the number of frames to show when collapsing subscene has been extended to general scene (timeline/xsheet) frame calculations
   - This change fixes an issue with Preview-Render not rendering past the last real level frame
   - For implicit scenes only, this change moves the auto-stop marker based to last key or last stop frame, whichever is first or last drawing if neither are present. (fixes #1055)
- Fixes Preview-Render ignoring the last Stop Hold
- Fixes implicit to explicit scene conversion to factor in Stop marker when determining where to end a level exposure without a stop hold. 
- Fixes Rollup/Rolldown issue incorrectly creating explicit cells
- Fixes reframe crashes caused by logic checking for stop hold frames (fixes #1143 and fixes #1163 )

### DIFF
--- a/toonz/sources/include/toonz/txsheet.h
+++ b/toonz/sources/include/toonz/txsheet.h
@@ -593,7 +593,7 @@ in TXsheetImp.
   bool isReferenceManagementIgnored(TDoubleParam *);
 
   void convertToImplicitHolds();
-  void convertToExplicitHolds();
+  void convertToExplicitHolds(int endPlayRange);
 
   NavigationTags *getNavigationTags() const { return m_navigationTags; }
   bool isFrameTagged(int frame) const;
--- a/toonz/sources/toonz/previewfxmanager.cpp
+++ b/toonz/sources/toonz/previewfxmanager.cpp
@@ -558,6 +558,9 @@ void PreviewFxInstance::updateFrameRange() {
   properties->getRange(m_start, m_end, m_step);
   if (m_end < 0) m_end = frameCount - 1;
 
+  if (m_end >= frameCount && Preferences::instance()->isImplicitHoldEnabled())
+    frameCount = m_end;
+
   // Intersect with the fx active frame range
   TRasterFxP rasterFx(m_fx);
   TFxTimeRegion timeRegion(rasterFx->getTimeRegion());
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -1283,39 +1283,6 @@ void removeFx(TXsheet *xsh, TFx *fx) {
 
 //-----------------------------------------------------------------------------
 
-int getChildFrameCount(TXsheet *childXsh) {
-  if (!childXsh) return 0;
-
-  int frameCount = childXsh->getFrameCount();
-
-  // For implicit holds, use last Stop Frame marker or last Key frame marker as
-  // frame count
-  if (Preferences::instance()->isImplicitHoldEnabled()) {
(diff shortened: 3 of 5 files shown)


artisteacher:
This seems to be working great so far! 

I haven't used the roll up/down commands before. Should rolling up change the step like this:
before
->
after

manongjohn:
> Should rolling up change the step like this:

Yes that is the expected behavior.  The 1st cell is moved to the end of the selection and all the other cells in the selection shift up one cell.

PR #1174, opened 2023-06-14
Restore current column indicator

manongjohn:
This PR restores the Current Column indicator. The default color has been changed to cyan to match the current cell indicator which can be changed in `Scene` preference.


manongjohn:
Looking for feedback....Based on @beeheemooth's suggestion, how's highlighting just the name for the Current Column Indicator instead of text highlight work for you?

image

artisteacher:
I think that's just fine. Definitely makes it clear!

PR #1179, opened 2023-06-25
Fix plastic render crash

### LINKED ISSUE #889: Render crashes with some mesh deformed animations
**Describe the bug**

Several of my students have been having crashes when rendering scenes that utilize the plastic tool. The version of T2D and the version of FFMPEG (ie. latest release vs. T2D's custom) doesn't seem to make a difference. Render format hasn't made a difference. Task rendering, especially to an image sequence is more likely to be successful but it's not a guarantee.

I have a couple scenes I can share privately that always crash.

**To Reproduce**
1. Make an animation with the plastic tool - longer/more complex animations seem more likely to crash (level type does not seem to make a difference)
2. Fast render to MP4 (other render formats and Task rendering can crash too)

[plastic-crash.txt](https://github.com/tahoma2d/tahoma2d/files/7955107/plastic-crash.txt)

**Device Information:**
 - OS: macOS Catalina & Monterey
 - Tahoma2D Version: Nightly builds from early 9/21 and from 1/20/22

manongjohn:
I believe this will fix crashes some may be having when rendering scenes with meshes. (fixes #889)

I tested this on a 700+ frame scene with all vector levels and a few meshes.  It would constantly crash during render after 92 frames with the following error:

```
ARB::createContext: wglCreateContextAttribsARB() failed (GL error code: 0x502) for format: QSurfaceFormat(version 2.0, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize -1, redBufferSize -1, greenBufferSize -1, blueBufferSize -1, alphaBufferSize -1, stencilBufferSize -1, samples -1, swapBehavior QSurfaceFormat::SwapBehavior(DefaultSwapBehavior), swapInterval 1, profile  QSurfaceFormat::OpenGLContextProfile(NoProfile)), shared context: 0x100cf (The operation completed successfully.)
GDI::createContext: wglCreateContext failed. (Unknown error 0xc00705aa.)
Unable to create a GL Context.
```

It appears that the plasticdeformerfx would constantly create a new context without releasing it when it was done, expecting it to happen later, but never does.  I believe it would eventually run out of memory (or something) to build more contexts and subsequently crashes when trying to use the context.   Where it would crash during render my vary depending on a user's system.

At any rate, I added logic to immediately release and destroy the newly created context as soon as it was done with it.  This allowed my scene to render without issue. 

EDIT:  I think this might also fix render issues with plastic where it appears to get stuck.  Sometimes my scene would crash, other times it would get stuck on a frame and never continue.
 
If anyone has scene with plastic meshes that kept failing to render, I would ask you to try with this PR and let me know the results.

### DIFF
--- a/toonz/sources/toonzlib/plasticdeformerfx.cpp
+++ b/toonz/sources/toonzlib/plasticdeformerfx.cpp
@@ -444,8 +444,10 @@ void PlasticDeformerFx::doCompute(TTile &tile, double frame,
 
     // ts->unloadTexture(texId);                                // Auto-released
     // due to display list destruction
-    context->deleteLater();
-    // context->doneCurrent();
+    //context->deleteLater();
+    context->moveToThread(0);
+    context->doneCurrent();
+    delete context;
   }
   assert(glGetError() == GL_NO_ERROR);
 }


artisteacher:
I don't have a scene that consistently crashes to test with, but this seems to work fine so far.

PR #1182, opened 2023-06-29
Fix studio palette drawtoggle for linux

konero:
Fixes https://github.com/opentoonz/opentoonz/issues/4949


manongjohn:
Thanks for porting over this fix!

PR #1183, opened 2023-07-03
refactor: Icon stuff (Tahoma2D)

konero:
This is a port of https://github.com/opentoonz/opentoonz/pull/4988 just to save you the trouble if you wanted to port it. I had to port it line by line as there was a lot of divergence between the bases and had to handle Tahoma stuff slightly different.

Basically, the way icon themes are handled currently is pretty bad, now almost everything happens in one place. Instead of using QIcon::theme we're just storing icon image paths in a QMap and retrieving them with the icon file base name (without extension). You can make icons with createQIcon, or generateIconPixmap/generateIconImage, all are automatically themed.

I also fixed the footermemo icon SVG here specific to Tahoma, which had a huge size and caused out of memory errors when trying to handle it with the svgRenderer at default size.

Other than that, there's a bit of performance increase, images were recomputed hundreds of times when the mouse was moved from TPanelTitleBar because it was dumped into the paint event (my bad), now they only compute once, but will recompute when the stylesheet changes to update the highlight color.

I also changed the mimetype icons.

image


manongjohn:
@konero I was definitely going to port it over as it is a very good change to make. Anything that improves performance is always a great thing to have.

Just wasnt sure when i would get to this, either before or after 1.4, but since you've done it for me, i think that answers the question. 😁

Thank you for the recent fixes! I really appreciate the help.

Curious, is there a way in the future to leverage this to automatically switch icon colors when switching themes without having to restart?

konero:
@manongjohn No worries!

> Curious, is there a way in the future to leverage this to automatically switch icon colors when switching themes without having to restart?

Indeed there should be, the main reason for the ThemeManager class is to hopefully centralize it a bit, I think that may be beyond my skill level though because I think I tried everything, there's no global way to update the GUI it seems we need to handle each widget individually, but because the Toonz code base is super massive the logic on a lot of stuff goes right over my head, so it's difficult to refactor. 🤣 

The closest compromise I found is the observer pattern, so we can make stuff a subclass of some observer class which would store all the identifier stuff, but each class would need some method I think to rerun however it set its icon (if that makes sense).

I have a wonky patch that goes on-top of the OT branch this is based on here that is a working hacky demonstration: https://github.com/opentoonz/opentoonz/commit/b168cc5c270f585760089cf70ef707e713c1468c

Only applied to the tool bar, it's actually not a lot of code, you could no doubt do a better job than that though.😂 

runtimeupdate

Lots of little issues might pop up, for example in this example although the tool bar icons update, the tool icons in the custom panels (if they're huge buttons with large icons) will shrink to tool bar size.

manongjohn:
@konero Reviewing the changes..thanks again for applying the changes you made to OT to T2D.  It would have taken me a bit of time to port all these changes.

I'm still reviewing the changes but tested the artifact quickly and I noticed that menu icons are showing despite the default advanced Interface option `Show Icons in Menu` option being turned off.  This option hides ALL menu icons, but with this PR it only appears to turn off some of them.

manongjohn:
Also found an issue in the Load Scene dialog:
image

Something odd with the `My Documents` and `Desktop` icon.  I recall fixing this once before, though I don't know if it was with these icons in particular. Will try and identify what might have been inadvertantly dropped to re-cause this.

toonz/sources/toonz/filebrowsermodel.cpp
```diff
@@ -1165,22 +1152,19 @@ void DvDirModelRootNode::refreshChildren() {
...
-    child->setPixmap(recolorPixmap(svgToPixmap(
-        getIconThemePath("actions/16/my_documents.svg"), QSize(16, 16))));
+    child->setPixmap(generateIconPixmap("my_documents"));
     m_specialNodes.push_back(child);
     addChild(child);
 
     child =
         new DvDirModelSpecialFileFolderNode(this, L"Desktop", getDesktopPath());
-    child->setPixmap(recolorPixmap(svgToPixmap(
-        getIconThemePath("actions/16/desktop.svg"), QSize(16, 16))));
```
manongjohn, line 1176: @konero To fix the oversized icons that were showing up in the File browser dialog, I had to introduce the QSize(16,16) on a few of the icons.  See commit https://github.com/tahoma2d/tahoma2d/commit/cea78e587a7e420485e8b6e2f75bd6fd8780d126

Removing this is causing these icons to appear large again.  Not sure if the image itself needs a change or not.

toonz/sources/toonz/mainwindow.h
```diff
@@ -205,22 +205,23 @@ class MainWindow final : public QMainWindow {
...
+                                           QString newStatusTip    = "");
 
   QAction *createMiscAction(const char *id, const char *name,
                             const char *defaultShortcut,
                             QString newStatusTip = "");
   QAction *createToolOptionsAction(const char *id, const char *name,
                                    const QString &defaultShortcut,
-                                   QString newStatusTip = "");
+                                   QString newStatusTip = "",
+                                   const char *iconSVGName = "");
```
manongjohn, line 220: Note to self: I will submit a followup PR to reverse the order of these 2 parameters and update the impacted calls so the order matches all the other create action definitions.

toonz/sources/toonzqt/gutil.cpp
```diff
@@ -185,343 +187,338 @@ QPixmap scalePixmapKeepingAspectRatio(QPixmap pixmap, QSize size,
...
+  QImage overImg(generateIconImage(iconSVGName + "_over", qreal(1.0), newSize));
+  QImage onImg(generateIconImage(iconSVGName + "_on", qreal(1.0), newSize));
+
+  QIcon icon;
+
+  // START_BUG_WORKAROUND: #20230627
+  // Set an empty pixmap for menu icons when hiding icons from menus is true,
+  // search bug ID for more info.
+#ifdef _WIN32
+  bool showIconInMenu = Preferences::instance()->getBoolValue(showIconsInMenu);
```
manongjohn, line 498: This needs to be 
```
  bool showIconInMenu = Preferences::instance()->isShowAdvancedOptionsEnabled() &&
                        Preferences::instance()->getBoolValue(showIconsInMenu);
```

manongjohn:
The memo icon seems to be compressed in both Xsheet and Timeline:

Before: 
image

After:
image

manongjohn:
Something causing a shift in the Motion Panel:

Before:
image

After:
image

manongjohn:
Can't close a Flipbook.  Trying to use the close button causes it to minimize.  Hovering over the Close button highlights the Minimize button instead of turning the Close button red like other panels.

konero:
Ah, thanks for checking, these are some odd issues. Weirdly I don't have a single issue in my Qt5.15.2 environment, but I've built with Qt5.9.9 and am also experiencing them now, I'll see if I can push some fixes.

manongjohn:
>  I've built with Qt5.9.9 and am also experiencing them now, I'll see if I can push some fixes.

Between 5.9 and 5.10 how icons were handled in Qt changed.  Due to other issues I've encountered trying to upgrade, I haven't updated Windows builds to 5.15.  MacOS and I believe Linux are on 5.15

What I found odd about these issues in the past was that even though the code setup was identical, the issue wasn't consistant between some icons.  I thought it was something different between 2 similar SVG images but couldn't find anything obvious.

konero:
I did some debug on createQIcon and for some reason a lot of the icons are being created at `QSize(100, 100)`, which is I believe default size of the QSvgRenderer, so you were right it's due to differences within the SVGs themselves.

If an icon has this line: `svg width="100%" height="100%" viewBox="0 0 16 16"` then Qt5.9 doesn't understand it and falls back to the default size, but it does understand `svg width="16px" height="16px"`, making the change fixes a lot of the issues. It seems Qt5.10 updated QSvgRenderer, which is why I didn't get the issue on Qt5.15.

I can update the SVGs, that might mean I'll need to do it on OpenToonz to keep in sync unless Tahoma2D on Windows could be updated to Qt5.10 at a minimum. It's a simple change to make for the SVGs though.

So far, I haven't been able to reproduce the FlipBook close button issue.

manongjohn:
When I tried upgrading to 5.10, it caused a number of issues in T2D that I couldn't figure out easily.  When i tried to go to 5.15, it caused drawing lag for a number of people and I don't know which version it would have started in.

I don't know that you need to fix it for OT since they won't be building on 5.9 and it's pretty rare that these SVG change.

I can't stay on 5.9 forever. I do plan on moving to a more recent QT version as some point when I have time to figure out why T2D has issues that OT doesn't.

toonz/sources/toonz/filebrowsermodel.cpp
konero: Thanks, I resolve this in latest commit differently

konero:
> When I tried upgrading to 5.10, it caused a number of issues in T2D that I couldn't figure out easily. When i tried to go to 5.15, it caused drawing lag for a number of people and I don't know which version it would have started in.

@manongjohn Ah that seems like a pain, I see, no worries we can workaround it, check the latest commit and see if that fixes the issues you noticed, I calculate the SVG image size with a special function which makes absolutely sure we're getting the size from the viewBox, if there is no viewBox it will return a null QSize and for that we'll fall back to the default svgRenderer size. On my end, it seems to fix the following issues:
- Incorrectly sized command bar icons
- File browser ginormous icons for desktop/my documents
- Footermemo being squished
- Hiding icons in menus

I still haven't been able to reproduce the FlipBook close button issue or the Motion Paths panel gap issue, can you check again to see if latest commit still produces them on your end?

konero:
OK I think that's all the issues fixed.🤞

manongjohn:
@konero Looks good!  All my reported issues have been resolved.

Thought you were just going to modify the SVG files. The fix for the size issue will help with future issues.  Many thanks!

I'll give a an additional day or 2 before I merge to give time for any linux/macOS users to review.

PR #1184, opened 2023-07-03
Range intermediate status bar and tooltip

TomDoingArt:
I have made a change to include a description for the Brush Tool Range option "Intermediate Stroke in Range" on the status bar.

image

The mouseover tooltip will show this:
image


manongjohn:
LGTM.  Thanks for the update!

PR #1198, opened 2023-08-12
Fix macOS lib path issue

manongjohn:
This fixes the issue mentioned in https://github.com/tahoma2d/tahoma2d/issues/1142#issuecomment-1675111219

The dependent library paths in `libgnutls.30.dlyb` (from the homebrew of `gnutls`) changed to look for them outside the current `@loader_path`.  This would cause T2D to crash on startup when it couldn't find it.

Modified the package script to detect this scenario and correct the paths accordingly.

### DIFF
--- a/ci-scripts/osx/tahoma-buildpkg.sh
+++ b/ci-scripts/osx/tahoma-buildpkg.sh
@@ -81,33 +81,49 @@ $QTDIR/bin/macdeployqt $TOONZDIR/Tahoma2D.app -verbose=0 -always-overwrite \
    -executable=$TOONZDIR/Tahoma2D.app/Contents/MacOS/tfarmserver 
 
 echo ">>> Correcting library paths"
-for X in `find $TOONZDIR/Tahoma2D.app/Contents -type f '(' -name *.dylib -o -name *.so ')' -exec otool -l {} \; | grep -e "^toonz" -e"name \/usr\/local" -e"@rpath" | sed -e"s/://" -e"s/ (.*$//" -e"s/^ *name //"`
-do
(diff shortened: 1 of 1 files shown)


manongjohn:
@artisteacher Please let me know if this fixes the startup crash. Will merge once confirmed.

artisteacher:
I'm still getting the same crash on open with this action build.

manongjohn:
Seems it didnt see the lib it needed to fix. Enabled debug mode to see if i can tell when the lib is copied over.

manongjohn:
@artisteacher  Was the recent crash message exactly the same or is it different?

artisteacher:
Looks the same.

manongjohn:
@artisteacher When you get a chance, can you check this again?

artisteacher:
There's still a crash, but for a different library this time.
Screenshot 2023-08-13 at 10 33 13 AM

manongjohn:
Thanks. Seems like the lib i copy over needs to be checked and fixed as well. Need to rewrite how the script looks for missing libs and work recursively.

manongjohn:
@artisteacher Please check again when you have a chance.

artisteacher:
And another one...
Screenshot 2023-08-14 at 3 42 53 PM

artisteacher:
And another:
Screenshot 2023-08-14 at 7 38 09 PM

manongjohn:
Sorry for all the passes.  I think I finally got the logic right to fix all the library paths.

Let me know.

artisteacher:
It opens! Thank you!

Issue #1107, opened 2022-12-13
Crash when moving or deleting project that is selected in the Browser file tree


artisteacher:
### Bug Description

If a project is actively selected (though not necessarily active/loaded) in the left side of the Browser, deleting or moving a directory outside of T2D causes a crash.

Deleting/moving files/directories outside of a project folder don't cause any crashes.

### Steps to reproduce

1. In the Browser room, select a project directory from the left file tree
2. Outside of T2D (with Finder or etc.), delete or move the project directory = crash

### Relevant logs/screenshots

[project-browser.txt](https://github.com/tahoma2d/tahoma2d/files/10213553/project-browser.txt)

### Software Version

1.3

### Operating System

macOS Monterey

PR #1201, opened 2023-08-16
Fix macOS backtrace error

manongjohn:
This fixes an issue with macOS backtrace reporting on crash.  The backtrace was showing `sh: addr2line: command not found` error.  It should be using `atos` .

### DIFF
--- a/ci-scripts/osx/tahoma-build.sh
+++ b/ci-scripts/osx/tahoma-build.sh
@@ -29,6 +29,7 @@ fi
 
 export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/opt/jpeg-turbo/lib/pkgconfig"
 cmake ../sources  $CANON_FLAG \
+      -DCMAKE_BUILD_TYPE=RelWithDebInfo \
       -DWITH_GPHOTO2=ON \
       -DQT_PATH=$USEQTLIB \
       -DTIFF_INCLUDE_DIR=../../thirdparty/tiff-4.2.0/libtiff/ \
--- a/ci-scripts/osx/tahoma-buildpkg.sh
+++ b/ci-scripts/osx/tahoma-buildpkg.sh
@@ -71,7 +71,7 @@ find $TOONZDIR/Tahoma2D.app/Contents/Frameworks/libgphoto2* -name *.la -exec rm 
 
 echo ">>> Configuring Tahoma2D.app for deployment"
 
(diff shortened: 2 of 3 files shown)


manongjohn:
Looking for macOS users who can test the artifact by causing a crash, somehow, then post the log so I can review.  Thanks.

artisteacher:
I'll give it a try!

artisteacher:
There's only a windows action build...

manongjohn:
Can you double check?  I do see the macOS action build.

artisteacher:
You are right - I must have clicked on the wrong one before.

artisteacher:
I followed the steps of #1107 to force the crash.
[backtrace.txt](https://github.com/tahoma2d/tahoma2d/files/12443231/backtrace.txt)
It does show the actual username, rather than USER like the standard crash reports, if that matters.

manongjohn:
Hmm.  My local build on Catalina shows filename and line number on some lines (T2D libs) when i reproduce that crash.  I wonder why it doesn't show on yours.

I just pushed a change to obfuscate usernames in logs.  Affects all builds, not just macOS.

artisteacher:
Weird. I wonder if it is one of those differences that happen on local builds, or if it has to do with a change in the OS or a difference in the libraries through Github? Here's the updated result.
[updated-backtrace.txt](https://github.com/tahoma2d/tahoma2d/files/12446766/updated-backtrace.txt)

I do recommend removing the ability to close the crash dialog without closing the application for macOS. If I close the dialog, I either get another automatic crash or non-responsive hanging or both. It's better to close the application while it's possible, rather than having to force-quit anyway - especially since I don't think continuing to run will ever be really possible.

manongjohn:
@artisteacher Can you create the crash log using this local build (Tahoma2D-osx.dmg): https://1drv.ms/u/s!Au-akGfpWFmGi_dpR10uO7TqotniYQ?e=o2WQ6E ?  I want to see if this one will show file/line numbers like it does when I run it on the machine it was built on.

artisteacher:
Here you go:
[new-backtrace.txt](https://github.com/tahoma2d/tahoma2d/files/12458774/new-backtrace.txt)

manongjohn:
@artisteacher Seems, when the app is run on a different machine than it was compiled on, it won't show the filename/line numbers.  Well, at least with the way I currently have it configured to do it.  I believe I have to create DSYM files for it to work on other machines.

I think I'll merge it as is and continue playing around with it.

Thanks for testing.

artisteacher:
I'm always happy to test!

PR #1208, opened 2023-08-30
Fix macos library path reference

manongjohn:
This PR fixes the issue of the macOS builds not fully starting, not even a splash screen, and just stops responding.

This is, in part, due to yet another set of missing frameworks and bad library path references caused by recent updates in Homebrew to webp and qt@5 packages.  Coupled with the crash handler not able to handle the crash at immediate startup causes it to get stuck.

This fix will only address the missing frameworks and library path issue.  Will needed to seprately handle why the crash handler is causing it to get stuck instead of crashing at start up 

Will merge this once I verify the artifact successfully opens.

### DIFF
--- a/ci-scripts/osx/tahoma-buildpkg.sh
+++ b/ci-scripts/osx/tahoma-buildpkg.sh
@@ -93,11 +93,15 @@ function checkLibFile() {
          if [ ! -f $TOONZDIR/Tahoma2D.app/Contents/Frameworks/$Y ]
          then
             local SRC=$DEPFILE	
-            local Z=`echo $DEPFILE | cut -c 1-24`
-            if [ "$Z" = "@loader_path/../../../.." ]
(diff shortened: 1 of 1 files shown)


manongjohn:
Seems beyond some lib path changes, something else has changed in homebrew that is causing further problems with macOS builds.

I noticed they updated QT 5.15.10 to 5.15.10.1 and I wonder whatever change was made is now causing some problems.

manongjohn:
Think I finally got it this time.  Homebrew updated qt@5 which somehow impacted how `macdeployqt` worked.  It would no longer deploying all the Qt frameworks it used to before as well as the dylib load library path changed.  Fixed it by manually copying the missed Qt frameworks as well as created a symbolic link to satisfy the library path changes.

manongjohn:
Confirmed build is working again. *...and there was much rejoicing...yaaaaay...*

PR #1210, opened 2023-09-02
Symbolicate macOS and Linux crash logs for debugging

manongjohn:
In order to provide filename/line number in backtrace information for macOS and Linux crash logs, this PR does the following:

1. This generates DSYM files for macOS and configures T2D to use them when generating backtrace.

As a result of including the DSYM files as part of the build, the overall footprint will increase by about 570 MB.

2. Compiles Linux builds to include symbol information.

As a result of doing this, the overall footprint will increase by about 200 MB.

### DIFF
--- a/ci-scripts/linux/tahoma-build.sh
+++ b/ci-scripts/linux/tahoma-build.sh
@@ -15,6 +15,7 @@ source /opt/qt515/bin/qt515-env.sh
 
 export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 cmake ../sources \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DWITH_GPHOTO2:BOOL=ON \
     -DWITH_SYSTEM_SUPERLU=ON
 
--- a/ci-scripts/linux/tahoma-buildpkg.sh
+++ b/ci-scripts/linux/tahoma-buildpkg.sh
@@ -78,7 +78,7 @@ then
 fi
 
 export LD_LIBRARY_PATH=appdir/usr/lib/tahoma2d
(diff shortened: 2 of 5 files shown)


manongjohn:
@artisteacher and/or other macOS users: Can you test this on something that crashes and post the results?

I know when I download the artifact and run on a scene that crashes for me, it shows file/line numbers where it can.  Prior artifacts didn't so I'm hopeful this will work and be helpful in the long run.

artisteacher:
Here's the result:
[new-crash-report.txt](https://github.com/tahoma2d/tahoma2d/files/12581496/new-crash-report.txt)
I think it's doing what you want?

manongjohn:
Yes it is. Thanks!

manongjohn:
Revised logic and also updated Linux builds to be able to show filenames and line numbers.  PR title and information updated to reflect the change.

artisteacher:
Here's the latest crash report, in case you need it:
[latest-crash-version.txt](https://github.com/tahoma2d/tahoma2d/files/12602378/latest-crash-version.txt)
FYI, I'm using the steps of #1107 to trigger the crashing.

I do recommend removing the ability to close the crash report window on macOS in order to "resume the program". I've never been able to get back to T2D, it's just an immediate new crash after the dialog. It would be better for closing the window to really close the app.

manongjohn:
Thanks for retesting.  It's working as intended.  I will verify the Linux version is doing what I want it to do also.

Will look at disabling the Resume functionality in a separate PR.

manongjohn:
I checked the Linux versions on a different VM and they work as expected.  Interestingly the gcc version was more in line with the Windows and macOS backtraces.  The clang version was either a line # off or didn't have a line number.   At this point, I'll take what I can get. 

Going to merge this so any macOS/Linux users using the latest nightly can hopefully give useful crash reports.

PR #1212, opened 2023-09-13
Fix SVG sequence file loading

manongjohn:
This PR fixes an issue when trying to load an SVG file sequence.

Modified it so SVG files are treated as Vector **Images** instead of Vector **Levels**, which allows the SVG files to be detected as a single image file or a series of files (filename.####.svg or filename_####.svg format).


manongjohn:
Discovered the PLI files are needed for palettes. Updated to restore the logic to create the PLI files when importing SVG sequenced files.

PR #1214, opened 2023-09-15
Disable resume after crash

manongjohn:
This will disable the ability to resume usage of T2D after encountering a crash by disabling the close window button and ignoring the request to close the crash dialog box by hitting the `ESC` key.


artisteacher:
Thanks for adding this.

It works - though the close window button doesn't display the disabled state like it probably should.

manongjohn:
Not sure why it doesn't show in a disabled state. Regardless, I found a way to hide the close button instead of just disabling it.

PR #1216, opened 2023-09-19
Control file history information

manongjohn:
File history is written in scene files (tnz), in vector files (pli) and, under certain circumstances, in smart raster history files (hst).  It  records the last time the file was saved along with machine and user information.  You can view this information by opening the file directly or by right-clicking a scene or level and selecting `Info...` in the Browser.

At a user's request, I 've added the following to control file history in `Preferences` -> `Saving`:

image

This allows the user to:
- Disable adding new entries to file history.
   - This only affect writing new entries.  Existing entries are kept.
- Specify an alternate username for file history instead of the default username

Changes to these settings requires a restart.

### DIFF
--- a/toonz/sources/common/tcontenthistory.cpp
+++ b/toonz/sources/common/tcontenthistory.cpp
@@ -26,8 +26,12 @@ using namespace std;
 
 //-------------------------------------------------------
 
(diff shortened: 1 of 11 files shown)


melieconiek:
nice! I'll give this a try as soon as I can!

PR #1217, opened 2023-09-21
Light Table

manongjohn:
This PR adds a new feature that mimics ToonBoom's Light Table feature.

Toggling `Light Table` on will display the current level at the level's normal opacity.  All other levels will appear at 30% of their  level's opacity. 

This will also work with `Onion Skin` and `Shift and Trace` features.

The feature can be toggled on via menu option `View` -> `Light Table` or right-click the Frames area on the Xsheet/Timeline and select `Light Table` from the context menu.

If added to Command/Quick toolbars, the icon will appear as

### DIFF
--- a/toonz/sources/include/toonz/onionskinmask.h
+++ b/toonz/sources/include/toonz/onionskinmask.h
@@ -46,7 +46,8 @@ public:
   };
 
 public:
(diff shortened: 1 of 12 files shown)


hhsyya:
This Light Table function is very good, but it’s a pity that the upper and lower layers cannot be edited and shifted.

PR #1218, opened 2023-09-21
Current Drawing on Top

manongjohn:
This PR adds a new feature that mimics ToonBoom's Current Drawing on Top feature.

The feature can be toggled on via menu option `View` -> `Current Drawing On Top` or right-click the Frames area on the Xsheet/Timeline and select `Current Drawing On Top` from the context menu.

If added to Command/Quick toolbars, the icon will appear as


F-Burning:
Amazing feature.

Thank you!

PR #1221, opened 2023-09-24
Magnet Tool slider - non-linear preference

manongjohn:
It was requested that the Magnet tool's slider non-linear mode be removed.

As some may still prefer the non-linear mode, I've added a `Preference` -> `Tools` setting to accommodate individual user preference.

The option is enabled by default and requires a restart when changing this setting.


manongjohn:
Had to resolve some conflicts due to recently merged changes.  

While I was at it, I changed the default Preference settings for `Magnet Tool Size Slider - Non-linear Mode` to OFF.

Will merge once all builds have completed successfully

Issue #1200, opened 2023-08-16
Importing preferences


BertWik:
### Bug Description

In the latest nightly build (August 15) importing preferences from an earlier version causes the program to crash.

### Relevant logs/screenshots

<img width="776" alt="Scherm­afbeelding 2023-08-16 om 10 31 57" src="https://github.com/tahoma2d/tahoma2d/assets/140490349/abc0cfd3-f07a-4453-852b-4def7234f61e">

### Software Version

Nightly (Specify below)

### Software Version Information

Build August 15

### Operating System

Mac OS Ventura

BertWik:
The latest nightly build (2023-08-29) crashes while trying to open.
I use a Mac M1

Regards,
Bert Wikkerink

PR #1227, opened 2023-10-03
Change when Magnet tool's non-linear preference is applied

manongjohn:
This PR changes the application of the Magnet tool's non-linear preference setting, added in #1221, to when the tool is first activated instead of first created.

Checking preferences at tool creation has a negative side effect on macOS (Linux too?) which results in the same issue as #1200. The crash is caused by preferences loading before the application name is set since tools are created  before T2D actually starts running to set the name. 

I plan on merging this once I confirm the macOS artifact imports preferences successfully.

### DIFF
--- a/toonz/sources/tnztools/magnettool.cpp
+++ b/toonz/sources/tnztools/magnettool.cpp
@@ -150,9 +150,6 @@ public:
   {
     bind(TTool::Vectors);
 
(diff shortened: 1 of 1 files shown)


manongjohn:
Verified macOS artifact can import preferences.

PR #1234, opened 2023-10-05
Update macOS  build env to Monterey

manongjohn:
Homebrew has stopped supporting qt@5 on Big Sur.  This is causing builds on Big Sur to fail when it tries to compile qt@5 and exceeds the build timeout of 6 hours.

This change will update to build on macOS 12 (Monterey).

### DIFF
--- a/.github/workflows/macOS_build.yml
+++ b/.github/workflows/macOS_build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   macOS:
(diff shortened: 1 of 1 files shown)


manongjohn:
@artisteacher Can you verify the artifact on this PR works?

I can't verify it until I can get a Monterey VM up and running.

artisteacher:
It opens without issue.

manongjohn:
Thanks for verifying!

Issue #1238, opened 2023-10-09
Excess fill is occurring when pen is moved after starting a fill.


TomDoingArt:
### Bug Description

Tahoma does the first fill, then after a split second does a second fill, filling all unclosed areas on the canvas.

### Steps to reproduce

Situation: Using a stylus/pen and the Normal fill tool on a SmartRaster level with refer visible ticked.
Touch the screen to initiate a fill on a shape.
Lift the stylus and move it away from the area touched.
Tahoma does the first fill, then after a split second does a second fill, filling all unclosed areas on the canvas.

### Relevant logs/screenshots

![correct](https://github.com/tahoma2d/tahoma2d/assets/44948041/25a42f8f-c591-47ae-8d0e-88dc15d1c313)

### Software Version

1.3.1

### Operating System

Windows 10

### Graphics Card

NVIDIA GTX 1080Ti

### Drawing Tablet

XP Pen

beeheemooth:
I was getting the same behavior when using the Frame Range mode of the Fill Tool (Smart Raster). Filling occurred immediately, as if the mode was not turned on. Apparently the fill operation started and ended immediately in the current frame, as if there were two clicks. I thought my stylus was broken...

manongjohn:
There are 2 issues going on here....

1. There is a bug where when you do a normal mode click+drag refer visible fill on a Smart Raster  level, will fill all empty pixels with the fill color when it encounters a different colored pixel.  It produces the same result as the video shown in this issue.

2. When tapping with a stylus and moving (without touching tablet), there is a `TabletPress` event followed by a bunch of `TabletMove` events before issuing a `TabletRelease` event.  Usually the cursor position of the press and move event are close enough, but some tablets may produce points far away from each other.

In the case of the 2nd issue, the `TabletPress` immediately performs the 1st fill, but then the `TabletMove` events, treated as a drag, also try do to a fill where it thinks the cursor is.  Normally these secondary fills fail because the pixel beneath the cursor has already been filled.  However, coupled with the 1st issue it flood fills the cell. 

  Side note:  Following a `TabletRelease` and `MousePress` event is issued and also attempts to do a fill operation, but usually fails because the pixel where it thinks the cursor is at already contains the fill color.

I have a fix for the 1st issue, which might help with both issues.  If anything it will stop filling the entire cell and only impact the adjacent fillable region.

PR #1240, opened 2023-10-11
Fix potential pen-drag actions when tapping

manongjohn:
This is in response to the issue 2 noted in https://github.com/tahoma2d/tahoma2d/issues/1238#issuecomment-1755218767.

Currently after a `TabletPress` event, the 1st `TabletMove` event is processed and the internal timer skips any following move events for a very brief period.  The problem this fixes is a simple pen tap can create a series of `TabletMove` events before the `TabletRelease` event shows up.  Depending on the tool, a pen-drag action could be triggered.

In the case of the Fill tool, this can cause unwanted fills depending on where the pen detects the cursor immediately after the tap.  I've modified the logic to ignore the 1st `TableMove` after a `TabletPress` so there is time for the `TabletRelease` to appear.  If the release doesn't show up in time, assumed an intential pen drag, then the next `TabletMove` after the skip timer will take over as normal.

I would have included this with #1239, however, this change can impact all tools and requires a bit more testing on all OSes to make sure nothing is broken.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -400,6 +400,12 @@ void SceneViewer::tabletEvent(QTabletEvent *e) {
         m_tabletState       = Released;
         mouseEvent.m_button = Qt::LeftButton;
         onRelease(mouseEvent);
+      } else if (m_tabletState == StartStroke) {
+        // Single tap of stylus still records TableMoves before TabletRelease.
+        // Skip the 1st TabletMove to give time for TabletRelease to show up
+        // This way we don't try to do a LeftButtonDrag operation (i.e. normal
+        // fill) too soon.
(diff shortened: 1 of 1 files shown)


manongjohn:
I've tested this on Windows with my tablet and have heard from another Windows user and seems to be ok.

Any macOS users available to test pen dragging behavior in various tools?

I don't expect it to but just want to make sure this change doesn't cause other issues.

artisteacher:
Sorry for the late response!

It's working well for me on macOS. The only issue I've noticed so far, which may have been a thing before, is the Geometric Tool's arc and multi-arc modes. You click to create the start of the arc, click to create the end, and then can only click to create the top of the arc without the ability to adjust the location. It would be helpful to be able to do a click and drag for the top of the arc, so that you can adjust the apex of the arc on the fly.

PR #1251, opened 2023-10-21
Sub-scene Navigation Bar

manongjohn:
This new feature will make it easier to visualize and navigate multiple nested sub-scenes.

image

The panel in the timeline/xsheet shows you where you are in nested sub-scenes (`Scene Depth`), relative to the main. You can click on any prior sub-scene breadcrumb to return to it quickly.  Breadcrumbs will also show what sub-scenes in the current timeline/xsheet you can open.  Click on the breadcrumb to enter it.  In cases of multiple sub-scenes in the current timeline/xsheet, a dropdown will be shown.  Select the one you want to enter.

This new panel can be toggled on from column header's context menu, a defined shortcut or in Preferences -> Scene

image

image

### DIFF
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2255,6 +2255,14 @@ Ruler {
 #RenameCellField {
   padding-right: 3;
 }
+#XsheetBreadcrumbs {
+  padding: 0;
+  margin: 0;
+  border-bottom: 1 solid #111111;
+}
+#XsheetBreadcrumbs::separator:horizontal {
+  margin: 0 0 0 2;
+}
 /* xsheetColumnHeader (Context Menus)
 ----------------------------------------------------------------------------- */
 #xsheetColumnAreaMenu_Preview {
--- a/stuff/config/qss/Darker/Darker.qss
+++ b/stuff/config/qss/Darker/Darker.qss
@@ -2255,6 +2255,14 @@ Ruler {
 #RenameCellField {
   padding-right: 3;
 }
+#XsheetBreadcrumbs {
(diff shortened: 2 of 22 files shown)


F-Burning:
Very useful, thank you ❤️

DarrenTAnims:
This looks great and is something I've wanted since I first started using subScenes.
I wonder if it would make subscenes easier to use to have a button on the column header, to open (to into) a subscene?

manongjohn:
@DarrenTAnims 

> 1. I wonder if it would make subscenes easier to use to have a button on the column header, to open (go into) a subscene?

For the Timeline layout, I'd probably have to widen the headers to accommodate a new button.  Not sure I want to really do that just for 1 specific level type.  If there were other things I could put in this new button column, I'd be up for it.  There really isn't room for a button in the Xsheet layout.

I'm open to suggestions, though with the breadcrumbs visible, it would be the same as having a button for it, but better in my opinion because now you don't have to be on the sub-scene column to enter it.

> 2. Could the text be moved to the far left, to leave room for subScenes with long names?

Sub-scene names are the column names from the sheets they are in.  I can't imaging they'll be _that_ long. :)  Also hard to imagine a significant amount of nested sub-scenes, but who knows...I've been wrong many times before. lol.

I was trying to line it up so the list was over where the frames started in Timeline view. I feel it looks nicer and easier to just look up instead of up and left, but that's just me.  Will take additional feedback on this before making a change.

DarrenTAnims:
I appreciate your thoughts @manongjohn. I need to download this and try it out to fully appreciate how it works in practice. I'm so pleased you've managed to add this though. It'll be a great addition.

PR #1252, opened 2023-10-21
Fix wrong raster project palette

### LINKED ISSUE #1229: Raster palette not loading
### Bug Description

Raster palettes are not saving in the latest nightly builds - not sure how long this has been going on. They seem to save in the sandbox but not for projects. Upon opening, the new colors are all gone. Saving a default raster palette also fails. Save as doesn’t seem to respond. 

### Steps to reproduce

Make a project & a scene
Create a raster level 
Add some colors to the palette (and optionally save as default palette)
Close T2D
Palette changes/default don’t load

### Relevant logs/screenshots

_No response_

### Software Version

Nightly (Specify below)

Oct. 3

### Operating System

macOS Ventura

### Graphics Card

_No response_

### Drawing Tablet

_No response_

manongjohn:
This fixes #1229.

- The loading of the Symmetry tool at startup was causing the default environment file to load instead of the one in user's profile, so it would start with the sandbox project.

Similar to the Perspective Tool, delayed the initialization of the Symmetry tool until after the user's profile information was loaded.

- Fixed a bug where the wrong raster palette was being shown when switching to a scene in a different project.

The raster palette for the old project was not being changed to the new project's raster palette.

In this scenario, I forced the new project's raster palette to load.

### DIFF
--- a/toonz/sources/tnztools/symmetrytool.cpp
+++ b/toonz/sources/tnztools/symmetrytool.cpp
@@ -239,17 +239,6 @@ SymmetryTool::SymmetryTool()
   m_color.setId("Color");
   m_useLineSymmetry.setId("UseLineSymmetry");
   m_preset.setId("SymmetryPreset");
-
-  std::wstring wpreset =
-      QString::fromStdString(SymmetryPreset.getValue()).toStdWString();
-  if (wpreset != CUSTOM_WSTR) {
-    initPresets();
-    if (!m_preset.isValue(wpreset)) wpreset = CUSTOM_WSTR;
-    m_preset.setValue(wpreset);
-    SymmetryPreset = m_preset.getValueAsString();
-    loadPreset();
-  } else
-    loadLastSymmetry();
 }
 
 //----------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


artisteacher:
Thanks so much! This fixes the default project on startup.

Just an FYI, loading a scene from a different project through the startup pop-up without changing the project still doesn't load the correct raster palette. This was a pre-existing behavior though.

manongjohn:
> Just an FYI, loading a scene from a different project through the startup pop-up without changing the project still doesn't load the correct raster palette. This was a pre-existing behavior though.

Apparently this happens in Windows also, and I never noticed it before.  Will look into also fixing it with this PR.

artisteacher:
I hadn't noticed it either. I have a very observant student right now!

artisteacher:
And just an FYI, it also seems to happen whenever loading a recent scene file from the file menu (assuming its a different project) but not when loading directly through the Browser.

manongjohn:
> And just an FYI, it also seems to happen whenever loading a recent scene file from the file menu (assuming its a different project) but not when loading directly through the Browser.

Ok.  I think I have it fixed for the cases you mentioned.  Let me know if there is a case where it still doesn't work.

PR #1263, opened 2023-10-30
Fix column settings menu below screen

manongjohn:
This will fix the issue where the column's additional settings menu partially appears below screen.  It will move the menu up to allow the entire menu to be visible as well as to the right so the original button is not covered.


artisteacher:
Works well on my end!

PR #1271, opened 2023-11-05
Onion Skin All Frames or New Exposures Toggle

melieconiek:
Adds the onion skin `Onion Skin On All Frames` or `Onion Skin On New Exposures` setting to the right click menu replacing the numbering mode check. Saves and restores both the `Onion Skin On All Frames` and `Extend Onion Skin to Scene` settings.

Rendering the onion skin on all frames is the default. Option is only visible when the onion skin is enabled, same as other onion skin settings.

### DIFF
--- a/toonz/sources/include/toonz/onionskinmask.h
+++ b/toonz/sources/include/toonz/onionskinmask.h
@@ -46,8 +46,7 @@ public:
   };
 
 public:
-  OnionSkinMask()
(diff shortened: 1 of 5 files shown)


manongjohn:
Looks fine.  Still not too keen on `... New Exposures`.

I'll include in 1.4 as it's a small enough, yet  useful, change.

Will have to update translation files again after this is merged to account for the new text, but will wait for anyone who are actively translating to submit their changes first.  Might not get translated.

manongjohn:
@melieconiek Thanks for making this change.

Issue #1248, opened 2023-10-19
Build fails on Manjaro/Arch (and other systems using OpenCV with OpenEXR 3.2+) if libdeflate isn't explicitly linked


Razzaline:
### Bug Description

When I compile Tahoma2D, following the Linux build instructions either as is or with a modified PKGBUILD, it always stops around 62% after it says it finished building something, e.g. `toonzqt` or `tnzstdfx`, with the error `make: *** [Makefile:136: all] Error 2`. If I try compiling again a few times, somehow it seems to get past that part and I can even get to 100% (after it says it finished building the `Tahoma2D` executable) but still have this error. Adding `LANG=C` and `VERBOSE=1` doesn't give any more helpful/meaningful output.

This has been a problem since around August or September, and I don't know what changed.

Edit: Possibly something to do with `openexr` updating from 3.1.11 to 3.2.0 and installing `libdeflate` as a dependency (has a linker error, usually further up around 42% or 43%, about undefined references in `libimage` to `libdeflate`).

### Steps to reproduce

1. Start building from command line, following instructions
2. Error occurs

### Relevant logs/screenshots

```
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_free_decompressor'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_alloc_compressor'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_zlib_decompress'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_zlib_compress_bound'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_zlib_compress'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_free_compressor'
/usr/bin/ld: ../lib/tahoma2d/libimage.so: undefined reference to `libdeflate_alloc_decompressor'
collect2: error: ld returned 1 exit status
```

### Software Version

Other (Specify below)

### Software Version Information

Commit 919d53815 (or whichever is latest at the time, since I try again with each commit)

### Operating System

Manjaro with Linux kernel 6.5.7

manongjohn:
Can you copy/paste the output of the build into a file and add it to this issue so i can see the real compile error which should have been thrown somewhere before it stopped?

Razzaline:
[...]

PR #1274, opened 2023-11-10
Link libdeflate for OpenEXR 3.2+ (cmake)

Razzaline:
This just adds a line to the main CMake file, fixing #1248 for now (not compiling in Linux because of libdeflate not being explicitly linked). If there's a better fix, I don't mind others suggesting something, or adding to or changing it.


manongjohn:
The build failed as I had expected since currently libdeflate is not a required dependency and therefore not loaded on current Ubuntu builds.

```
/usr/bin/ld: cannot find -ldeflate
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Razzaline:
Ah, okay. I just looked into it and it does appear that Ubuntu and a lot of other distributions don't have OpenEXR 3.2+ in their package repositories. I don't really know how to proceed or check for that, though... Maybe with a CMake module checking the OpenEXR version? Technically, OpenCV can be compiled without OpenEXR, but every distro that I looked at seems to require OpenEXR with OpenCV.

manongjohn:
As part of T2D builds, I locally compile OpenCV to adhere to an LGPL compliant license. It's possible I have exluded OpenEXR on purpose which explains why it isnt looking for it or its dependencies (i.e. libdeflate).

Razzaline:
Okay, that makes sense. After you mentioned that OpenEXR wasn't included, I checked and noticed that in the CI build scripts too.

Sorry about the merge commit there >.< I hadn't realised it wouldn't ask me for confirmation when I told it to sync. But I've added a hopefully slightly better solution? It checks for the OpenEXR version and if it doesn't find it I think it should just ignore it. I don't have a lot of experience with CMake so I'm not entirely sure. Otherwise, it does seem to be linking libdeflate properly for me.

Razzaline:
I can probably add the same logic for Windows and Mac OS, in the unlikely event that someone's build environment happens to have OpenEXR 3.2+ on an OpenEXR-enabled build, although I guess that might happen eventually anyway so it's good to have the checks in place.

Razzaline:
For now I just added the Apple build environment stuff. Adding the check on Windows might involve creating a FindOpenEXR.cmake (there are no prebuilt binary releases of OpenEXR for Windows) so I thought I should leave it for later.

Razzaline:
I'm planning to finish this soon if possible (and was hoping to before the 1.4 release), but I've been stuck trying to figure out the Windows part of it (hence running a build in a virtual machine and encountering https://github.com/tahoma2d/tahoma2d/issues/1300). I've made a few additional changes I haven't pushed yet, but if all else fails I could just push what I have with the Windows parts commented out (though they are written in a way that under normal circumstances CMake should just skip over that).

manongjohn:
Closing this PR since I've submitted my own PR with different changes.

Issue #1259, opened 2023-10-27
Align and Distribute Panel 


erayzesen:
**Is your feature request related to a problem? Please describe**
I feel the lack of an align tool, especially for quickly positioning drawings on vector levels (such as placing a head in the exact center of the body or aligning the drawing to the center of the scene, etc.) 

Of course, it would be great if the same feature could be implemented for raster levels as well. For positioning in raster levels, data from pixel drawing boundaries can be used. Just like in vectors, there's actually a need for knowing the boundaries, using a bounding box approach, for this feature. The selected object or objects can be centered, aligned to the right, or aligned to the left based on their boundaries.

**Describe the solution you'd like**
Very simple panel implementation is like : 

Relative to (Checkbox)- Canvas, first selected, last selected
Align (Buttons): 
Left , Center, Right , Top, ,Bottom...

It could be something similar. In fact, it can be entirely inspired by the align panel in Inkscape. 

![Screenshot_2023-10-23_00-29-31](https://github.com/tahoma2d/tahoma2d/assets/65820625/f2132181-cdbe-4ba5-a236-f31ed3509628)

beeheemooth:
Oh yea, these are really useful options (I remember them in Flash)! I’m not sure that this way it will be possible to align something in raster levels, but in vector levels these options are definitely lacking in Tahoma/OT. Aligning objects is the weak point of the program at the moment; everything needs to be done “by eye”.

erayzesen:
> Oh yea, these are really useful options (I remember them in Flash)! I’m not sure that this way it will be possible to align something in raster levels, but in vector levels these options are definitely lacking in Tahoma/OT. Aligning objects is the weak point of the program at the moment; everything needs to be done “by eye”.

I don't know a  information about the raster levels side but I think vector levels needs this feature. And yes, flash(or adobe animate) has this feature. To be honest, I don't care which software will be the inspiration for implementing this feature. Just as long as we have this capability. I don't think it would be very difficult to add this feature. I thought to add it myself, but I couldn't even compile Tahoma or Opentoonz. : (

beeheemooth:
There is another problem. In Flash it is possible to select objects from different layers at the same time, but in Tahoma/OT you cannot do this - you always work only with the current column (I remember that this was unexpected news for me when I started doing animation in OT). Thus, the function, if it is implemented, will only be able to work with objects (groups) within the same level.

erayzesen:
[...]

PR #1275, opened 2023-11-11
Vector Control Point/Stroke Alignment Commands

manongjohn:
This adds a set of new commands for aligning or evenly distributing vector control points (via `Control Point Editor Tool`) and vector strokes or groups of strokes (via `Selection Tool`).

The commands can be found in the the new item `Panels` -> `Align and Distribute Panel`:

image

You can align strokes relative to the following options:
- `Selection Area` - Relative to selection area box
- `First Selected` - Relative to the first selected object
- `Last Selected` - Relative to the last selected object
- `Smallest Object` - Relative to the smallest object in the selection area
- `Largest Object` - Relative to the largest object in the selection area
- `Camera` - Relative to the camera box

For aligning strokes or group of strokes, the edge of the bounding box is used for left, right, top, and bottom alignments.  Center alignments and distribution commands use the horizontal or vertical midline of the bounding box.

Distribution commands use the 2 furthest objects as anchors and space the remaining objects between them evenly.

Additionally, I have converted the following menu options into commands (with icons)

image

All these can be added to Command Bars and assigned shortcuts.


RodneyBaker:
This is an excellent addition to vector tools.
Looking good!

erayzesen:
OMG. What a beautiful piece of work! I had opened a similar feature request in the tahoma and opentoonz subdirectories a few days ago. 
https://github.com/tahoma2d/tahoma2d/issues/1259

While I may not contribute to the software development part, I'd be happy to help with testing if there's anything I can do. I use Linux, but I can also test your Windows builds.

RodneyBaker:
There is a Linux download as well @erayzesen 
That can be found at the bottom of every PR (Pull Request) where it says "Show All Checks".
Those same automated builds can be found via the Actions link at the very top of these github pages.

erayzesen:
> There is a Linux download as well @erayzesen That can be found at the bottom of every PR (Pull Request) where it says "Show All Checks". Those same automated builds can be found via the Actions link at the very top of these github pages.

Thank you @RodneyBaker .  I 'm testing windows builds right now. Then I can  also test linux builds.

manongjohn:
Forgot there was an open FR for Alignment/Distrute commands.  I did use Inkscape to see what options it provided.

I only implemented selection alignment/distribution for ease of use. As such, I didnt create a panel for it. 

Having said that, aligning/distribute based on at least the camera dimensions could be useful. Either i can make a separate set of commands for that or actually build a new panel and provide other options (i.e. largest/smallest object).  Something to think about. 😁

artisteacher:
> actually build a new panel and provide other options (i.e. largest/smallest object)

I vote for a new panel!

erayzesen:
I tested on linux and windows builds for all scenarios. On windows side everything works well. But linux clang builds give me some bugs to me when I use aligment commands in selection tool mode on vector levels.   

My steps:
- Draw a rectangle and a circle in vector layer
- -Fill two shapes to different color.
- Select all in selection mode.
- Use alignment commands. 
- Tahoma is crashing or shapes disappear.

But if I convert one of shapes to group, all align commands work well. I don't know how can i get crash reports. 

> Having said that, aligning/distribute based on at least the camera dimensions could be useful. Either i can make a separate set of commands for that or actually build a new panel and provide other options (i.e. largest/smallest object). Something to think about.

Maybe a sub menu can be added for this in the alignment menu. 
Edit-> Alignment->Relative to
|
->Bounding box
->Camera

manongjohn:
> * Tahoma is crashing or shapes disappear.
> 
> But if I convert one of shapes to group, all align commands work well. I don't know how can i get crash reports.

When it crashes, it doesn't bring up the Crash Report dialog with the crash report?

erayzesen:
> > * Tahoma is crashing or shapes disappear.
> > 
> > But if I convert one of shapes to group, all align commands work well. I don't know how can i get crash reports.
> 
> When it crashes, it doesn't bring up the Crash Report dialog with the crash report?

Opening a popup. But Tahoma is closing then it. I can't see a report because of that. If there is an alternative way, I can try.

manongjohn:
@erayzesen Can you download the gcc version and see if the crash happens there and if the crash reporter shows up? What distro are you on?

erayzesen:
I tried the same steps again, I didn't get any bug or crash. Alignment commands works well.  Maybe it's not about alignment commands. I saved the scene where I got a bug, can i send it to you?

> @erayzesen Can you download the gcc version and see if the crash happens there and if the crash reporter shows up? What distro are you on?

Of course.

erayzesen:
Ok I finded the problem. It's about undo-redo functions. I'm updating my steps for you; 

-Create a vector level
-Draw a rectangle shape
-fill the shape with bucket tool
-Draw a circle shape
-Fill it bucket tool
-Select all
-Apply edit->alignment->align left
-Control-z (undo)
-Apply edit->aligment->align center horizontally

The problem is starting undo-redo functions. I will try these steps with linux-gcc version and windows version.

erayzesen:
I got this bug in all linux builds . I didn't get any bugs in windows build.

manongjohn:
I will see if i can duplicate in my Ubuntu VM later today. What distro are you on?

erayzesen:
> I will see if i can duplicate in my Ubuntu VM later today. What distro are you on?

I'm using last version of Linux Mint. I also recorded a video for you. 

https://github.com/tahoma2d/tahoma2d/assets/65820625/b60bad7a-fb64-4bda-af24-7d8a6664c34b

manongjohn:
Interesting. It looks like it tried to open the crash log. Have ypu tried running command line? Might throw something the console.  Other option is to run it through xdb.

manongjohn:
I just downloaded the Linux clang artifact and tried it out on my Ubuntu 20.04 VM.  I could not duplicate the issue.  It worked fine without crashing when I performed your steps.

erayzesen:
I started Tahoma2D from terminal and I duplicate the issue. Output is:

```
create context:0x345e040 [thread:0xd05cac80]
create context:0x3420040 [thread:0xd05cac80]
create context:0x7fe4c8007490 [thread:0xd05cac80]
create context:0x7fe480005430 [thread:0x96df0640]
QCursor: Cannot create bitmap cursor; invalid bitmap(s)
QCursor: Cannot create bitmap cursor; invalid bitmap(s)
/tmp/.mount_TahomaSK551t/AppRun: line 14:  3760 Segmentation fault      (core dumped) $APPDIR/usr/bin/Tahoma2D
```

I don't know why it's happen. 

> I just downloaded the Linux clang artifact and tried it out on my Ubuntu 20.04 VM.

Ubuntu version of Linux Mint is 22.04.Could it be related to this?

artisteacher:
I'm not getting the crash on macOS.

RodneyBaker:
I'm still investigating this one because I'm getting different results at different times.

I have noted that when aligning objects (say groups of geometric shapes), after the alignment is complete the objects are sometimes deselected.  This requires the user to reselect the shapes again if they wish to continue to align those shapes.

I note that in programs such as Coreldraw the selection is maintained after the alignment is completed.

In the test I just did prior to submitting this feedback the selection was maintained... so... not sure what is going on here.
Edit:  Ah, it seems this (also) might related to Undo and when an alignment is undone the selection is released.

manongjohn:
> I vote for a new panel!

Panel created! Added relative alignment options, similar to Inkscape.  PR description has been updated with details.

manongjohn:
Added aligning to First/Last Select, similar to Inkscape.

I had to modify how the selected strokes/points was stored internally to preserve the order of selected items. Please report any weirdness when working with selected strokes/command points.

artisteacher:
Love the new panel. The additional alignment area options are so helpful!

artisteacher:
It might make sense to add the vector arrangement commands to this panel too - ie. move to front, back one & etc.

RodneyBaker:
@manongjohn An observation regarding the relative to camera option when using the distribute tools...

If we change the size or position of the camera via Timeline/Xsheet the distribution still remains with the previous camera setting even if the scene is saved and reopened with the new camera position.

The distribution relative to camera likely uses the actual camera resolution as opposed to the current camera size and location.
As such I'm not sure how to better to guide users so they don't think adjusting the camera via keyframing will effect this feature when using the relative to camera option.

Apologies if this isn't clear.
I'm still exploring.

Note that we do get the new camera size that effects the relative to camera option when we change the resolution via Camera Settings.  An unfortunate side effect of this feature being for vectors however is that the option to resize camera to current level is not an option for vector levels... only for raster/smart raster.

manongjohn:
> The distribution relative to camera likely uses the actual camera resolution as opposed to the current camera size and location.

Yes.  Align/Distribute uses the original camera dimension and position, even if the camera is animated.  Would changing it to say `Camera Dimensions`  be better?  I don't want to have to include camera animation when using that mode.

RodneyBaker:
Perhaps ```Camera Settings```?
For my part I think I can just remember the difference but someone new would need to discover that.
Of course somone new might not know how to change Camera Settings either.  ;)

```Camera Dimensions``` to me reflects the same idea that it is the current dimensions of the camera that will be used.

manongjohn:
`Camera Settings` would seem rather odd in this context.  I think the moment it says `Camera XXX` people will likely think it's based on current position/scale of camera.  Instead of `Camera Resolution`, how about `Camera Size` or `Camera Area`?

RodneyBaker:
All of these seem equally confusing so perhaps just leaving it as is stating ```Camera``` might be best.  
Users will discover proper usage soon enough.  :)

manongjohn:
I think the problem with accounting for camera animation is when the camera is rotated, scaled and/or skewed.   What should happen in those cases?

erayzesen:
Although it doesn't bother me, the camera option can be confusing from a software logic perspective. The other added options are already able to address all possible needs. However, if this tool can also be applied among selected levels, then there may not be a need for this option. Can this feature, within the selected drawing levels, allow the positions to be changed based on the bounding boxes of the levels in the relevant frame? Of course, if this is possible, it would be used in the animate tool, directly manipulating the positions of the levels.

manongjohn:
Playing around a bit, I have a change for adjusting for camera animation which I will push in a bit.  Alignments on rotated/sheered cameras  would be based on the midpoints of the edges being aligned against.

Vector Selection allows for different selection Modes (specific frames, whole level, other vector levels, etc), however align/distribute currently only work against the current level and frame.  If it applied to other levels/frames, then it would be based on the selection area of the strokes on those levels/frames, not the current one, I think.  Are you expecting it to be based on the current selection area defined by the currently selected strokes?

I think I need some example scenarios of how this would apply to other levels/frames.

erayzesen:
> Playing around a bit, I have a change for adjusting for camera animation which I will push in a bit. Alignments on rotated/sheered cameras would be based on the midpoints of the edges being aligned against.


> Playing around a bit, I have a change for adjusting for camera animation which I will push in a bit. Alignments on rotated/sheered cameras would be based on the midpoints of the edges being aligned against.

So, an alternative expression like 'Camera's Bounding Box' can be used to define this.

> Are you expecting it to be based on the current selection area defined by the currently selected strokes?

My previous writings might not have been very clear. These features at the vector level are fantastic, and I believe they integrate well with the software. What I was referring to was the ability for the position between levels to change when the animate tool is selected as an alternative to alignment based on the camera. (assuming it's technically possible) This is because your work with the control point editor tool works on points in vector levels, the selection tool works on shapes and groups, and I'm curious if this feature could be applied between selected levels in the animate tool.

If this were the case, instead of alignment options based on the camera dimensions we're discussing, I could draw the rectangle I want to align to a different level and position the other levels according to that layer. It's similar to aligning different layers in other software like Inkscape.

manongjohn:
> It's similar to aligning different layers in other software like Inkscape. With this way, nobody needs camera option.

I'm still a newb when it comes to Inkscape.  Didn't realize it had layers, much less aligning things across layers. lol

I just played with it briefly, and now I understand what you mean.  Vector selection does allow to work across multiple vector levels within the scene frame, and even multiple frames.  Would need to play a bit to see how to make it behave similar to Inkscape.

erayzesen:
@manongjohn  I just wonder whether  it's possible.  Maybe it's not possible on Tahoma2D or the implementing is very hard. If I'm not wrong, Tahoma2d/Opentoonz doesn't has the user scenario just like the modifying multiple levels.

manongjohn:
It's possible, but would need to add new Vector Selection Modes to allow for selecting strokes in other visible layers.  I think that would take a bit of coding and is out of scope of this particular feature.  If that is something worth adding, a new FR should be requested for that. Aligning/Distributing across multiple layers could then addressed at the same time.

erayzesen:
> It's possible, but would need to add new Vector Selection Modes to allow for selecting strokes in other visible layers. I think that would take a bit of coding and is out of scope of this particular feature. If that is something worth adding, a new FR should be requested for that. Aligning/Distributing across multiple layers could then addressed at the same time.

I understand, you're right. This actually requires the addition of the 'Multiple Layer Selection and Editing' feature first, doesn't it? I can grasp how much work it might involve, considering Gimp has been lacking this feature for years : ) 

Anyway, to be honest, I mainly work with vector levels, and this feature will make me and many others happy. This feature you've implemented was one of the two missing features in the vector drawing aspect of my Tahoma2d. The other was the absence of the option to round the corners of rectangles (like in Inkscape, Illustrator, or Adobe Animate).  I'm planning to make an another feature request for that as well, but I don't think I'll be as lucky. : )

I hope in the future I'll have the time and knowledge to contribute to Tahoma2D. I have some fantastic features in mind that could be added.

erayzesen:
I want to share that is what's a great feature. I can make good rounded rectangles with this feature. Thank you again for adding this feature. 

https://github.com/tahoma2d/tahoma2d/assets/65820625/1de0a92d-8e84-4aeb-ba18-4d06348b716a

PR #1277, opened 2023-11-12
Add icons for Export commands

manongjohn:
There was a request for some icons for a couple of Export commands.  Since I appear to be a glutton for punishment, I decided to make one for all Export commands.

image


DarrenTAnims:
These icons look really nice and clean.
However, traditionally, as far as I recall, arrows moving up or left show importing. Down or right show exporting, saving out, rendering, printing, etc.

manongjohn:
I used other already T2D icons to build these in order to maintain cohesiveness.  The Import commands use a down arrow and the few Export commands we did have (i.e. Export level, Export Scene from Browser menu) use an Up arrow.

I didnt create the referenced icons. 🙂

PR #1281, opened 2023-11-15
Minor documentation fixes

Razzaline:
While editing the CMake files for https://github.com/tahoma2d/tahoma2d/pull/1274, I noticed that the instructions for building on Windows said to put the libjpeg-turbo64 files into libjpeg-turbo64, but the CMake file itself looks for libjpeg-turbo, so I corrected that. In the process, I also noticed that the numbered list in [CONTRIBUTING.md](https://github.com/tahoma2d/tahoma2d/blob/master/CONTRIBUTING.md) was all zeros, so I fixed that too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,25 +10,25 @@ I will first review the request, then I'll accept it, add comments for rework, o
 
 ### Workflow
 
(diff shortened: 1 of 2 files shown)


manongjohn:
Thanks for fixing the instructions.

PR #1285, opened 2023-11-18
Warn when unable to load/read file

manongjohn:
This change was primarily to fix a crash that was caused when a user loaded a JPEG file but could not be properly read. The file was actually a WEBP file with a JPEG extension for some reason.

When a file cannot be opened or read properly, it will give a message that it can't load the file and abort the operation.

I've extended this fix to loading other file types which can happen via Load Level, loading into Flipbook or loading a scene.

### DIFF
--- a/toonz/sources/common/tiio/tiio_bmp.cpp
+++ b/toonz/sources/common/tiio/tiio_bmp.cpp
@@ -10,6 +10,7 @@
 #include "tpixelgr.h"
 #include "tproperty.h"
 //#include "tconvert.h"
+#include "texception.h"
 #include <stdio.h>
 
 //=========================================================
@@ -156,7 +157,10 @@ void BmpReader::open(FILE *file) {
 //---------------------------------------------------------
 
 void BmpReader::readHeader() {
-  if (!m_chan) return;
+  if (!m_chan) {
+    throw TException("Can't open file");
+    return;
+  }
 
   fseek(m_chan, 0L, SEEK_END);
   m_header.biFilesize = ftell(m_chan);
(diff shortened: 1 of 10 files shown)


artisteacher:
Thank you!

artisteacher:
I'm still getting a crash:
[load-crash.txt](https://github.com/tahoma2d/tahoma2d/files/13420405/load-crash.txt)

manongjohn:
Seems there is a difference in the libjpeg package in Big Sur vs Monterey.   Need to set up my Monterey build environment so I can find a fix for this.

manongjohn:
@artisteacher I've put in a check to verify the JPEG file is in fact a JPEG file before processing so it doesn't run into issues with the jpeg library.

artisteacher:
Sorry, I'm still getting a crash on import of that image.
[load-still-crashing.txt](https://github.com/tahoma2d/tahoma2d/files/13480679/load-still-crashing.txt)

manongjohn:
@artisteacher  Actually, it's a different error/issue, now. I noticed this error while testing this fix in my Monterey VM.

This problem seems to have started with macOS Monterey but was fine in Big Sur.  When certain errors are thrown, we "catch" the error and in some cases  we intentionally "rethrow" the error again so another part of the system can deal with it.  Apparently Monterey suddenly doesn't like "rethrowing" our custom exception types anymore and immediately crashes.

I will not be fixing that with this PR as it's a bigger issue.

manongjohn:
This will not be included in 1.4.  I'm still looking into why it's crashing on a catch-rethrow in some cases but not others.

It will be moved to 1.5, or a 1.4.1 fix release if there is one.

PR #1286, opened 2023-11-18
Hide title bar for docked Custom Panel & Preproduction Board

manongjohn:
Like most other panels, if a Custom Panel or Preproduction Board is docked, the title bar will now be hidden when the panels are locked.


artisteacher:
This works great. Just noticed the same issue with the preproduction board, in case it matters.

manongjohn:
@artisteacher Thanks for pointing it out.  Fix pushed.

PR #1255, opened 2023-10-26
Update translation files for v1.4

manongjohn:
Hello folks. I've updated the translation files for v1.4 if anyone who wants to help update them. Any help would be appreciated!

You have until at least Nov 27 to submit translations. You can either attach them to this PR or submit your own.

Here are the translation files: [translations.zip](https://github.com/tahoma2d/tahoma2d/files/13179783/translations.zip)

PR #1298, opened 2023-11-30
Localisation additions

Razzaline:
I've added to some of the current Chinese translations. I don't think I'll be able to do all of it before 1.4, but I can still add more later. (I'd also wanted to do some translations for other languages, but I know I won't be able to get to them.)


manongjohn:
FYI, I'm going to close this since I've pulled your changes directly into my branch for #1255 so I can combine with other translation file updates and build the QM files for the release.

PR #1307, opened 2023-12-10
Clipping Mask Column (Raster/Smart Raster)

manongjohn:
This extends the Clipping Mask Column feature to Raster and Smart Raster levels.

- You can mix and match masks of different level types.
- When using the `Selection Tool` to manipulate the mask, you cannot see the outline of the mask like you can with Vector masks.

-----

Known issues:
- Drawings clipped by Raster/Smart Raster masks do not show in 3D view.
- When using Selection Tool on the mask, the selected area shows the original mask image while it is being moved or manipulated. (Not planning to fix)

### DIFF
--- a/toonz/sources/common/tfx/tfxcachemanager.cpp
+++ b/toonz/sources/common/tfx/tfxcachemanager.cpp
@@ -271,6 +271,8 @@ ResourceData TFxCacheManager::getResource(const std::string &alias,
                                           const TRenderSettings &rs) {
   TCacheResourceP result, temp;
 
+  if (rs.m_applyMask) return ResourceData(0, result);
+
   // Seek the associated infos
   Imp::ResourceInstanceDataMap::iterator jt =
       m_imp->m_resourcesData.find(alias);
--- a/toonz/sources/common/tgl/tgl.cpp
+++ b/toonz/sources/common/tgl/tgl.cpp
@@ -550,6 +550,55 @@ void tglDraw(const TRectD &rect, const TRaster32P &tex, bool blending) {
   texture->unlock();
 }
 
+//-------------------------------------------------------------------
+
+void tglDrawMask(const TRectD& rect, const TRaster32P& ras) {
+  TRaster32P mask = ras->clone();
+  int lx          = rect.getLx();
+
+  // Set used pixels to full color/alpha for masking purposes
+  for (int y = 0; y < mask->getLy(); y++) {
+    TPixel32 *pix    = mask->pixels(y);
+    TPixel32 *pixEnd = pix + lx;
+    while (pix < pixEnd) {
+      if (*pix != TPixel32(0, 0, 0, 0)) *pix = TPixel32(255, 255, 255, 255);
+      pix++;
+    }
+  }
+
+  glPushAttrib(GL_ALL_ATTRIB_BITS);
+
+  glPushMatrix();
+  glLoadIdentity();
+
+  glRasterPos2d(rect.x0, rect.y0);
+  glPixelStorei(GL_UNPACK_ROW_LENGTH, 0);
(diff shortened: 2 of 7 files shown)


beeheemooth:
_`When using Selection Tool on the mask, the selected area shows the original mask image while it is being moved or manipulated. (Not planning to fix)`_

`That's` actually good. It would be even better if a Savebox was displayed when using the Selection Tool, so that the user would have an idea of ​​the dimensions of the invisible part of the raster (at least in Smart Raster) without having to disable the mask.

What happens to the display of MyPaint styles of a masked column? When the mask is enabled in Camera View they appear with sharp edges, but in Preview mode they appear correctly. Is there a chance to see MyPaint styles in Camera View in both columns?

manongjohn:
Did you mean this?

image

I wasn't expecting this, actually.  I was mostly expecting the final render to be more like the `Viewer`.  I just realized the results differ because of the way the mask is applied in the Viewer vs how it's applied during Preview/Render.  Need to see what I can do about this.

beeheemooth:
> Did you mean this? 
Yes.

Codeoloketuyi:
Does this clipping mask resemble the one used in clip studio and another painting software

manongjohn:
T2D 1.4 currently has the Vector Level based version of this PR. Use that as comparison to other software if you like.  I personally cannot say how similar or not T2D's Clipping mask is to other software, though as for how Clipping Masks generally work, it does that job.

As for this PR, it just extends Clipping Masks so you can us Raster/Smart Raster levels instead of Vector levels.

Side note, this should not be confused with an Opacity Mask that some software also have.

Dixxi0509:
When will this come, I can't really use vector due to it crashing my computer(tho I can use the level just not the brushes so idk if I can use raster brushes/vector level and use the clipping mask)

manongjohn:
> When will this come, I can't really use vector due to it crashing my computer(tho I can use the level just not the brushes so idk if I can use raster brushes/vector level and use the clipping mask)

I'm actively still working out some issues with it. Most of the issues are somewhat minor visual issues in the Viewer in normal mode, see known issues in the description. Seems to be working otherwise.

Feel free to download the artifact from this PR and try it out. Could use the feedback for any other issues.

Dixxi0509:
Can you send me the link?

manongjohn:
> Can you send me the link?

As with any PR, you can do the following:
- Scroll down in PR to where it says `All checks have passed`
- Click on the `Show all checks` to the right of it.
- Click on the `Details` for your OS and it will take you to another page
- On the next page click on `Summary` link (upper left side of page)
- Under the `Artifacts` section, download the portable version. Don't use the installed version.
- Recommend unzipping to a different place from your main T2D to test it out.

Dixxi0509:
I got this

manongjohn:
I've removed the WIP/Draft status on this ticket. Even though non-vector based masks currently don't work in 3D View Mode. I am still working on it but wanted to open this up for further testing.  I am strongly leaning towards moving forward with it even if 3D View Mode isn't working.

artisteacher:
Seems to work well so far on macOS. I think the 3D View is the least important...

Codeoloketuyi:
Does the clipping mask work after a camera key frame as been added.
Just wondering.

manongjohn:
> Does the clipping mask work after a camera key frame as been added. Just wondering.

Download the artifact from this PR and try it out. :)

manongjohn:
In order to include this in 1.5, I'm planning on merging this by the end of the week barring someone finding a major issue.

Although I dont generally like releasing something with an obvious flaw, i think this feature's usefulness, as is, outweighs the issue that it doesnt show in 3D mode properly. I will continue to work on fixing it.

PR #1309, opened 2023-12-11
Fix Timeline/Xsheet Frame Range processing

manongjohn:
In level edit mode (Level Strip), the `Frame Range` option takes the 1st and last selected frame ids and performs the operation on all frames in between.

In scene edit mode (Timeline/Xsheet) the `Frame Range` option would take the 1st and last selected from, from a Level Strip point of view, and perform the operation as if it was in the Level Strip.  This does not work right if the exposed frames are out of level strip order (i.e 1-3-2, vs 1-2-3), or if only specific frames are exposed between the 1st and last selected frame.

This PR changes the behavior of the `Frame Range` option when working in the timeline/xsheet.  It will move through the scene frames from the 1st to the last selected and only affect the exposed frames in that span as well as affect them in the order it was exposed.

The only `Frame Range` option that worked correctly was the Fill Tool's Normal fill mode. The following were fixed:
- Fill Tool using non-Normal selection mode
- Tape Tool for Smart Raster
- Eraser Tool for all level types

This also fixes a crash caused by using the Vector Segment Eraser on a level sequence that starts with 0000 when used in Scene edit mode.

### DIFF
--- a/toonz/sources/tnztools/filltool.cpp
+++ b/toonz/sources/tnztools/filltool.cpp
@@ -1289,10 +1289,7 @@ void SequencePainter::processSequence(TXshSimpleLevel *sl, TFrameId firstFid,
     // Setto il fid come corrente per notificare il cambiamento dell'immagine
     TTool::Application *app = TTool::getApplication();
     if (app) {
-      if (app->getCurrentFrame()->isEditingScene())
-        app->getCurrentFrame()->setFrame(fid.getNumber());
-      else
-        app->getCurrentFrame()->setFid(fid);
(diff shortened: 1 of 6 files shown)


beeheemooth:
What happens if a number of frames contain frames with the same number? Will they be processed again?

manongjohn:
> What happens if a number of frames contain frames with the same number? Will they be processed again?

It will normally ignore held frames, but if there is a break in the hold and the same image is exposed in 2 different places, then yes. The selection area will move so it's possible it can affect the same image in 2 different places in that scenario.

Issue #1313, opened 2023-12-18
pasted a frame, drew and then undo'd for it to crash


Minced-Meat:
### Bug Description

copy by using selection tool, pasted a frame, drew and then undo'd for it to crash

### Steps to reproduce

copy by using selection tool, pasted a frame, drew and then undo'd for it to crash

### Relevant logs/screenshots

Tahoma2D 1.4  (Build Dec  1 2023)

Report Date: 2023-12-18 17:55:19
Crash Reason: EXCEPTION_ACCESS_VIOLATION

Build ABI: x86_64-little_endian-llp64
Operating System: Windows 10 (10.0)
OS Kernel: 10.0.19045
CPU Threads: 6

GPU Vendor: NVIDIA Corporation
GPU Model: NVIDIA GeForce GTX 1650/PCIe/SSE2
GPU Version: 4.6.0 NVIDIA 536.99

Crash File: C:\Users\USER\AppData\Local\Tahoma2D\Tahoma2D\crash\Crash-20231218-175519.log
Mini Dump File: C:\Users\USER\AppData\Local\Tahoma2D\Tahoma2D\crash\Crash-20231218-175519.dmp

Application Dir: C:/Users/USER/Documents/software/Tahoma2D-portable-win (1)/Tahoma2D
Stuff Dir: C:\Users\USER\Documents\software\Tahoma2D-portable-win (1)\Tahoma2D\tahomastuff

Project Name: sandbox
Scene Name: doodles
Project Path: C:\Users\USER\Documents\software\Tahoma2D-portable-win (1)\Tahoma2D\tahomastuff\sandbox\tahomaproject.xml
Scene Path: C:\Users\USER\Documents\software\Tahoma2D-portable-win (1)\Tahoma2D\tahomastuff\sandbox\scenes\doodles.tnz

==== Backtrace ====
  1> RtlDeleteAce [0x7ff8e20c8b1c] <ntdll.dll>
  2> RtlRaiseException [0x7ff8e20a12d6] <ntdll.dll>
  3> KiUserExceptionDispatcher [0x7ff8e20f0ebe] <ntdll.dll>
  4> QByteArray::QByteArray [0x5a3920d3] <Qt5Core.dll>
  5> TFrameId::TFrameId [0x7ff8967bcbc2] <tnzcore.dll>
  6> TXshSimpleLevel::renumber {txshsimplelevel.cpp:2211} <toonzlib.dll>
[...]

PR #1316, opened 2023-12-19
Fix undo selection paste into new frame + undo renumbering

manongjohn:
While trying to reproduce the crash mentioned in #1313, I found an issue with Undo when selecting part of a Raster/Smart Raster level and then pasting into a held frame w/ `Creation In Hold Cells` toggled `On`. A new blank frame is created with the paste, but an Undo doesn't remove the newly created frame.  There after, the undo history becomes messed up.

Corrected the logic to record the creation of the blank frame as part of the paste so it is undone correctly.

Also fixed an issue with undo and cell renumbering. After removing the newly created frame as part of the undo, an update to the level was triggered. This reset the old frame ids list which can cause the renumber to crash if the list was empty.  This may be why the crash occurred in #1313, but I could not duplicate the original issue to confirm.  Modified the logic to update the level after the renumber is complete.

### DIFF
--- a/toonz/sources/include/tools/rasterselection.h
+++ b/toonz/sources/include/tools/rasterselection.h
@@ -55,6 +55,10 @@ class DVAPI RasterSelection final : public TSelection {
   bool m_isPastedSelection;
   bool m_noAntialiasing;
 
+  bool m_createdFrame;
+  bool m_createdLevel;
+  bool m_renumberedLevel;
+
 private:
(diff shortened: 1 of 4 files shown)


beeheemooth:
Could this also be the reason for the inability to save new mesh frames after renumbering? I still get red frames when trying to save a mesh whose frames have been renumbered.

manongjohn:
No. It's a different issue

PR #1330, opened 2024-01-03
Move Columns and Cells without Drag Bars + selection/drag fixes

manongjohn:
This PR changes how columns and cells can be moved. 

You can chose the old way of movement via drag bars or the new way of just clicking and moving it. The new method makes it easier to target and is more in line with other software.

By default, the column and cell drag bars will be hidden but can be turned on in `Preferences` -> `Scene` -> `Show Column and Cell Drag Bars`.

image

When drag bars are shown, you must select and move things the old way.

When drag bars are not shown, simply click-drag a single exposed cell or a single non-empty column.  For multiple cells/columns, click-drag any cell or column inside the selection area.  NOTE: If you click on a selection area that does not have an exposed cell or key or does not have a non-empty column, the selection area will reset.

Some modifications have been made regarding how to select multiple columns or cells.

`ALT` key now selects keys instead of `CTRL`. (only when not using drag bars)
 
Ways to select multiple cells:
- Click-drag an unselected empty or implicit cell to begin cell selection. (existing behavior)
   - Stop Frame holds are not considered blank frames.  They are treated as exposed cells for the purpose of dragging.
- `CTRL`+Click-drag an unselected exposed cell to begin cell selection. (new behavior)
   - `CTRL`+Click-drag on an already selected exposed cell will resort to normal ctrl-modified cell movement behavior
- Click on a cell, then `SHIFT`+Click another cell.  A selection area will be created between the 2 cells (existing behavior)
- `CTRL` with or without other modifiers is still used to select keys

Ways to select multiple columns (Clicks are done on column names or thumbnails):
- Click-drag an unselected empty column to begin column selection. (existing behavior)
- `CTRL`+Click individual columns to add/remove column from the selection area (existing behavior)
- `CTRL`+Click-drag a column to begin column selection (new behavior)
- Click on a column, then `SHIFT`+Click another column. A selection area will be created between the 2 columns (existing behavior)

To cancel any selection area, click on any unselected column or cell. (new behavior)

Additional changes and fixes:
- When not using drag bars, you can change the timeline layout to `Compact` (default), `Roomy` or `Minimum`
- Shifted timeline/xsheet text, UI elements and targeting areas to account for no drag bars.
- When drag bars are not shown, the lock column indicator displays as black dashes in the original drag bar area.
- Fixed an issue where you could not start a multi-column selection when the starting column is a Note Level
- Fixed it so sound is not played if you click on sound icon while pressing a modifier key.
- Fixed Palette level cells not highlighting when column is selected
- Fixed issue with selection resetting when using `SHIFT`+`CTRL`-Click to select an area + keys at the same time
- Fixed issue with `CTRL`-Drag not allowing copied cells into empty frames

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -228,6 +228,8 @@ protected:
   std::map<PredefinedRange, NumberRange> _ranges;
   std::map<PredefinedFlag, bool> _flags;
 
+  QString _layoutName;
+
 public:
   virtual ~Orientation(){};
   virtual CellPosition xyToPosition(const QPoint &xy,
@@ -286,7 +288,10 @@ public:
   virtual int cellHeight() const     = 0;
   virtual int foldedCellSize() const = 0;
 
+  virtual QString layoutName() const { return _layoutName; }
+
 protected:
+  void setLayoutName(QString layoutName) { _layoutName = layoutName; }
   void addRect(PredefinedRect which, const QRect &rect);
   void addLine(PredefinedLine which, const QLine &line);
   void addDimension(PredefinedDimension which, int dimension);
@@ -298,7 +303,8 @@ protected:
 
 // Enumerates all orientations available in the system as global const objects.
 class DVAPI Orientations {
-  const Orientation *_topToBottom, *_leftToRight;
+  const Orientation *_topToBottom, *_leftToRight_roomy,
+      *_leftToRight_nodragcompact, *_leftToRight_nodragminimum;
   std::vector<const Orientation *> _all;
 
   Orientations();
@@ -308,7 +314,7 @@ public:
 
   static const Orientations &instance();
 
-  static const int COUNT = 2;
+  static const int COUNT = 4;
 
   static const Orientation *topToBottom();
   static const Orientation *leftToRight();
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -444,6 +444,9 @@ public:
   bool isShowDragBarsEnabled() const {
     return getBoolValue(showDragBars);
   }
+  QString getTimelineLayoutPreference() const {
+    return getStringValue(timelineLayoutPreference);
+  }
 
   // Animation  tab
   int getKeyframeType() const { return getIntValue(keyframeType); }
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -441,6 +441,9 @@ public:
   bool isShowFrameNumberWithLettersEnabled() const {
     return getBoolValue(showFrameNumberWithLetters);
   }
+  bool isShowDragBarsEnabled() const {
+    return getBoolValue(showDragBars);
+  }
 
   // Animation  tab
   int getKeyframeType() const { return getIntValue(keyframeType); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -148,6 +148,7 @@ enum PreferencesItemId {
   levelNameDisplayType,
   showFrameNumberWithLetters,
   showDragBars,
+  timelineLayoutPreference,
 
   //----------
   // Animation
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -147,6 +147,7 @@ enum PreferencesItemId {
   currentCellColor,
   levelNameDisplayType,
   showFrameNumberWithLetters,
+  showDragBars,
 
   //----------
   // Animation
--- a/toonz/sources/toonz/keyframemover.cpp
+++ b/toonz/sources/toonz/keyframemover.cpp
@@ -393,7 +393,7 @@ bool KeyframeMoverTool::canMove(const TPoint &pos) {
   if (usePos.x < 0) return false;
 
   int col      = usePos.x;
-  int startCol = getViewer()->xyToPosition(m_startPos).layer() - m_offset.y;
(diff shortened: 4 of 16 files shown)


beeheemooth:
I tried this and I want to say that it is super-mega-great! 
Once again this is the case when I don't want to go back to the old version! 
Thank you very much for your work! 
This feature alone is enough to release a new version!

So far I have not found any glitches and all column and frame behavior is logical.
 Regarding Ctrl+Shift+Click on selected frames: it would be nice if they were duplicated along with the keys (if they are selected).

beeheemooth:
If you select frames with keys in two columns, copying does not occur while holding Ctrl

Frames in multiple columns are also not duplicated if you accidentally select them using Ctrl+Drag

manongjohn:
> If you select frames with keys in two columns, copying does not occur while holding Ctrl
>
> Frames in multiple columns are also not duplicated if you accidentally select them using Ctrl+Drag

I think these are both existing behaviors in the current release unless I'm misunderstanding what you are doing. Videos also helpful 😉

beeheemooth:
ezgif-5-1e043ec5ef

beeheemooth:
ezgif-5-6ffb58cadd

artisteacher:
Love this so, so much!!!

The vertical alignment looks off on macOS. IMO, either the extra space of the dragbar should be removed entirely, or the text should be centered vertically. 
Screenshot 2024-01-03 at 8 09 55 PM

I don't really like that ctrl sometimes duplicates and sometimes selects keys and sometimes does both. Like I always expect ctrl+shift to select frames + keyframes - not select the keyframes but duplicate the frames...

Holding ctrl and dragging to start a selection of frames + keys, results in moving a copy of the first frame, instead of just making a selection including keys. I think it makes more sense to just make a selection of frames + keys instead.

I'd love a shortcut or modifier or action to select all implicit cells in a level or perhaps layer at once - the equivalent of clicking on the dragbar to select the entire level.

manongjohn:
Strange. I get the same results regardless if I use Alt+Ctrl-select  or Ctrl-select followed by Ctrl to move.  However, I do notice a difference depending on which column I click to do the move.  If I select the upper columns, it doesn't duplicate. If I select the bottom most column, it duplicates.

manongjohn:
@artisteacher 

I left the space for the locked column indicator which is why i didn't move the text and other things up.  I'll relook at raising them in this PR once I've ironed out any selection/move oddities introduced by this PR.

It crossed my mind to allow some combination of modifier keys + click on unselected cell to select all exposed frames in the clicked level. Question, is which combination of mod keys should do that?  This would work with and without drag bars showing.

The CTRL modifier does confuse things because of it's behavior of selecting keyframes.  Are you talking about existing behavior?  For now, I'm focused on making sure things behave similarly when drag bars are used or not used.

The various ways that the modifier keys impacts selections and movements is starting to drive me batty. lol

artisteacher:
I wonder if the select all frames could just be from double-clicking on the layer header. Ie. double-clicking on the text lets you rename while double-clicking off of the text selects all the implicit frames. 

I don't think that ctrl+shift duplicates anything when selecting with the dragbar.

beeheemooth:
Selecting frames along with keys by default is a great idea! It's more likely that you want to move or copy frames along with keys than that you want to move just cells (or keys). Perhaps holding down a modifier button to select may make sense when you want to move only frames or only keys.

manongjohn:
Problem with selecting keys is it limits moving to up/down the current columns so making that the default behavior would be problematic.

manongjohn:
> select all frames could just be from double-clicking

I kind of like this idea, but instead we use a modifier key  + double-click any filled cell in the column and it will auto-select all the exposed frames in the column.  Maybe a different mod key can be used to select connected frames, up to and including a stop frame.

manongjohn:
> I don't think that ctrl+shift duplicates anything when selecting with the dragbar.

It does. Apparently this logic has been there since the beginning.  It is a quick way to duplicate and insert cells in place.

I've decided against the Double-click suggestion for selecting all.  I don't know about anyone else, but I find double-clicking difficult to do with a pen and there are some issues with it because a normal click is always registered first and processed before a double-click is registered. This can cause unexpected results when certain modifiers are pressed in an active selection, like Shift+Ctrl (duplicates and inserts cells)

Instead I may provide a series of context menu options for Frame Selection options which users can then assign their own shortcut keys to. This will need to be in a separate PR.

@beeheemooth I believe I may have fixed the issue you were reporting.  Please let me know.

artisteacher:
https://github.com/tahoma2d/tahoma2d/assets/21159116/434c43c8-011f-47ba-a77d-7932b4883d93

1. Ctrl+click on the first frame
2. Still holding ctrl, also hold shift before clicking to select the last frame

This selects keys and duplicates frames at the same time.

beeheemooth:
What is happening at the moment in the program is amazing !!!

 All selections are copied and moved very conveniently and exactly as one would expect.

I'll continue testing. Need to check something else with the selection of frames.

artisteacher:
**Ctrl+click+drag to make a selection**

Starting where there's a keyframe:

https://github.com/tahoma2d/tahoma2d/assets/21159116/059c357e-fe17-492a-9b14-b3d92a65c871


Vs. Starting the selection without a keyframe:

https://github.com/tahoma2d/tahoma2d/assets/21159116/72440612-4c0f-4b0f-98c4-cef4175c0454

manongjohn:
I've switched the keyframe selection key to `ALT` for non-drag bar users only.  `CTRL` will now enable selection action instead.

For those using drag bars, `CTRL` will continue to select keyframes.

artisteacher:
The new minimal and compact options save so much vertical space! This will be amazing for small screens. Compact is a good default though.

manongjohn:
Let's make this official.  Thanks all for the testing and feedback!

Issue #1332, opened 2024-01-04
Potential crash from undo + autosave


artisteacher:
### Bug Description

A student was drawing in T2D today and experienced a crash after they used undo. Surprisingly, when they opened the scene in T2D again, it had just been saved & they didn't lose any work. Which makes me think this is related to auto-save. 

They did also notice the undo history was off - the last line drawn was not what was undone. 

### Steps to reproduce

I'm not sure how reproducible it is - unless there's another factor, it might just be a matter of auto-save running at the wrong time, or maybe auto-save getting interrupted...

### Relevant logs/screenshots

[undo-crash.txt](https://github.com/tahoma2d/tahoma2d/files/13824996/undo-crash.txt)

### Software Version

1.4

### Operating System

macOS

### Drawing Tablet

Wacom

manongjohn:
Based on the log, it appears to lead up to auto-save though I can't see the actual crash line.

Seems the person was drawing or something which halted auto-save.  The moment they stopped, the paused auto-save was triggered and somewhere it crashed in there.

I'm not sure exactly where the undo fits in unless the undo was immediately done after lifting the pen and while it was still auto-saving.  Conceivably, this could have changed something internally leading to the crash.  I may need to somehow introduce a delay in undo if autosaving is still in progress.

Quick review of the auto-save and auto-save suspend logic, there may also be a small potential of 2 autosaves overlapping, the normal timed one + a delayed  auto-save.  Perhaps this may be one reason for some corrupted saves or similar auto-save crashes. Will put some logic to guard around this possibility.

PR #1341, opened 2024-01-12
Fix actions that shouldn't happen while autosaving

manongjohn:
While reviewing #1332, I noticed there is a potential for overlapping save operations under certain scenarios which could potentially result in file corruption.  Additionally, #1332 suggests an undo operation was performed while an autosave was in progress.  This could cause a crash when it tries to access something that may no longer exist.

I've added logic to internally flag when a save is in progress so that any additional calls to save while actively saving will wait until the flag is cleared.

Scenarios that should be covered by this fix
- Timed autosave + suspended autosave due to active drawing or something
- Timed autosave + manual save by user
- Timed autosave + Undo/Redo

NOTE: It's hard to duplicate overlapping saving issues naturally.  I mostly regression tested for normal saving and autosaving operations as well as a simple saving error (file locked) and ensuring T2D didn't completely freeze up.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1585,6 +1585,11 @@ bool IoCmd::saveScene(const TFilePath &path, int flags) {
   CleanupParameters oldCP(*cp);
   cp->assign(&CleanupParameters::GlobalParameters);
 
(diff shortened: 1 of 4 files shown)


artisteacher:
Thanks for tackling this!

PR #1352, opened 2024-01-18
UHD capture camera + "save scene as" updates levels path

janimatic:
Hello,
this is my first Pull Request.
I guess it would be better to split in 2 PRs but i am new to it, forgive me, 'ill try to make seprate PR per commit in the future.

1 - UHD capture camera
A single line change adds UHD camera support for capture
It's a workaround because i guess the resolutions should not be hardcoded.

2 - "save scene as" updates levels path
this works fine, but i am open to suggestions as i am new to toonz code (I am exploring the existing code logic).
I'll remove the ifdefs once you review my code and give me your feedback.
In addition, "Save all" is called on "save scene as" to make sure no changes are lost.

Potential problems:
* using a search/replace logic with old / new scene name to create a correct level path could fail if the levels where not using scene subfolder already... I'd rather use the path formated when getUseSubScenePath option is on, but i couldn't find the correct method for that...

Thank you !
Happy 2d animating...

Julien

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -644,8 +644,15 @@ bool SaveSceneAsPopup::execute() {
   if (isSpaceString(QString::fromStdString(
           fp.getName())))  // Lol. Cmon! Really necessary?
     return false;
-
+#ifdef DEBUG_saveSceneAs
+  bool result = IoCmd::saveScene(fp, 0);
+  if (result)
+      return IoCmd::saveAll(0);
+  else
+    return false;
+#else
   return IoCmd::saveScene(fp, 0);
+#endif
 }
 
 void SaveSceneAsPopup::initFolder() {
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -644,8 +644,45 @@ bool SaveSceneAsPopup::execute() {
   if (isSpaceString(QString::fromStdString(
           fp.getName())))  // Lol. Cmon! Really necessary?
     return false;
+
+#ifdef FEATURE_copyLevelsOnSaveSceneAs
+  // copy and replace all levels if project option "Separate assets into scene sub-folders" is checked
+  TProjectP project = TProjectManager::instance()->getCurrentProject();
+  if (project->getUseSubScenePath()) {
+    ToonzScene *scene = TApp::instance()->getCurrentScene()->getScene();
+    TLevelSet *levelSet = scene->getLevelSet();
+    for (int i = 0; i < levelSet->getLevelCount(); i++) {
+      TXshLevel *lvl = levelSet->getLevel(i);
+      auto pathQString = lvl->getPath().getQString();
+      QString oldSceneName = QString::fromStdString(scene->getScenePath().getName());
+      QString newSceneName = QString::fromStdString(fp.getName());
+      std::cout << "SaveSceneAsPopup: oldSceneName: " 
+                << oldSceneName.toStdString()
+                << " newSceneName : " << newSceneName.toStdString()
+                << std::endl;
+      if (oldSceneName != newSceneName) {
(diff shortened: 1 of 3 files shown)


toonz/sources/toonz/filebrowserpopup.cpp
```diff
@@ -645,7 +645,51 @@ bool SaveSceneAsPopup::execute() {
...
+    }
+    // 3 - Notify xsheet scene and level of this change, of crash sooner or later ?
+    TApp::instance()->getCurrentXsheet()->notifyXsheetChanged();
+    TApp::instance()->getCurrentScene()->notifyCastChange();
+    TApp::instance()->getCurrentLevel()->notifyLevelChange();
+  }
+#endif
+#ifdef DEBUG_saveSceneAs
+  bool result = IoCmd::saveScene(fp, 0);
+  // automatic "save all" (otherwize, unsaved level changes would be lost)
```
manongjohn, line 685: I don't think forcing a `saveAll()` after a `saveScene()` is necessary. There might be reasons the user wishes to save just the scene file as a different name and not save level changes.

After saveScene(), unsaved levels are still flagged as dirty, so the user will be forced to either save all or abandon the changes.

janimatic: Hello John,
thanks for the suggestion.
What about a popup on "save scene as", to give user the choice to save levels ?
It happened to many many times getting red drawings, because in other softwares save scene saves everything.
Actually levels are also saved anyway if (project->getUseSubScenePath()), but level->setPath should work following your logic ? (dirty state for future user saveAll).
Tell me if i am supposed to write those changes (well i am not sure how to edit code in the same PR).
Thank you !

manongjohn: > What about a popup on "save scene as", to give user the choice to save levels ?

One of the reasons we moved `Save Scene` and `Save Scene As` into a sub-menu was to avoid confusion since we want people to use `Save All`.   I'd rather not introduce a pop-up for advance users who are more likely to be using this option instead of new users.

> Actually levels are also saved anyway if (project->getUseSubScenePath()), but level->setPath should work following your logic ? (dirty state for future user saveAll).

Yes, I saw you were saving it.  If setPath doesn't make the level dirty, you can always `setDirty(true)`

> Tell me if i am supposed to write those changes (well i am not sure how to edit code in the same PR).

Just update the branch you made your changes in and push the changes up and it will update the PR automatically.

Note: I noticed you used your `master` branch.  You should create a separate branch for changes and not touch your master.  This way, if you want to make other changes, they won't include stuff that either has not been officially merged yet since things can be merged in any order.

manongjohn:
> A single line change adds UHD camera support for capture
> It's a workaround because i guess the resolutions should not be hardcoded.

I'm not 100% sure, but I believe the list is hardcoded because there isn't a way to get a list of supported resolutions from the webcam.

janimatic:
> > A single line change adds UHD camera support for capture
> > It's a workaround because i guess the resolutions should not be hardcoded.
> 
> I'm not 100% sure, but I believe the list is hardcoded because there isn't a way to get a list of supported resolutions from the webcam.
I found a discussion about this problem :
https://forum.qt.io/topic/68904/how-to-get-the-supported-resolution-of-a-qcamera/5
We could later add direct support for decklink card (with their sdk) which support definition detection, but well using hardcoded resolutions is working...

janimatic:
I've created a new branch called FEATURE_copyLevelsOnSaveAs, with this plus a new commit :
In the last commit, i've added a new preference option in the saving category : Automatically Save Levels On 'Save Scene As
(it defaults to false to follow your suggestion, but i turn it on, to avoid red drawings on scene save as)
But then i am forced to create a new PR, no ?
Sorry for my limited nowledge of github...

manongjohn:
> But then i am forced to create a new PR, no ?
> Sorry for my limited nowledge of github...

Yes. Cancel this one and create a new PR using your new branch.

janimatic:
closing this  : opening a new PR on a separate branch here https://github.com/tahoma2d/tahoma2d/pull/1355

PR #1355, opened 2024-01-19
Feature copy levels on save scene as

janimatic:
* "save scene as" updates levels path
In addition, "Save all" is called on "save scene as" to make sure no changes are lost.

Potential problems:
using a search/replace logic with old / new scene name to create a correct level path could fail if the levels where not using scene subfolder already... I'd rather use the path formated when getUseSubScenePath option is on, but i couldn't find the correct method for that...

* - preference 
Saving/Automatically Save Levels On 'Save Scene As'
(it defaults to false, but i turn it on, to avoid red drawings on scene save as)
There might be reasons the user wishes to save just the scene file as a different name and not save level changes.

* - UHD capture camera
A single line change adds UHD camera support for capture
It's a workaround because i guess the resolutions should not be hardcoded.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -163,6 +163,7 @@ public:
     return getBoolValue(rasterOptimizedMemory);
   }
   bool isAutosaveEnabled() const { return getBoolValue(autosaveEnabled); }
+  bool isSaveLevelsOnSaveSceneEnabled() const { return getBoolValue(saveLevelsOnSaveSceneEnabled); }
   int getAutosavePeriod() const {
     return getIntValue(autosavePeriod);
   }  // minutes
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -75,6 +75,7 @@ enum PreferencesItemId {
   defaultProjectPath,
   recordFileHistory,
   recordAsUsername,
+  saveLevelsOnSaveSceneEnabled,
 
   //----------
   // Import / Export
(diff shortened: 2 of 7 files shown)


manongjohn:
To continue the discussion about Saving Levels with Save Scene As....

So when you Save Scene As, without immediately saving levels, it saves the scene and switches to it, but leaves levels in a dirty state.  So when you save via Save All later, the levels are still missing after reloading?  

Sounds like something internally didn't change somewhere when updating the path so it's still saving to the old location.

janimatic:
hello John,
as far as i can understand it,
when the preference "save level on scene scene as" is disabled,
level->setPath implementation is the problem :
* If i use keepFrame == true, the "save scene as" works, but when reloading the scene only modified drawings are valid.
setDirtyFlag doesn't force to write untouched frames when doing saveAll later, and even looping frameids to do level->touchFrame() like this :
```
          std::vector<TFrameId> ids;
          lvl->getFids(ids);
          for (auto id : ids) {
            lvl->getSimpleLevel()->touchFrame(id);
          }
```
doesn't force a future write of all frames on saveAll. (that's strange indeed) :
Only the newly painted frames will be saved.
* using keepFrame == false (default) obviously reloads the level from disk, so level should be writen in a new folder before.
I guess the current implementation of setPath forces us to keep real levels paths in sync with the scene.
I guess we could either  :
* rewrite setPath logic using the member variable m_path
* or always save levels on save scene as, when project->getUseSubScenePath() is true

best regards!
Julien

janimatic:
ok i got it working...
i looked into the code from TXshSimpleLevel::save & TXshSimpleLevel::saveSimpleLevel
and realised the old levels should be copied in the new directory if isSaveLevelsOnSaveSceneEnabled is false...
Because unmodified frames are read from disk on save...
I am using TXshSimpleLevel::copyFiles

I am reverted merge branch master because the master has a bug that prevent me from animating... (undo of brush strokes is broken)

Thanks for your advise, we are almost there (i still need to do some clean up)

manongjohn:
I wont be able to merge this because there are conflicts between your branch and master that must be resolved.

PR #1359, opened 2024-01-21
saveLevelsOnSaveAs + UHD capture cam + clean debug asserts

janimatic:
* preference Saving/Automatically Save Levels On 'Save Scene As' (it defaults to false : There might be reasons the user wishes to save just the scene file as a different name and not save level changes).

* "save scene as" updates levels path when project settings uses "separate assets into scene subfolders" option
- if preference Saving/Automatically Save Levels On 'Save Scene As' is checked, "Save all" is called on "save scene as"
- otherwise, the user must use SaveAll to write levels changes (prompt before exit).

Potential problems / todo :
using a search/replace logic with old / new scene name to create a correct level path could fail if the levels where not using scene subfolder already... I'd rather use the path formated when getUseSubScenePath option is on, but i couldn't find the correct method for that...

* UHD capture camera A single line change adds UHD camera support for capture It's a workaround because i guess the resolutions should not be hardcoded.

* removed debug asserts

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -2223,6 +2223,163 @@ bool IoCmd::loadScene() {
   }
 }
 //===========================================================================
+// IoCmd::saveSceneVersion()
+//---------------------------------------------------------------------------
+
+
+bool IoCmd::saveSceneAs(const TFilePath &fp) {
+  // copy and replace all levels if project option "Separate assets into scene
+  // sub-folders" is checked
+  TProjectP project = TProjectManager::instance()->getCurrentProject();
+  bool saveTwice    = false;
+  if (project->getUseSubScenePath()) {
+    // save scene a first time before updating levels paths just to make
+    // getDefaultLevelPath method happy (and avoid serach repalace old/new scene
+    // name)...
+    saveTwice   = true;
+    bool result = IoCmd::saveScene(fp, 0);
+
+    ToonzScene *scene   = TApp::instance()->getCurrentScene()->getScene();
+    TLevelSet *levelSet = scene->getLevelSet();
+    for (int i = 0; i < levelSet->getLevelCount(); i++) {
+      TXshLevel *lvl = levelSet->getLevel(i);
+      auto oldPath   = lvl->getPath();
+      auto newPath = scene->getDefaultLevelPath(lvl->getType(), lvl->getName());
+      if (oldPath != newPath) {
+        std::cout << "SaveSceneAsPopup: level oldPath: "
+                  << oldPath.getQString().toStdString()
+                  << " newPath : " << newPath.getQString().toStdString()
+                  << std::endl;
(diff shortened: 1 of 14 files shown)


janimatic:
hey John,
thank you for guiding me through github.
As mentioned, I'd like to enhance one thing in save scene as feature (a more robust path formating when getUseSubScenePath option is on, instead of search/replace, i need to look deeper into toonz code). And for the lazy graphist like me, i suggest to add a simple "save version" menu entry (just incrementing scene number suffix).

janimatic:
Hello,
ok i commited the suggested changes (more robust level path formating, IoCmd::saveSceneVersion() + IoCmd::saveSceneAs(...) + MenuFileAction "Save Scene version").
For me the pr is ok now...
notes :
- I couldn't understand how to use existing toonz numbered file naming methods so I wrote my own
- I've put some Utility static methods for file naming in iocommand implementation, because statics in tfilepath header would need a long rebuild for each code change... Maybe you will suggest a more central place.

thanks, and let's not forget animating !

toonz/sources/toonz/iocommand.cpp
```diff
@@ -2223,6 +2223,163 @@ bool IoCmd::loadScene() {
...
+    bool result = IoCmd::saveScene(fp, 0);
+
+    ToonzScene *scene   = TApp::instance()->getCurrentScene()->getScene();
+    TLevelSet *levelSet = scene->getLevelSet();
+    for (int i = 0; i < levelSet->getLevelCount(); i++) {
+      TXshLevel *lvl = levelSet->getLevel(i);
+      auto oldPath   = lvl->getPath();
+      auto newPath = scene->getDefaultLevelPath(lvl->getType(), lvl->getName());
+      if (oldPath != newPath) {
+        std::cout << "SaveSceneAsPopup: level oldPath: "
```
manongjohn, line 2249: Please remove these `std::cout` in your code.  These will never be seen unless running command line and doesn't really add any value, except maybe for debugging purposes.

janimatic: hey John,
that's done,
Is message center often used ?
thanks

manongjohn: > Is message center often used ?

Not really. I think it might be used more so by the Render Farm. Havent had the need to use Message Center for any particular reason.

manongjohn:
I ran into an issue with `Save Scene Version` when used on a scene that was not previously saved.  It caused it to save in the cache directory rather than prompting for a location to save it.  Any additional save after that throws an error.

Steps to reproduce:
1. Start a `New Scene` with 1 level
2. Click on `Save Scene Version`
    - Scene saves as `untitled###` in the cache folder.
    - It should have promoted for new location.
4. Make another change or just `Save All` or any other save option and you get the following message

image

janimatic:
@manongjohn 
thanks for testing!
that should be fixed now...

manongjohn:
@janimatic LGTM.

Thank you for making these changes.

PR #1364, opened 2024-01-23
Shift and Trace - Hide original position

manongjohn:
When using Shift and Trace, the original position is now only shown when you `Edit Shift`.

For those that still want to see the original position, a new `View` -> `Show Shift Origin` option is available.

image


artisteacher:
So if I exit "Edit Shift" by pressing B for the brush (Edit Shift gets deselected in the menu), I still see the shift origin. If I exit "Edit Shift" by deselecting it, then the shift origin hides as expected.

manongjohn:
Seems behind the scenes it thinks it's still editing the shift, even though the settings has been unchecked.  Will fix.

PR #1371, opened 2024-01-27
[WIP] OCA import

janimatic:
OCA import aims to achieve the i promised by ocaio....
Currently, importing external raster images is unclear to me. 
There is no simple importLevel command.
I m try to refer to code found in :
LevelCreatePopup pasteRasterImageInCellWithoutUndo LoadLevelPopup TImageReader::load() ResourceImportDialog...
Can you please give me some guidance to acheive the importOcaFrame method ?
Thanks !


janimatic:
ps: to clarify my question
everything seemp to work as expected so far,
but 
`simpleLevel->setFrame(fid, img);`
doesn't actually add the raster pictures to the xsheet level.

I guess I should manipulate the xsheet cells, but how ?

manongjohn:
If you have everything loaded internally, as parting of  loading the level, you'll need to create a new column and add cells into the xsheet.

See what `LoadLevelPopup` class found in `filebrowser.cpp` is doing?

If you would like to discuss in more real time, feel free to join T2D's Discord.

janimatic:
thanks,
i'll talk to you about it asap !

manongjohn:
Is this still work in progress?  If so, might want to either make it a Draft or add [WIP] to the title until it is ready.

janimatic:
After 1h45mn, the macos build seems to be stuck on OpenCv's IlmImf.
I am not sure if i can do something (it's certainly not related to this code)...

manongjohn:
> After 1h45mn, the macos build seems to be stuck on OpenCv's IlmImf. I am not sure if i can do something (it's certainly not related to this code)...

Its some oddity with macOS builds.  Sometimes it gets stuck there, other times it doesn't. Not sure why it is doing that.  Usually have to cancel that build and re-run it.

janimatic:
should i commit an blank change to force a rebuild ? i don't seem to have the rights to rerun it from github

manongjohn:
> should i commit an blank change to force a rebuild ? i don't seem to have the rights to rerun it from github

No. I will rerun it.

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,379 @@ void ExportOCACommand::execute() {
...
+  // see LoadLevelPopup, TImageP TImageReader::load(), ResourceImportDialog
+  TFrameId fid(jsonFrame["frameNumber"].toInt());
+  // sl->setFrame(fid, TImageP());
+  auto path       = TFilePath(m_parentDir.getQString() + "/" +
+                        jsonFrame["fileName"].toString());
+  bool fileExists = TFileStatus(path).doesExist();
+  TImageP img;
+  if (fileExists) {
+    auto imageReader = TImageReaderP(path);
+    img = imageReader->load();
```
manongjohn, line 892: May need to put this in a try/catch block since it's possible the load may fail. Additionally should verify `img` != 0 in the next line.

In my quick test, I OCA exported my test scene from a different T2D instance and then tried to Import it into this PR artifact and crashed on the next line because it failed to load one of my images.

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,379 @@ void ExportOCACommand::execute() {
...
+    // IoCmd::exposeLevel( sl, 0, m_xsheet->getColumnCount(), fids);
+    // m_xsheet->exposeLevel(0, m_xsheet->getFirstFreeColumnIndex(), sl, fids, true);
+    // I think this method should be fixed to use the fids for destination time in the column
+    // (eventually adding an optional bool argument timingUseFids,
+    // but i don't get why we pass fids if we're not using those ??)...
+
+    // that's the very simple solution...
+    int emptyColumnIndex = m_xsheet->getFirstFreeColumnIndex();
+    for (auto fid : fids)
+      m_xsheet->setCell(fid.getNumber(), emptyColumnIndex, TXshCell(sl, fid));
```
manongjohn, line 796: Fid has 2 meanings which will change if you are in Level Edit mode (Level Strip) or Scene Edit mode (Xsheet)

In Level Edit mode, the Fid is the frame number as it appears in the level strip.  This usually begins with 1.  It's rare that it starts with 0.

In Scene mode, the Fid is the Scene Frame position, which starts at 0 (internally).  For placement in the xsheet, you need to do `fid.getNumber() - 1`.  If Frame 1 is the beginning of the animation, then internally the 1st frame is 0.  No doing this would cause the animation to start on Scene Frame 2.  The only time this may be a problem is if the OCA data somehow has `"frameNumber": 0` since there isn't a -1 Scene Frame and will crash when trying to load

toonz/sources/toonz/ocaio.cpp
manongjohn: I believe, the reason why `img` is 0 here is because my level had blank frames before the 1st drawing.  The OCA Export created an entry for the blank frame but the Import doesn't handle empty frames.

```    "layers": [
        {
            "animated": true,
            "blendingMode": "normal",
            "childLayers": [
            ],
            "fileType": "png",
            "frames": [
                {
                    "duration": 1,
                    "fileName": "",
                    "frameNumber": 1,
                    "height": 0,
                    "name": "_blank",
                    "opacity": 1,
                    "position": [
                        960,
                        540
                    ],
                    "width": 0
                },
```

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,379 @@ void ExportOCACommand::execute() {
...
+
+    // column properties
+    auto column = m_xsheet->getColumn(emptyColumnIndex);
```
manongjohn, line 799: Request: Please set the column's name to the jsonLayer["name"], if populated.  This way the column name matches the 1st level in it, similar to how columns are auto-named upon creation.

This is usually done through the TStageObject:
```
    TStageObject *obj = xsh->getStageObject(TStageObjectId::ColumnId(m_col));
    std::string str   = m_columnName.toStdString();
    obj->setName(str);
```

toonz/sources/toonz/ocaio.cpp
janimatic: @manongjohn 
That's strange, doesExists (QFileInfo.exists()) should have skip those empty frames...
Maybe we should check QFileInfo.exists() https://stackoverflow.com/questions/23930754/qfileinfo-exists-and-isfile-error
or use 
```
#include <sstream>
#include <iomanip>
static bool fileExists(std::string fileName) {
  std::ifstream infile(fileName.c_str());
  return infile.good();
}
```
found in my previous pr [iocommand.cpp](https://github.com/tahoma2d/tahoma2d/pull/1359/files#diff-c22e50ed5b65b43cdb2bd77320dae1eeb8526a5ae38420eeb3e0a900d9c3fd51) line 2284
Anyway that should be fixed...

toonz/sources/toonz/ocaio.cpp
janimatic: I also am offseting any negative frame range now...

janimatic:
* by the way i don't have my previous PR. What is the correct way to update my fork without creating a mess, as i am working on multiple PRs ? ?

manongjohn:
> * by the way i don't have my previous PR. What is the correct way to update my fork without creating a mess, as i am working on multiple PRs ? ?

Normally i switch to my master, fetch/pull tahoma2d master into it and then push to sync it remotely.

Then i go to whatever branch i need to and either rebase against origin/master or pull origin/master into it.  I only usually on update the branch i am working on if there is a conflict noted in the PR or i saw something new merged into master that i may need to update.

toonz/sources/toonz/ocaio.cpp
janimatic: i've just tested, and QFileInfo correctly returns false with an empty string argument.. It's strange that you had that issue with empty frames, in previous version, but well...

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,448 @@ void ExportOCACommand::execute() {
...
+                " height: " + QString::number(info->m_ly) 
+                + " bits: " + QString::number(info->m_bitsPerSample) );
+
+  } else {
+    DVGui::error(QObject::tr("importOcaFrame file not found! : ") + path.getQString());
+    img = sl->createEmptyFrame();
+  }
+  sl->setFrame(fid, img);
+  if (!Preferences::instance()->isImplicitHoldEnabled()) {
+      //do something with the image duration ?
```
manongjohn, line 979: This doesn't need to be done here.  Image duration needs to be handled only when loading into the xsheet

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,448 @@ void ExportOCACommand::execute() {
...
+/// <param name="jsonLayer"></param>
+
+void OCAIo::OCAInputData::importOcaFrame(const QJsonObject &jsonFrame,
+                                         TXshSimpleLevel *sl) {
+  // doesExists (QFileInfo.exists()) should skip empty frames ! but...
+  TFrameId fid(jsonFrame["frameNumber"].toInt());
+  if (jsonFrame["name"].toString() == "_blank" ||
+      jsonFrame["fileName"].toString() == "") {
+    TImageP img = sl->createEmptyFrame();
+    sl->setFrame(fid, img);
```
manongjohn, line 943: I think this is incorrect.  We should be skipping "_blank" frames completely not creating empty image frames.

Based on what I see generated from the OCR Export, the "_blank" frames are actually "empty" cells with no image.  Here you are creating a frame with an empty image.

Blank frames, in T2D are represented by a specific FID (-1, I think) without a level.  I am not at my computer to check, but "blank" frames are not added to the level so we should not be creating these.

toonz/sources/toonz/ocaio.cpp
```diff
@@ -534,4 +535,448 @@ void ExportOCACommand::execute() {
...
+      if (fid.getNumber() <= 0) {
+        DVGui::warning(
+            QObject::tr("importOcaLayer : skipping frame -1! ") +
+            jsonLayer["name"].toString());
+        continue;
+      }
+      // -1 converts framenumber to index, frameOffset moves the frame range above 0...
+      m_xsheet->setCell(fid.getNumber() - 1 + frameOffset, emptyColumnIndex,
+                        TXshCell(sl, fid));
+    }
```
manongjohn, line 841: the `jsonLayer["duration"]` should be applied here, only if duration > 1 and  `!Preferences::instance()->isImplicitHoldEnabled()` (explicit holds mode).  Essentially copy the same `TXshCell(sl, fid)` into the next few cells based on the duration value.

manongjohn:
> * I cannot test child levels as i never used it, can you give me some guidance please ?

In T2D, a childLevel is actually a sub-scene/xsheet.  Childlevels can be imbedded within other levels.  We have a special level for these.  Look at how the column command `Collapse` works.  This command creates childLevels and adds levels into it.

janimatic:
@manongjohn 
i've added a Qt class in ocaio 

```
class LoadOCAPopup final : public FileBrowserPopup {
  Q_OBJECT

  QList<QLabel *> layerNameLabelList;
  QList<QComboBox *> levelMergeStatusList;
  OCAIo::OCAInputData *m_data;

public:
  LoadOCAPopup(OCAIo::OCAInputData *data);
  bool execute() override;
  bool executeApply() override;
  bool doLoad();

   public slots:
    void onEditingFinished();
};

```
But i don't know how to run moc in visual studio 2019 (in vs2017 i could right clic on a header and run moc : i had a Qt vs addin... but i was surpised everythink worked without that horrible addin till now...)

ocaio.obj : error LNK2001: symbole externe non résolu "public: virtual struct QMetaObject const * __cdecl LoadOCAPopup::metaObject(void)const " (?metaObject@LoadOCAPopup@@UEBAPEBUQMetaObject@@XZ)

Any idea how to fix this ?
thanks !

janimatic:
i guess everything went fine till now because the project was generated by cmake...
https://developercommunity.visualstudio.com/t/qt-moc-starts-over/580465
Then, maybe is there a way to rerun cmake to regenerate the vc project (and have moc custom build rules set in it) with breaking everything ?

janimatic:
I re-runed cmake and it didn't fix it... 
I've added the class in the same files (ocaio.h/.cpp)
i will try to look in the CMakeList.txt

manongjohn:
If you added to existing files, then its not a cmake issue as far as I know.

Fyi, i've never run moc before so not sure what's going on unless i tried to build it myself

janimatic:
ok that's fixed...
i just needed to edit toonz/sources/toonz/CMakeLists.txt 
and add ocaio.h in MOC_HEADERS 
Indeed a simple rebuild works...
Thank you !

janimatic:
hello John,
i've just commited the new ui code with selective import.
Your usefull suggestions remains to be implemented (see commit description).

janimatic:
I have disable children layers support, but the rest should be fine...

manongjohn:
I'm ok with not handling `childLayers` right now, actually.  Based on what I just read about them in OCR specs, the `groupLayers` are akin to folders which we currently don't have.  Not sure we should be converting them into sub-scenes since there are no frames in a groupLayer to expose.

I'm actually working on timeline/xsheet Folders for T2D right now so we can implement childLayers later.   However, I do notice that T2D/OT's OCR Export treats Sub-Scenes/Xsheets as `groupLayers`, which I suppose it accurate outside of our world.  This is a bit of a problem because the sub-scenes have frame information that is lost. Granted, most other software likely do not have a similar concept of a timeline within another timeline.

Not expecting this now, but may be nice to also get `vectorLayer` import and exports working at some point.  We have SVG - PLI conversion methods.

manongjohn:
Some additional thoughts regarding importing childLayer:

For the time being, it might be nice to load the childLayers as regular layers and ignore the grouping information so the user can still load those layers into the scene.  The user can then manually group/collapse them if they want.  A warning should be given that the group is being ignored but that layers are still being imported.  Can that be done?

manongjohn:
This is my test OCR Export from T2D that I'm trying to reimport: [text_explicit.zip](https://github.com/tahoma2d/tahoma2d/files/14129281/text_explicit.zip)


Here is what my test scene looks like before OCR Export when using Explicit Holds Mode:
image

Few issues I noticed importing the test export:
1. It's still creating image frames for "_blanks" (see red frames below).  To set an Empty cell in the xsheet, you need to use `TXshCell(0, TFrameId::EMPTY_FRAME)`
2. At one point during the loading process the progress bar goes backwards
3. If I import the OCA data successfully, chose New Scene without saving or Load an existing scene and then try to reimport OCA file again, I get a crash.  Crash reason seems to vary.

Here is what happens when I try to reimport the test export using Explicit Holds mode:
image

janimatic:
Thanks for testing John,
First, i am trying to sumuraize your suggestion. Tell me if it's ok for you :

** Todo Now ** 
* import child layers as normal layers, but popup message at the end of the import to notify the user that the layer's groups have been ignored.
* ocaImportPopup  (selective import ui) : fix the case of no filename or folder filename name in your test/screenshot (Attempting to load my OCA exported test scene, this UI appears bit buggy and cumbersome:). 
* I personnally find very usefull to access this selective import ui , instead of having many popup messages. But i find the vertical slider (for antialiasing) strange. Maybe that should be a simple int field.
* ocaImportPopup  : fold the import option by default so you don't see that big interface (but see if that's possible/simple with Qt : with a QToolBox ? https://stackoverflow.com/questions/32476006/how-to-make-an-expandable-collapsable-section-widget-in-qt)
* Insert stop frames after the last frame + its duration?
* create empty frames with TXshCell(0, TFrameId::EMPTY_FRAME) (not white frame, not error/red frame)
* progress bar goes backwards at some point : the progress bar just counts layers, maybe it should be more granular, used in import frames loop ?
* fix crash case : _If I import the OCA data successfully, chose New Scene without saving or Load an existing scene and then try to reimport OCA file again, I get a crash. Crash reason seems to vary._ I will try to reprocuce in debug mode...

** Todo Later ?** 
* Import/Export svg at some point / convert vector to svg if necessary (PLI ?)
* OCA doesn't support frame tags (NavigationTags) this would be very usefull to mark keys / breakdowns / inbetweens / etc
drawing ... for the onion skining and fliping. Use OCA metadata to transfert frame tags ?

** my own todo later list : 
- add "replace" to mergingStatus
replace would avoid duplicate (but load would still be usefull to compare
and make changes manually) : the level name should be incremented ?

** Cancelled / cumbersome
- cancel "merge" merge would add only missing frames ?

** for other PRs/tools
+ for another PR/tool : gltf import/export
- keyframe animation and transformations transfert for camera, pegs, etc...
that would be another tool and another open file format gltf (can be json or binary).

+ for another PR/tool : gltf import/export
- blendingMode (implement blendingMode in viewport with fbo pingpong ? + ui
ComboBoxes on column headers ?)

manongjohn:
Before I continue, I just want to say "Thank you!" for working on this. :)


> ### Todo Now
> * import child layers as normal layers, but popup message at the end of the import to notify the user that the layer's groups have been ignored.

Yes Please.

> * ocaImportPopup  (selective import ui) : fix the case of no filename or folder filename name in your test/screenshot (Attempting to load my OCA exported test scene, this UI appears bit buggy and cumbersome:).

Yes

> * I personnally find very usefull to access this selective import ui , instead of having many popup messages. But i find the vertical slider (for antialiasing) strange. Maybe that should be a simple int field. The idea of the ocaImportPopup  first came from the case where the level with same name (raster) could replace a level with palette info (smart raster) causing essential info loss.

I'm not sure why a Raster level would be overwriting a Smart Raster level much less it's palette.  Regular Raster and Smart Raster levels are 2 different level types with different file formats and extensions.  When looking for existing levels, you can safely ignore any files with a .TLP and .TLV extension as these are specific to Smart Rasters which OCR Import would never be dealing with.  Additionally, it is probably also wise to see what levels are loaded into the scene that have yet to be saved and look for Raster (ignore all other level types) level names already being used

> * ocaImportPopup  : fold the import option by default so you don't see that big interface (but see if that's possible/simple with Qt : with a QToolBox ? https://stackoverflow.com/questions/32476006/how-to-make-an-expandable-collapsable-section-widget-in-qt https://copyprogramming.com/howto/how-to-make-a-nestable-expandable-widget-in-qt )

My suggestion is to not even bother with that import option and just load the layers in.  You will need to handle for duplicate level names if you are importing into a project.  Keep in mind that Raster levels are usually loaded into the `+extras` folder which is shared by ANY scene in the project so even though the scene may not have a particular level name loaded, that doesn't mean it doesn't already exist.  In this case, I would suggest just forcing a new level name and make sure all images for that level have the same name.

> * Insert stop frames after the last frame + its duration?

Stop frames (`TXshCell(sl, TFrameId::STOP_FRAME)`) are only really needed for scenes in Implicit Hold mode. After loading all the frames for the layer you can place a stop frame after the end of the last frame.  Stop Frames should not be used for scene in Explicit Holds mode.
 
> * create empty frames with TXshCell(0, TFrameId::EMPTY_FRAME) (not white frame, not error/red frame)

Yes.  All "_blanks" should just be populating the xsheet/timeline column with this cell information.

> * progress bar goes backwards at some point : the progress bar just counts layers, maybe it should be more granular, used in import frames loop ?

The progress bar moves right as it starts reading, but as it's loading the levels into the scene, it starts moving left.  I don't know of any progress bar that moves like that.  It looks very odd.

> * fix crash case : _If I import the OCA data successfully, chose New Scene without saving or Load an existing scene and then try to reimport OCA file again, I get a crash. Crash reason seems to vary._ I will try to reprocuce in debug mode...

Yes.

> Todo Later ?
> Import/Export svg at some point / convert vector to svg if necessary (PLI ?)

I'm not expecting this with this PR.  This can definitely done as an enhancement later.

> OCA doesn't support frame tags (NavigationTags) this would be very usefull to mark keys / breakdowns / inbetweens / etc
drawing ... for the onion skining and fliping. Use OCA metadata to transfert frame tags ?

NavigationTags are per Scene frame, not per level.  These would not be good for marking keys/inbetweens/etc.  If you want to mark keys/breakdowns/inbetweens, then you want to use Cell Marks instead. Cell Marks have no navigation functionality and are a rather limited feature at the moment.


Some of this stuff is probably best explained in a more real-time conversation. I strongly suggest joining the Discord to discuss things in more details and to help clarify stuff.  I'm not sure where you are in the world, but even if I'm not available, there are still a number of users in the Discord who I'm sure can help explain certain T2D concepts.

janimatic:
Hello John,
I hope you're doing well.
hum...
Now it looks fine, i can save the scene, but if later i do SaveAll i get a crash...
Images files are copied in /extras unnumbered (A..tif instead of A.0001.png,A.0002.png) (so i get only one image from the image sequence in /extras)
Stack trace :
SaveAll>SceneResources::save>SceneLevel::save()
where 
actualFp is {m_path=L"P:\\projets\\recherches\\projet\\docs\\tahoma2d\\exercices2d\\extras\\testImportOCA011\\psd-flame-icon..tif" }
and oldActualPath is the same (looks like only psd case is implemented to copy files)
But the case for tif (i guess my pref is set to convert extras to tif)  is implemented a couple a lines later :it saves with no frame in filename
```
        if (actualFp.getType() == "tif" || actualFp.getType() == "tiff" ||
            actualFp.getType() == "tga" || actualFp.getType() == "tzi") {
          TFilePath clnin = oldActualPath.withNoFrame().withType("cln");
          if (TSystem::doesExistFileOrLevel(clnin))
            TSystem::copyFile(actualFp.withNoFrame().withType("cln"), clnin);
        }

```
then a lot of toonzlib.dll!saveBackup(TFilePath path) calls
and the  TSystem::readDirectory with variables that looks bad...

I am not sure if the problem is in my code, thus.
I guess toonz prefers to copy always in extras (when using load level), but i find it really annoying.

A load level test outside of OCA context :
Create a new scene > load a numbered png sequence level in A > reply "import" to the question asked > save scene as 
TXshSimpleLevel::copyFiles tries to copy from a non existing folder and crashes : 
dst: {m_path=L"P:\\projets\\recherches\\projet\\docs\\tahoma2d\\exercices2d\\extras\\testLoadLevel002\\A..png" }
src: {m_path=L"P:\\projets\\recherches\\projet\\docs\\tahoma2d\\exercices2d\\extras\\untitled171\\A..png" }
the real path should be P:\projets\recherches\projet\medias\incoming\text_explicit\test_explicit_oca\A..png

I must admit it might be caused by my other PR with save scene as...
But the same problem arise with SaveAll (in place of save scene as), which i didn't change.
IoCmd::saveAll>SceneResources::save>SceneLevel::save() (same method in the end...)

Thank you !

manongjohn:
I will need to look into it in more detail when I can.

Normally, upon loading a level/file, a user is prompted whether to `Load` the files from their original location or `Import` it into the current project.  As far as I know, imported files are always copied as is into the `+extras` folder for the project.  We should be doing the same.

For raster based file, if there are multiple image files for the same level, they must have the frame number as part of the filename, regardless if its png, tiff or any raster based extension.  Saving as A..xxx is incorrect.

I argue that all raster images should be saved with a frame number if one cannot be extracted from the file name.  Images loaded without a frame number are single image levels. If the user needs to add more images to the level, not having a frame number will prevent the user from adding more images.

manongjohn:
Importing vs loading OCR image files.

As i think about this more, i guess the question is, shouldn't the filenames be preserved from the OCR export whenever possible?. Unless its an issue of an already existing filename, why are you changing the filename at all? Honestly not even sure current Import detects existing filenames and prompts user for action. May just overwrite it, regardless. Need to check.

janimatic:
Hey John,
I would not change the external images filename at all, but i faced issues where toonz would set the level path to /extras by itself.
I'd definately prefer to keep image sequences external, where they are, (unless those are created inside toonz).
Sorry, that's a lot of informations to deal with at the same time.
I'll try to fix the other issue https://github.com/tahoma2d/tahoma2d/issues/1384 when i have the time, but that's a different topic which i need to I need to clarify first...

manongjohn:
> i faced issues where toonz would set the level path to /extras by itself

That's pretty normal for creating levels while working in T2D.  You should review the logic for Load Level/Scene to see where it prompts for Load or Import and how it uses the reponse. We should provide the same options for all ocr files being imported.

janimatic:
Not particularly proud of this but :
* added m_originalPath to TXshSimpleLevel
* setOriginalPath when creating level from external files (importOcaLayer)
* getOriginalPath and use it if not empty, instead of m_path (IoCmd::saveSceneAs)
* COMMENT OUT an essential call saveBackup(dDstPath); in TXshSimpleLevel::save
Otherwise save all was crashing...
Now the save methods works, levels images doesn't turn white on save...
But when reopening the scene the folder contains stupid frames named with 2 dots (like "pikachu..tif") which prevent the level from loading correctly.
They must have been copied at some step by the file management octopuss ?
If i remove those "***..tif" from disk, the loading of level works (valid images are correctly copied). 
and obvisouly everything else...
I hope people will give their opinions to the re-design choices here:
https://github.com/opentoonz/opentoonz/issues/5286 marked as a question while it's a critical bug in my opinion

manongjohn:
I think you are trying to do too much with this PR and some issues need to be handled separately.

I also think you may not be importing images quite right, leading to your saving issues. I say this since we load external images and save all the time, presumable without issue. That's not to say there arent a few bugs.

As soon as I can, I'll clone your branch and look in more detail to see what it is you are doing/not doing.

We'll discuss further in Discord. I'll create a thread for it.

janimatic:
fantastic, thanks for your support.
Indeed i started to play with master to get a better understanding of loading/saving and existing issues, I'll probably discard the last commit in the end but hopefully it gives an idea of where the problems happens, when we'll make a cleaner solution.
Also i suppose my preferences (scenepath + ask question on load + auto save levels on scen scene) doesn't make things easier...

janimatic:
How do you load external images ? Do you avoid the scenepath project option ? Do you prefer load or import ? Do you work only on already on saved scenes ?

manongjohn:
You can load external images via `Level` -> `Load Level`, or drag/drop file into viewer/timeline or xsheet which are always placed in the `+extras` folder. I've done both with new and existing scenes without issue. I'm fairly certain there are users using files created in other software, like Krita, without issue.

Loading images referenced in a OCA export file should be no different than when loading a file manually. The OCA export file tells you where to get the files. User should be prompted if they want to load it or import it.

I personally prefer to use `Import` instead of `Load` but I again have done both successfully.  

I would hazard a guess that most users are not using `$scenePath`. I believe that there are probably some issues when it is used, especialy if turned on in the middle of a project. That really needs to be looked at separately.

janimatic:
> I would hazard a guess that most users are not using `$scenePath`. I believe that there are probably some issues when it is used, especialy if turned on in the middle of a project. That really needs to be looked at separately.
Yes that must be the main point of my concern (I never tried Tahoma without "Project Settings>separate assets into scene sub-folders" option, since i thought it would just be a mess for level naming).
How do you practically name your levels to avoid names clashes if they all are stored in /drawings folder -if not using "Project Settings>separate assets into scene sub-folders"- ? Considering that default level names are A, B, etc...
Do you use "Project Settings>separate assets into scene sub-folders" and successfully import levels at all ?
I don't work in big studios anymore, i  am an indepednant filmaker,  but I like to organize my shots to spend less time with computer problems.
I can follow your advises, if that's not possible for the moment to use that project option. Just tell me : for now i just need clarification of normal uses cases.
I would then lock a first version of this feature in this PR, to the main Tahoma usage/preferences, for a first working version (trying to avoid any unmanageable case in the current state of the api).
Later, hopefully I'll find the energy to fix the scenepath problems and avoid/delay all this file management in a separate topic/PR.
In the end separating shots would need to work for me.
What do you think?

manongjohn:
T2D always generates those lettered filenames and doesn't duplicate within the same project, so that isn't an issue.  Even if you try to name a file manually as you are creating a new level, I believe it double checks and will prevent you from reusing an existing filename in the current project.

File name conflict are likely only an issue when Importing from other software or a different T2D/OT project.  I would, at the very least, provide the user the option to overwrite or rename the file.  This has broader impacts when importing a scene since now the scene also needs to be changed on import.  But that's another issue...

There might be valid reasons you want to Import and overwrite existing files.  Consider this.  You are working with another animator to do some stuff. You export and share the scene with this person so you can continue working on other parts of the animation. The other person does their stuff and sends the scene back to you.  In this case, you would want to import any new or updated files and replace the existing one.  Another case would be updating images in another software and then needing to update the files you loaded previously into T2D.  In this case, though, I'd probably be using Load instead of Import so the same file is easily accessible from a common place.

The primary goal of this PR is to be able to read and load the information in an OCA export.  Anything beyond that is likely outside the scope of this PR.

If you use existing level loading routines used by other parts of T2D then that shouldn't be much of an issue with loading and eventual saving.  With some of the issues you are currently having while saving, I think you're not quite loading image files right so I'll need to review your changes more thoroughly to see what's going on.

janimatic:
i switched "github desktop" to Repository settings>fork behavior>for my own purpose and pushed, but it still affects the original repo... sorry

manongjohn:
@janimatic I'm closing this since we'll be going with PR #1392, which contains your early commits + changes and corrections I made for loading image files.

Many thanks for getting this feature started.

PR #1385, opened 2024-02-03
Timeline/Xsheet Folders

manongjohn:
This new feature allows you to group columns into Folders on the timeline/xsheet.

image
image

General Folder information
- Folders are represented as columns and items in folders will follow immediately below it.
- Folders can be nested in other folders.
- Open/close folders by clicking on the triangle in the thumbnail area, using right-click menu option or shortcut (if defined)
  - For `Minimum` x-sheet layout, you can long press the folder to open/close 
- Xsheet folder items appear left of the folder column when open
- The stacking order of columns, in or out of folders, is still important from a rendering standpoint.  Folders have no impact to rendering.
- Toggling a folder's `Preview` off, `Camera Stand` off or, Lock` on  or setting `Opacity` < 100% or set a color `Filter` will override individual folder item settings.
  - When a folder's setting overrides folder item settings the folder item's settings cannot be changed.
  - Folder item's settings will be restored when the folder's settings are set back to normal.
- Folder cells can be Cell Marked.
- Other cells cannot be moved into folder cells.
- Scenes saved with folder information cannot be opened in prior versions of T2D (or OT for that matter).
- When a folder is selected for operations (i.e movement), all items in the folder are automatically selected.

Other changes:
- Movement of columns has changed.
  - Selected columns are no longer moved immediately.
  - An indicator between columns is shown where the selected column(s) can be moved to.
  - When non-contiguous columns are moved, they are moved to be next to each other.
- Can now resize the timeline's header by dragging the layer/cell divider left or right
-----
Creating folders:
1. Use the `New Folder` menu option in the `Level` -> `New`
2. Use the `New Folder` context menu option of column headers in the time timeline/xsheet
3. Select multiple columns and `Group` (Ctrl+G)
   - This will create a folder above the top/left most column.
   - All columns in the selection will be moved to be under the new folder column.

Adding items to folders:
1. Drag and drop columns directly onto a folder column
2. Drag and drop columns between 2 columns already in a folder
3. Create a new column between 2 columns already in a folder
4. Drag columns to the bottom of the last item in the folder. (indicator will change slightly)

Removing items from folders:
1. Drag and drop columns to any place outside the folder
   - If the last item in the folder is moved out, the folder will remain as an empty folder.
2. Select desired column(s) and `Ungroup` (Shift+Ctrl+G).
   - If the ungroup selection includes the folder, the folder will be removed as well
   - In case of nested groups, it will only ungroup the columns from the current folder, not all folders
   - If a specific column is removed from a folder, it will be shifted down below the folder it was in.

Nesting folders:
1. Create folders as above with at least 1 item already in the folder
2. Drag a folder column into another folder, similar to adding items to a folder

### DIFF
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2362,9 +2362,9 @@ XsheetViewer {
   qproperty-SelectedSoundColumnColor: #7692a1;
   qproperty-SoundColumnHlColor: #34FE5E;
   qproperty-SoundColumnTrackColor: rgba(0, 0, 0, 0.8);
-  qproperty-FolderColumnColor: #e7c9a9;
-  qproperty-FolderColumnBorderColor: #dcb081;
-  qproperty-SelectedFolderColumnColor: #fbcdbe;
+  qproperty-FolderColumnColor: #c4a972;
+  qproperty-FolderColumnBorderColor: #b5934e;
+  qproperty-SelectedFolderColumnColor: #dbaf8b;
   qproperty-ActiveCameraColor: #4073a3;
   qproperty-SelectedActiveCameraColor: #617db8;
   qproperty-OtherCameraColor: #5e9aa3;
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2362,6 +2362,9 @@ XsheetViewer {
   qproperty-SelectedSoundColumnColor: #7692a1;
   qproperty-SoundColumnHlColor: #34FE5E;
   qproperty-SoundColumnTrackColor: rgba(0, 0, 0, 0.8);
+  qproperty-FolderColumnColor: #e7c9a9;
+  qproperty-FolderColumnBorderColor: #dcb081;
+  qproperty-SelectedFolderColumnColor: #fbcdbe;
   qproperty-ActiveCameraColor: #4073a3;
   qproperty-SelectedActiveCameraColor: #617db8;
   qproperty-OtherCameraColor: #5e9aa3;
--- a/stuff/config/qss/Darker/Darker.qss
+++ b/stuff/config/qss/Darker/Darker.qss
@@ -2362,9 +2362,9 @@ XsheetViewer {
   qproperty-SelectedSoundColumnColor: #7692a1;
   qproperty-SoundColumnHlColor: #34FE5E;
   qproperty-SoundColumnTrackColor: rgba(0, 0, 0, 0.8);
-  qproperty-FolderColumnColor: #e7c9a9;
-  qproperty-FolderColumnBorderColor: #dcb081;
-  qproperty-SelectedFolderColumnColor: #fbcdbe;
+  qproperty-FolderColumnColor: #c4a972;
+  qproperty-FolderColumnBorderColor: #b5934e;
+  qproperty-SelectedFolderColumnColor: #dbaf8b;
   qproperty-ActiveCameraColor: #4073a3;
   qproperty-SelectedActiveCameraColor: #617db8;
   qproperty-OtherCameraColor: #5e9aa3;
--- a/stuff/config/qss/Darker/Darker.qss
+++ b/stuff/config/qss/Darker/Darker.qss
@@ -2362,6 +2362,9 @@ XsheetViewer {
   qproperty-SelectedSoundColumnColor: #7692a1;
   qproperty-SoundColumnHlColor: #34FE5E;
   qproperty-SoundColumnTrackColor: rgba(0, 0, 0, 0.8);
+  qproperty-FolderColumnColor: #e7c9a9;
+  qproperty-FolderColumnBorderColor: #dcb081;
+  qproperty-SelectedFolderColumnColor: #fbcdbe;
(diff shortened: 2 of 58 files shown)


janimatic:
great ! 👍 
(oca not ocr)

papio-pl:
Will the folders work with clipping mask? I mean if you put a mask layer before a folder will it affect everything this folder contains?

manongjohn:
> Will the folders work with clipping mask? I mean if you put a mask layer before a folder will it affect everything this folder contains?

Unfortunately no. You'd need to use sub-scenes to do that.

papio-pl:
> > Will the folders work with clipping mask? I mean if you put a mask layer before a folder will it affect everything this folder contains?
> 
> Unfortunately no. You'd need to use sub-scenes to do that.

Thank you. I thought so:) Anyway it's a great feature to keep things organized. 

Maybe in the future there will be possible to add workaround for the clippng mask which will just simply add hidden column/layer to every drawing layer in the folder.

artisteacher:
Really nice functionality! But maybe a different color?

manongjohn:
> Really nice functionality! But maybe a different color?

I went with a manila color since i think of th e old manila folders when i picture a folder. What color do you suggest?  Provide RGB or hex value if possible.

artisteacher:
I totally get the reference, just don't love how it plays with the other colors.

Could go with a grey - darker than a note level so that "utility" layers have related colors.
#7e7e7e
Screenshot 2024-02-05 at 7 51 59 PM

Could stick with a more muted manila-ish to better match the color intensity:
#C4A972
Screenshot 2024-02-05 at 7 59 38 PM


Could go a completely different direction with a yellow-green:
#799763
Screenshot 2024-02-05 at 7 55 42 PM

I can come up with more suggestions if you'd like.

manongjohn:
I want to avoid using a color similar to other colors we're already using. The grey reminds me of Note levels and the green, Smart Raster.

Muted manila might be the way to go, though I am biased at this point. 😁 After seeing it so often while working on this feature, i'm used to it being there so its a little harder for me to imagine something else. Additionally, manila seems to work ok, I think, with Light theme.

That being said, I'm still open to seeing other color options. Just needs to work with light and dark theme, or give me 2, one for dark theme and variation of it for the light theme.

artisteacher:
On a different note, in the minimum layout, there's some extra grey margin on the left of layer icons.
Screenshot 2024-02-05 at 8 46 48 PM

artisteacher:
> That being said, I'm still open to seeing other color options. Just needs to work with light and dark theme, or give me 2, one for dark theme and variation of it for the light theme.

Muted manila is fine. Though if you want another option, I don't think we have any orange levels. Sort of goes with the logo.
#CF774F
Screenshot 2024-02-05 at 8 50 22 PM
Could keep the current manila for the light theme - it looks fine there.

manongjohn:
> On a different note, in the minimum layout, there's some extra grey margin on the left of layer icons.

Odd, in my screen it looks correct.  I noticed that on a screenshoot posted in Discord.  I wonder what could be causing that.  Is it just the folder columns doing that or columns not in folders also doing that?

manongjohn:
Also, is it only doing that when xsheet is in Minimal mode? Is it fine otherwise?

artisteacher:
Looks like just the camera and folder layers. Layers outside the folder are fine.

Minimal Xsheet's visibility buttons have issues within the folder:
Screenshot 2024-02-05 at 9 42 00 PM

manongjohn:
Not sure what to do about folders in `Minimum` Xsheet layout.

Unlike the other xsheet layouts, I just compress the thumbnail to allow things to fit somewhat nicely, though with enough folder nesting it would result in a similar view. 

There isn't anything I can compress in minimum xsheet layout.  Can't really move the header separator.  Even if I moved the separator, what would I expand to make things not in folders, or stuff in different levels of folders line up?

artisteacher:
I guess the only thing that could be expanded is potentially the area with the column name.

manongjohn:
> On a different note, in the minimum layout, there's some extra grey margin on the left of layer icons. Screenshot 2024-02-05 at 8 46 48 PM

Can you verify the extra space is gone?

artisteacher:
Minimum timeline view is all set. Also the more muted color fits in so much better on the dark themes!

manongjohn:
> Minimum timeline view is all set. Also the more muted color fits in so much better on the dark themes!

Thanks for verifying and I agree, it does fit better.  Thanks for the suggestion!

manongjohn:
Need feedback...considering the following nested folders in Timeline:

image

In an attempt to avoid excessive xsheet header compression, Is this method for showing nested folders still helpful?

image

image

Still need to fix `Minimal` layout but I can compress buttons or the name height.

artisteacher:
> Is this method for showing nested folders still helpful?

I think it's less intuitive what is a parent/child. What about adding the folder color to the edge of the header when it is the left boundary of a folder?
Screenshot 2024-02-11 at 10 47 58 PM

artisteacher:
So much better! 

It would be nice to have a faster way to open the folders in minimum Xsheet view - like a long hold on the header or something.

manongjohn:
Ok.  With the last 2 commits I just pushed, I think I'm finished with this.  Will keep open a few more days for anyone who wants to test the latest changes/fixes.

PR #1392, opened 2024-02-18
OCA Import

manongjohn:
This OCA Import PR is based on the early commits of @janimatic's OCA Import PR #1371

This is a cleaned up and simplified version of the OCA Import logic that was being introduced.  I had to revise it for various reasons and chose to undertake it myself due to the complexity of trying to explain how image files needed to be loaded and handled in T2D.  Additional corrections were made.

OCA Import notes:
- OCA Layer Types
  - `paintlayer` - Supported, except for `blendingModes`.
  - `vectorlayer` - Supported via automatic conversion of SVG to PLI format
  - `grouplayer` - sub-layers will be imported without grouping
    - This will change in a separate PR after Timeline/Xsheet Folders (PR #1385) is merged.
- User will be prompted to `Import` or `Load`
  - Importing places PNG files in `+extras` and SVG files in `+drawings`
  - Loading (instead of Importing) vector layers will create the PLI file in the source directory.
- Layers without image files will not be imported.
- When importing, if the level already exists, user will be prompted to either keep, overwrite or rename.
- In case of import issues, warnings and errors will be displayed as it imports each layer
  - User can opt to ignore all future warnings and errors to speed up loading.

### DIFF
--- a/toonz/sources/tnztools/controlpointselection.cpp
+++ b/toonz/sources/tnztools/controlpointselection.cpp
@@ -1031,7 +1031,7 @@ void ControlPointSelection::addMenuItems(QMenu *menu) {
   menu->addAction(CommandManager::instance()->getAction("MI_SetLinearControlPoint"));
   menu->addAction(CommandManager::instance()->getAction("MI_SetNonLinearControlPoint"));
   menu->addSeparator();
-  //assert(ret); makes debug build fail
+  //assert(ret);
 }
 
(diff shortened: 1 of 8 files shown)


manongjohn:
@janimatic Thank you for introducing this new feature.

I'm sure it will be helpful to a number of people who use other drawing software, like Krita in their pipeline with T2D.

PR #1393, opened 2024-02-19
Fix Vector Stroke Group/Ungroup issues

manongjohn:
This will fix the following vector stroke group/ungroup issues:

- Ungroup now works even if the selection includes an ungrouped stroke.
- When inside a group, `Select All` will only select strokes inside the group.
- `Enter Group` and `Exit Group` actions are now properly recorded in History
- Corrected issues when using group/ungroup while inside a group.
- Fixed group undo/redo actions working on incorrect strokes caused by group reordering strokes


melieconiek:
Seems to work! Thank you for this!

RodneyBaker:
Some very nice fixes that most might not even notice on the surface but all of which make a big difference.
Many thanks!

Looks good here.

PR #1394, opened 2024-02-21
Fix crash entering Level Strip

manongjohn:
This fixes a crash when clicking on the level strip.

Crash caused by changes made in #1388.


manongjohn:
Since this bug renders the Level Strip useless in the nightly, I'm going to merge this now.

PR #1395, opened 2024-02-21
Fixes to SVG Import and Export

manongjohn:
I'm sure there are a number of issues surrounding SVG Import, and to a degree SVG Export.  This PR fixes only some of the more common issues people have run into.

### **SVG Import (SVG -> PLI converison)**
- Fixes fill and stroke opacity not being applied
  - Opacity was being read in but was not making palette entries for them, resulting in default fills, usually black, being used
- Default fill and stroke color to black when not specified
  - In some cases, fill/stroke information was provided but not color so it used whatever style it successfully used before. It now defaults to Style 1, which is usually black.
  - Forces default fill if there is no stroke information or stroke width = 0.  It is likely a fill-shape
  - Forces default fill if it's a closed shape without fill information
- Fixes incorrect application of stroke scaling information
  - Was applying scale information to stroke width incorrectly.
  - The scaling is now applied to the overall stroke for proper placement and sizing
- Fixes unnecessary grouping of every stroke.
  - If there is a fill associated with a stroke (i.e. closed shapes), the stroke is grouped and filled
  - If a fill is applied on a single stroke, the stroke will be ungrouped. 
  - If a fill is applied on a multi-stroke shape, the strokes will remain grouped.
  - If no fill is associated with the stroke, no grouping is performed.
- Fixes an issue where there was minimal fill information (i.e just opacity), but nothing else.
  - This now applies the default fill with the information provided
- Fixes stroke width importing wider than indicated in the file
  - This is likely an issue where device pixel ratio > 1 (usually 2) but the information in the file only assumes 1, causing stroke width to be doubled.
  - When building the stroke, it now adjusts for device pixel ratio > 1 to maintain proper stroke width
- Fixes imported SVG image displaying in the lower-right area of the canvas
  - Now centers the image on the canvas if width/height or viewBox information is provided
  - If none of the above is found, it will use the bounding box of all the strokes in the image to center
- Fixes importing filled shapes with cutouts
  - These shapes are multi-stroke shapes.  Added some logic to handle these properly
  - Subject to how the write records the outer path as well as our sometimes problematic fill process.
- Fixed closed shapes having an extra point causing circles to be deformed.
- Fixed moveTo logic not properly moving relative after moving absolute
- Fixed missing fills on shapes made by single strokes. (strokes with fills).
  - Added 0 width stroke to connect endpoints so fill can take place.

### **SVG Export (PLI -> SVG conversion)**
- Corrects how it writes `stroke-width` information to the file so it can be read by Inkscape.
- Now writes out `fill-opacity`, `stroke-linecap`, `stroke-linejoin`, `stroke-miterlimit` and `stroke-opacity`
- For closed shapes, now writes out the 'Z' in the path instruction

----
If you run into issues testing, please provide SVG file.  It would be helpful to know what generated the SVG as well.

### DIFF
--- a/toonz/sources/image/CMakeLists.txt
+++ b/toonz/sources/image/CMakeLists.txt
@@ -171,4 +171,4 @@ elseif(BUILD_ENV_UNIXLIKE)
     endif()
 endif()
 
-target_link_libraries(image Qt5::Core Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
+target_link_libraries(image Qt5::Core Qt5::Widgets Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -900,6 +900,7 @@ int nsvg__parseAttr(struct NSVGParser *p, const char *name, const char *value) {
       attr->fillColor = nsvg__parseColor(value);
     }
   } else if (strcmp(name, "fill-opacity") == 0) {
+    attr->hasFill = 1;
     attr->fillOpacity = nsvg__parseFloat(value);
   } else if (strcmp(name, "stroke") == 0) {
     if (strcmp(value, "none") == 0) {
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include <math.h>
 
+#include <QApplication>
+#include <QDesktopWidget>
 #include <QTextStream>
 #include <QFile>
 
@@ -2107,6 +2109,8 @@ TStroke *buildStroke(NSVGpath *path, float width, float scale) {
 //-----------------------------------------------------------------------------
 
 TImageP TImageReaderSvg::load() {
+  static int devPixRatio = QApplication::desktop()->devicePixelRatio();
+
   NSVGimage *svgImg =
       nsvgParseFromFile(m_path.getQString().toStdString().c_str());
   if (!svgImg) return TImageP();
@@ -2135,8 +2139,9 @@ TImageP TImageReaderSvg::load() {
     // vapp->setPalette(plt.getPointer());
     int startStrokeIndex = vimage->getStrokeCount();
     for (; path; path = path->next) {
-      TStroke *s = buildStroke(path, shape->hasStroke ? shape->strokeWidth : 0,
-                               shape->scale);
+      TStroke *s = buildStroke(
+          path, shape->hasStroke ? shape->strokeWidth / devPixRatio : 0,
+          shape->scale);
       if (!s) continue;
       s->setStyle(inkIndex);
       vimage->addStroke(s);
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -918,7 +918,8 @@ int nsvg__parseAttr(struct NSVGParser *p, const char *name, const char *value) {
     if (strcmp(value, "none") == 0) {
       attr->hasFillNone = 1;
     } else {
-      attr->fillColor = nsvg__parseColor(value);
+      attr->hasFillNone = 0;
+      attr->fillColor   = nsvg__parseColor(value);
     }
   } else if (strcmp(name, "fill-opacity") == 0) {
     attr->hasFillInfo = 1;
@@ -928,7 +929,8 @@ int nsvg__parseAttr(struct NSVGParser *p, const char *name, const char *value) {
     if (strcmp(value, "none") == 0) {
       attr->hasStrokeNone = 1;
     } else {
-      attr->strokeColor = nsvg__parseColor(value);
+      attr->hasStrokeInfo = 0;
+      attr->strokeColor   = nsvg__parseColor(value);
       if (!attr->strokeWidth) attr->strokeWidth = 1;
     }
   } else if (strcmp(name, "stroke-width") == 0) {
@@ -2232,21 +2234,23 @@ TImageP TImageReaderSvg::load() {
       TStroke *s = buildStroke(path, strokeWidth, shape->scale);
       if (!s) continue;
       s->setStyle(inkIndex);
-      vimage->addStroke(s);
+      int currentIndex = vimage->addStroke(s);
       strokeCount++;
-      int currentIndex = vimage->getStrokeCount() - 1;
       if (s->isSelfLoop() && !shape->hasFillNone) applyFill = true;
       // Single unconnected stroke shape with fill
       if (!s->isSelfLoop() && !shape->hasFillNone) {
+        int x = 1;
         // Create a connecting line for fill
         std::vector<TPointD> pts;
         pts.push_back(s->getControlPoint(0));
         pts.push_back(s->getControlPoint(s->getControlPointCount()));
-        TStroke *hiddenStroke = new TStroke(pts);
-        hiddenStroke->setStyle(0);
-        vimage->addStroke(hiddenStroke);
+        if (pts.front() != pts.back()) {
+          TStroke *hiddenStroke = new TStroke(pts);
+          hiddenStroke->setStyle(0);
+          if (vimage->addStroke(hiddenStroke) >= 0) x++;
+        }
         // Immediately group and fill
-        vimage->group(currentIndex, 2);
+        vimage->group(currentIndex, x);
         vimage->enterGroup(startStrokeIndex);
         vimage->selectFill(TRectD(-9999999, -9999999, 9999999, 9999999), 0,
                            paintIndex, true, true, false);
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -34,12 +34,15 @@ struct NSVGpath {
 };
 
 struct NSVGshape {
+  char id[50];
   unsigned int fillColor;    // Fill color
+  char hasFillColor;         // Flag indicating we have a fill color
   unsigned int strokeColor;  // Stroke color
+  char hasStrokeColor;       // Flag indicating we have a stroke color
   float strokeWidth;         // Stroke width
   float scale;               // Stroke scale
-  char hasFill;              // Flag indicating if fill exists.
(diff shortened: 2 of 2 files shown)


artisteacher:
Thanks for tackling this!

I've made a test SVG in Affinity Designer and it has several issues. It looks like this on import:
Screenshot 2024-02-21 at 7 28 59 PM
And the original SVG looks like this:
Screenshot 2024-02-21 at 7 27 44 PM
Issues:
- Styles didn't translate over (this was expected with the gradients, though it's too bad it ignores the colors entirely - if gradients aren't possible, maybe it should just use the first color listed in the gradient)
- The cut-out in the middle is filled in rather than transparent as it should be
- The corners/points are not sharp

I tried a couple versions of the file - with alternate vs winding set for the fill rule and changing the join/cap settings but they all import the same.
[SVG.zip](https://github.com/tahoma2d/tahoma2d/files/14366679/SVG.zip)

manongjohn:
Unfortunatly gradients and cutouts are not currently supported. There is nothing I can do about that at this time. 🙁

artisteacher:
Took another look at the SVG - part of the problem was the gradient ended up rasterized due to the transparent component. I updated the file so there is no transparency/rasterization. 

Updated SVG:
no-transparency
New import result:
Screenshot 2024-02-22 at 12 07 49 AM
You can at least see that the cut-out shape is there, even if it is incorrectly filled, and the shape is less perfect than the original. There are 2 fill rules - winding (non-zero) and alternate (even-odd) - theoretically, using one of those should get the correct result. I get if that's out of scope for now, though I think it's an important piece of SVG import. 

I just noticed what is distorting the points is the zero-thickness lines - visibility for these in on by default in Preferences. 

A different issue is that extra fillable regions are being created. This should just be two overlapping shapes -  where each shape is a single fillable region. The overlap is creating a third, unwanted region.
With generated vector styles & preview on:
Screenshot 2024-02-22 at 12 38 10 AM
 At least one of the shapes needs to be grouped to display as intended: 
Screenshot 2024-02-22 at 12 39 42 AM
This might be why there was a grouping of every stroke.


As far as the colors themselves go, someday, linear and radial gradients with just 2 colorstops could automatically be interpreted as generated vector styles. For more complex gradients that don't have a style equivalent, perhaps the colors could be loaded into the level palette so that they are at least available to be used. Or even better, the generated styles could be enhanced to allow for more colorstops.

manongjohn:
Ok, I think I somewhat fixed importing shapes with cutouts.

Typically shapes with cutouts are multi-stroke shapes. If the inner stroke is completely inside the outer stroke, most of the time it should fill correctly.  I had a case where it didn't fill at all until I shifted the points in Inkscape.  If the inner stroke is partially stretched to cross over the outer stroke, it may fill correctly. 

From Inkscape:
image

Imported into T2D:
image

How it fills is subject to how the SVG creator write out each of the stroke control points as well as our "lovely" fill logic, which is prone to errors.

Visual example of how the outer stroke (shown in black) was represented

image

artisteacher:
Much better! It makes sense for just the multi-stroke shapes to be grouped - since that's more how they behave in the original programs.

Do you know why the circle cut-out isn't quite circular?

I wonder if zero-thickness line visibility should be off by default? Every year I have a few kids that have their lines missing in renders as they didn't realize the lines had no thickness. But I guess then they'll end up with hidden lines which could be its own problem. Maybe there could be some other visual indicator that the lines are zero-thickness - a different color, multiple colors or etc.?

manongjohn:
> Do you know why the circle cut-out isn't quite circular?

Are you referring to the upper left shape in my drawing or the circle cutout in yours?

In your drawings, if I had to guess, that is where the ends of the strokes meet to create the shape and its applying the stroke ending information to all strokes in that multi-stroke shape making it appear off.

manongjohn:
Hmm. Upon closer inspection of the control points of your image, I think its treating the shape closure point as a cusp point instead of a belzier point. Seems to be fine in my Inkscape generated circles.

artisteacher:
Yes, the circle cut-out in my SVG. Looking with the control point editor, it looks like T2D is adding an additional vertex in the problem area:
In Designer (the red line is showing the curve direction):
Screenshot 2024-02-22 at 11 54 08 AM
In T2D:
Screenshot 2024-02-22 at 11 53 46 AM
If I reverse the curve, the extra control point gets added on the other side.
It looks like this happens to any circles made with Designer:
Import with preview on:
Screenshot 2024-02-22 at 12 08 48 PM
Original:
more-circles

manongjohn:
Just checked circles created from an on-line vector graphics editor and even rechecked the one created in Inkscape, and I see it's doing the same thing when importing.

I'm fairly certain it's the true start/end points of the stroke.  T2D likely doesn't handle closing closed shapes properly.  Will need to see what I can do about it.

TomDoingArt:
This is a great fix, thank you. More stability and predictability for SVG import. 
Looking at SVG stuff the past week I have come to appreciate how complex the SVG files can be. 
Here is a test case that shows how fill can become disassociated from its original lines.
In Inkscape I have 3 rectangles, each with a different line width. I was then experimenting with fills. 
It does not recognize the fill with partial opacity. It changes it to Black instead.
It also seems to get lost on position of the data, unless I have misread these results. 
It seems to put the fills in hierarchy appearance order rather than using the position x,y values.

SquaresDifferentWidths_WithPartialOpacityGrayFill
TahomaResultFromInkscapeExport
TheSVGfile
SquaresDifferentWidths_WithPartialOpacityGrayFill_Inkscape

manongjohn:
Please attach the SVG file

manongjohn:
The reason the fill opacity isn't being applied is because the data just shows `opacity:xxx`  In my Inkscape files, the fill opacity is recorded as `fill-opacity:xxx`.  I wonder why yours is different and if that's meant t be applied to both Stroke and Fill

TomDoingArt:
I'm messing up the file attachment process here. There are two SVG files, one to open in Inkscape. One in the simple format to open in Tahoma2D.
SquaresDifferentWidths_WithPartialOpacityGrayFill
This is the simple SVG file to drag into Tahoma.

TomDoingArt:
I am using this version: Inkscape 1.3.2 (091e20e, 2023-11-25, custom)

manongjohn:
Seems to be an issue with the way I am defaulting and applying fill/stroke information.  It's causing some of the rectangles to be filled with black incorrectly. Need to revise the logic.

Apparently there is an object opacity setting which is on top of fill/stroke opacity.  Will add support for it.

The SVG Importer does not handle a LOT of settings and scenarios and I don't expect to come close to fixing it with this PR.  I will need to be handled over time, especially if it's a more common issue. Hopefully this will handle the more basic use cases.

TomDoingArt:
I agree with your assessment. The fixes in this PR are definitely helping. 
I tested import on a file with a set of objects. Got a completely black rectangle in a T2D before these changes.
After these changes, still a black filled rectangle in the back but at least I can see some shapes showing up.
image

manongjohn:
Attach SVG. Let me see what's going on.

TomDoingArt:
When I attach SVG file using the paper clip icon, it interprets and displays it. If you right click that image, you can save it as a local .svg file.

TomDoingArt:
By the way, I found a W3C test suite for svg testing. 
This is where I got that file from:
https://github.com/web-platform-tests/wpt/blob/master/svg/import/paths-data-01-t-manual.svg

TomDoingArt:
More information about it here: https://github.com/w3c/svgwg/wiki/Testing

manongjohn:
I took a look at that test svg and here are my findings.

For reference, this is what I see in T2D:
image

1. Missing text is due to a `text` element that is currently not supported in T2D - I will not be fix this at this time because text in our vector images is handled differently than SVG images do.
2. Shapes defined by an unclosed stroke has fill (`M,C` and `M,C,S,m,c,s`) - I will attempt to fix this by adding a 0-width stroke to connect the end-points of a stroke so hopefully our fill can work
3. Strokes were separated based on the instruction on how it was to be created (`M,C,S,m,c,s`) - I will attempt to fix this.  The `s` directive apparently is not handled correctly in T2D.
4. Missing square (middle of `M,C,S,m,c,s`) - Apparently this is due to 2 overlapping exact rectangles. Not sure why. Likely some odd issue with our fill logic. Will look into, but may not fix

TomDoingArt:
Nice. Thank you for the analysis on this. 
I'm going to do more reading on the svg syntax.

manongjohn:
> 3\. Strokes were separated based on the instruction on how it was to be created (`M,C,S,m,c,s`) - I will attempt to fix this.  The `s` directive apparently is not handled correctly in T2D.

Determined it's not the 's' instruction on how to create the shape/line.  Somehow T2D is messing up this multi-line shape when it moves from absolute positioning 'M' to relative positioning 'm'

TomDoingArt:
Interesting. I want to look at the Tahoma code at some point. Do you intend to support the various versions of SVG? I am aware of 1.1, 1.2, and 2 is under development.

artisteacher:
I'm still seeing extra vertices creates on inner circles. Outer circles seem fine.

manongjohn:
Provide file?

artisteacher:
updated

manongjohn:
Hmm.  From what I can tell my looking at the drawing instructions on the file and other files I have with circles in it, when circles (or any closed shapes really) are drawn using absolute positioning, it seems to draw it perfectly.  If it draws the the circle using relative positioning, it somehow adds an extra control point.

artisteacher:
That is definitely it. Exporting the same file with relative coordinates toggled off displays as expected.

manongjohn:
Found the issue.  When checking the 1st and last points in a closed shapes, if they are not the same it adds some points to help close it. Problem is the compared points have to be exact, down to the precision. In this case the points were essentially the same but slightly different due to precision. The 1st point was absolute with .2 precision and the last point was calculated and had .4 precision.  I've adjusted to compare only .2 to avoid this issue. Pushing a change soon.

artisteacher:
That did it! SVG are so much more usable now.

artisteacher:
Ok, new interesting problem. This svg originally had a stroke with pressure changes - which exported as a second vector shape. So there should be one shape for the "outline" and one shape for the "fill". The result in T2D is a group with both the fill + outline shapes (where the outline area lost it's fill) and a second "fill" shape, that is directly underneath.
stroke

manongjohn:
Loading the paths individually, it does shape and fill them properly.  It doesn't like it when they are combined.

I'm pretty certain this is due to fill logic issues when filling regions encompassed by other already filled regions, or something to that effect.  I encountered this in another feature that I mentioned in Discord that I was toying with .   This is a much bigger issue and will not be fixed with this PR.

manongjohn:
I think from a more simple use case and knowing there are general fill issues, I think this is ready to move forward.  I will likely merge this tomorrow unless someone fines a new issue that requires fixing now.

manongjohn:
@artisteacher @TomDoingArt Even though I'm sure there are a lot more issues to be fixed, I think what you have help me find and fix is a big step to making SVG Import a lot more useful.  Thank you!

PR #1400, opened 2024-02-25
Update default layout and command bars

manongjohn:
Made the following UI changes:

- Added Level Strip to the 2D room
- Added the following buttons to the default Quick Toolbar and Command bars
  image
  - New Note Level
  - New Folder
  - Create Blank Drawing
  - Duplicate Drawing
  - Stop Frame Hold
- Added New Folder icon


artisteacher:
Nice! It might be helpful to add the insert command too.

manongjohn:
If I were to add `Insert`, I'd likely want to add `Delete`, `Cut`, `Copy`, `Paste`.  Where does it end? :)

These commands are also contextual, not really specific to the timeline/xsheet so I'm kinda iffy about adding them there.

artisteacher:
The difference I see is that every keyboard has the ability to delete/copy/cut/paste without any customization. Not every keyboard has an insert key, so faster access to the command would be helpful for those of us without one. 

On the other hand, I get the reservations.

manongjohn:
Moving forward with this as is for now.  Plenty of time to make more changes for 1.5

PR #1401, opened 2024-02-25
Show Symmetry and Perspective Grids commands and buttons

manongjohn:
In order to make using the Symmetry and Perspective Tools a bit more intuitive, I've made the following changes:

1. Created `Symmetry Guide` and `Perspective Grids` commands
2. Moved the visibility toggles from the `Grid and Overlays` button to dedicated buttons on the viewer titlebar
  image
3. Options to toggle symmetry guides and perspective grids are also shown in the `View` menu.
4. By default the Symmetry/Perspective buttons are `Off`, but will be automatically switched `On` when entering the tool.
5. User defined perspective grids are only enabled when the Perspective Grids button is on.  If not on, the default perspective (Vanishing Point at camera center) is used.


artisteacher:
So much more intuitive! This is great!

PR #1408, opened 2024-03-08
Fix fillable regions after SVG import

manongjohn:
This is a follow-up to #1395.

It is a semi-workaround to an issue with how fill regions were calculated with or without grouping when using SVG import.

In the old SVG import, every individual stroke was grouped.  In order to fill areas, you had to ungroup everything.  The ungrouping of strokes would look for new fillable regions as each stroke was ungrouped.

In new SVG import, individual strokes were not grouped so the calculation of fillable regions is done at the end. The results were different than before and were often wrong or incomplete.

This ultimately is due to our intersection/fill region detection logic which appears to do better when there are fewer strokes and intersections being used to calculate regions as in the case of ungrouping. It appears to be worse if there are too many available intersections to consider at once.

In order to find more accurate fill regions, this change will now calculate fillable regions as each stroke in the SVG is imported, rather than wait until the end.

### DIFF
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -2256,6 +2256,7 @@ TImageP TImageReaderSvg::load() {
                            paintIndex, true, true, false);
         vimage->exitGroup();
       }
+      vimage->findRegions();
     }
     if (startStrokeIndex == vimage->getStrokeCount()) continue;
 


artisteacher:
Nice! My last problem file is handled slightly better now. It at least looks correct on import. 

The left shows the import. The right shows what happens when moving with the select tool:
Screenshot 2024-03-08 at 12 27 33 AM

Still, moving in a good direction! The other files I tested with before are looking/behaving great.

manongjohn:
Interesting. When I imported that file to retest, it still loaded incorrectly for me.  I wonder why it loaded better for you. 🤔

Not surprised moving it did that. Movement does cause a recalculation of fillable regions and we still have some issues with that. Hopefully we can get it fixed soon. 🙂

Merging this shortly.

PR #1409, opened 2024-03-09
Fix OCA import/export frame range

manongjohn:
This changes OCA Import/Export's frame range as follows:

1. Export frames based on timeline/xsheet frame count instead of using Output Setting's frame range so the entirety of the scene is always exported.
   - This also applies to sub-scenes which can have more or less frames than the main.
   - Fixed an issue with using the wrong frame range for folders within sub-scenes.
2. Do not update scene's Output Settings frame range using Import's `startTime` and `endtime`.
    - This will allow the scene's Output Setting frame range to remain dynamic, changing as frames are added to extend the scene after import.

### DIFF
--- a/toonz/sources/toonz/ocaio.cpp
+++ b/toonz/sources/toonz/ocaio.cpp
@@ -187,12 +187,9 @@ bool OCAData::buildSubSceneGroup(QJsonObject &json, const QList<int> &rows,
 
   // Build a list of child rows
   QList<int> crows;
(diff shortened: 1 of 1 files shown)


manongjohn:
Windows builds are having an issue due to a Github issue.  It's been reported. Will retrigger the Windows build once the problem has been fixed.

manongjohn:
Windows build restarted.

PR #1411, opened 2024-03-09
Fix Fx not working with Implicit Holds

manongjohn:
This fixes an issue with some Fx not working in Preview mode/Render when using Implicit Holds. The workaround used to require adding Stop Hold Frames for all levels at the end of the scene.  This fix should eliminate the need for that workaround.

I had previously added logic for handling a preview/rendering performance issue with Particle Fx when Implicit Holds were used. When determining what textures to use, it needed to ignore implicit holds otherwise it would try to use them as additional textures. That logic had impacted some fx where implicit holds needed to be factored in.

Modified the logic to constrain the previously added logic to any `Particles` Fx, `Tiled Particles Iwa` and the `Flow Paint Brush Iwa` fx; essentially any fx where an image was being used as a texture but extended exposure didn't matter.

I did go through all stock Fx and they seemed to be working fine without any workarounds.  There were a few whose performance was slow, but they were also slow in OT. Additional feedback would be great.

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -612,7 +612,7 @@ TFxTimeRegion TFx::getTimeRegion(bool ignoreImplicit) const {
     if (port && port->isConnected() && !port->isaControlPort()) {
       TFx *fx = port->getFx();
       std::wstring fxName = fx->getName();
-      tr += fx->getTimeRegion((fx->getFxType() == "Toonz_columnFx" ? true : ignoreImplicit));
+      tr += fx->getTimeRegion(ignoreImplicit);
     }
   }
 
--- a/toonz/sources/stdfx/iwa_flowpaintbrushfx.cpp
+++ b/toonz/sources/stdfx/iwa_flowpaintbrushfx.cpp
@@ -36,7 +36,7 @@ void Iwa_FlowPaintBrushFx::getBrushRasters(std::vector<TRasterP> &brushRasters,
                                            const TRenderSettings &ri) {
   // ブラシテクスチャ情報
   TPointD b_offset;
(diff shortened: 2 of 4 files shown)


artisteacher:
This starts out great with just one particle or gradient or etc., but once getting there's a slightly more complicated FX setup, it breaks down. 
Screenshot 2024-03-12 at 9 27 15 PM
Preview stops responding and there's just a black or white screen. Regenerating, adding stop holds, and changing the active frame all have no impact.
Screenshot 2024-03-12 at 9 27 26 PM
This is a problem for every type of preview - ie. in the Viewer, Swatch Viewer, and Preview window.
Even after removing the FX, preview is still broken:
Screenshot 2024-03-12 at 9 28 35 PM


On a related note, but not unique to this PR, one of my students noticed today that when the FX editor is open but either the Scene or Output nodes are active, the preview is blank. They expected to see a preview.
Screenshot 2024-03-12 at 9 34 19 PM

manongjohn:
@artisteacher The problem appears to be with the Multi Radial Gradient Fx passing to the Particle node as a texture. OT has the same issue with it.

I think it might be an issue with passing Zerary Fx Columns (Fx that have actual columns) through Particle nodes.  I tried the same thing with Checkerboard fx and had the same problem. I'll probably handle that as a separately.

manongjohn:
Merging this now even though I believe there is 1 scenario in which a stop hold frame is needed at the end of a level.  It would be needed at the end of a level that has a color change or maybe a plastic deformation and is connected to a Particle Fx as a texture.  Without the stop frame, it won't use implicit held frames that have color/plastic deformation as a texture.

I am working on fixing the issue with Zerary Fx connected to Particle Fx as a texture.  It should also fix the scenario I just mentioned.

PR #1417, opened 2024-03-15
Fix slow preview generation of particle fx

manongjohn:
There is an issue with generating the preview of some fx or combinations of fx when they are connected to a Particle Fx as a texture.  The preview generation of 1 frame never completes.

From my investigation, there are some fx that are internally defined as having an unlimited timespan (i.e Zerary Fx; fx that has a column in the timeline/xsheet). The Particle Fx processes every frame of the fx's timespan as a source texture to determine which image to use for that frame. This is obviously problematic for time and resources.

Particle Fx should probably be fixed to do this differently.  For now, I've modified all Particle Fx, `Timed Particles Iwa` and `Flow Paint Brush Iwa` to limit any unlimited timespan based on the last frame of the scene.  This change impacts viewer preview mode, preview-renders, output-renders, task renders.

It will use the following logic to determine the last frame:
- Stop Play Marker, if enabled (preview modes only)
- last exposed drawing/stop frame or last key (if no stop frame); whichever is furthest out

For viewer preview mode only, if using Implicit Holds and you are beyond the last frame identified above, it will use the current frame as the last frame so it can still preivew the frame at that spot.

Warning: Even with this change Particle fx rendering may still be slow per frame, but should still finish.

Additional fix:
- Fixed Task viewer not able to split and render beyond the last exposed frame for scenes using implicit holds

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -600,7 +600,7 @@ int TFx::getReferenceColumnIndex() const {
 
 //--------------------------------------------------
 
-TFxTimeRegion TFx::getTimeRegion(bool ignoreImplicit) const {
+TFxTimeRegion TFx::getTimeRegion() const {
   if (m_imp->m_portTable.empty()) return TFxTimeRegion::createUnlimited();
 
   TFxTimeRegion tr((std::numeric_limits<double>::max)(),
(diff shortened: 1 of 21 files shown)


artisteacher:
It's so nice not needing the stop holds anymore with this sort of setup! This is working great on my end.

PR #1419, opened 2024-03-15
Fx Browser enhancement and Fx Room update

manongjohn:
Following enhancements have been made to the Fx Browser:

image

- Fx Browser is now a dockable pane
- Fx can be drag-dropped from the browser to the schematic
  - Add Fx - drag and drop in any unoccupied area
  - Insert Fx - drag and drop over a link between 2 fx
  - Replace Fx - drag and drop over an existing Fx
- Updated the default Fx room layout as shown above.


artisteacher:
This is a nice update! 

When the swatch viewer is enabled, and there are a lot of settings (ie. any of the particles are active), so that you need to scroll, you can see a little padding at the top of the boundary that seems unnecessary:
Screenshot 2024-03-16 at 3 01 18 PM

PR #1425, opened 2024-03-19
Fix popups hiding behind main window

manongjohn:
This change was primarily made to fix an issue on macOS builds where some popups may initially appear or move behind the main window when the main window has the focus.  This behavior appears to be macOS specific.

Some popups are defined using the `QDialog` base class but the window is not set as a `Qt::Tool` which would keep the popup in front of the main window.

T2D has a custom dialog class `DVGui::Dialog` used by most other windows.  This class sets the window as `Qt::Tool`, so I've opted to convert any popup defined with `QDialog` to use the custom class for consistency.

This change impacts the following popups:
- Camera Settings
- Cleanup
- Repeat (Cell menu)
- File Browser via Load Scene, Load Level, etc.
- Histogram
- Preferences
- Scene Settings
  - Edit Cell Marks
  - Edit Column Color Filters


artisteacher:
So much better! Thank you!

PR #1428, opened 2024-03-23
Inbetween from Timeline/Xsheet and Hook movement tweening

manongjohn:
This allows inbetweening vector frames directly in the timeline/xsheet in the same way inbetweens are done in the `Level Strip`.

- Similiar to Inbetweening in the Level Strip, you must select a starting explicit frame and an ending explicit frame with 1 or more explicit frames between them. Held frames do not count.
- You can select multiple vector column levels to inbetween at the same time
- Within the area of selection of multiple columns
  - Non-vector and empty columns are ignored
  - The first and last exposed frames of each column are used for inbetweening. They do not all need to have the same number of frames

After selecting the frames, use the `Cells` menu or right-click a vector cell and select the desired interpolation:

image

Can also trigger the inbetween if you set shortcuts for the interpolation

Inbetween menu options and shortcuts also work in the Level Strip.

Additional change:
- Hook movements are now also tweened via Level Strip and Timeline/Xsheet Inbetween

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1489,7 +1489,96 @@ void getLevelSetFromData(const TCellData *cellData,
       levelSet.insert(tmpLevel.getPointer());
     }
   }
-}
+};
+
+//=============================================================================
+// inbetweenWithoutUndo
+//-----------------------------------------------------------------------------
+
+void inbetweenWithoutUndo(TXshSimpleLevel *sl, std::vector<TFrameId> fids,
+                          TInbetween::TweenAlgorithm algorithm) {
+  if (!sl || fids.empty()) return;
+
+  int m              = fids.size();
+  TVectorImageP img0 = sl->getFrame(fids[0], false);
+  TVectorImageP img1 = sl->getFrame(fids[m - 1], false);
+  if (!img0 || !img1) return;
+
+  TInbetween inbetween(img0, img1);
+
+  for (int i = 1; i < m - 1; ++i) {
+    double t = m > 1 ? (double)i / (double)(m - 1) : 0.5;
+    t        = TInbetween::interpolation(t, algorithm);
+
+    TVectorImageP vi = inbetween.tween(t);
+    sl->setFrame(fids[i], vi);
+    IconGenerator::instance()->invalidate(sl, fids[i]);
+  }
+};
+
+//=============================================================================
+// UndoInbetween
+//-----------------------------------------------------------------------------
+
+class UndoInbetween final : public TUndo {
+  TXshSimpleLevelP m_level;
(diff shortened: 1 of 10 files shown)


RodneyBaker:
@manongjohn I'm probably just missing it but...

> Can also trigger the inbetween if you set shortcuts for the interpolation

I can't find the commands to set shortcut keys for this.
I did start with a clean install just to double check.

manongjohn:
Should be here:

image

RodneyBaker:
Thanks @manongjohn  That does in fact work.
Not sure what I was doing initially to make that not work.

manongjohn:
> Thanks @manongjohn That does in fact work. Not sure what I was doing initially to make that not work.

Maybe you set the other interpolation shortcuts?  There is a set of shortcuts (`xxxx Interpolation`) which changes the default interpolation setting in Preferences that is used when setting animation keys on the timeline/xsheet.

papio-pl:
Great addition. It will save a lot of time.

A little off topic question: How hard it would be to include hooks to vector inbetweening? Everytime when I'm using vector inbetweenning I need to manualy move hooks to match the drawings. And everytime I think it should be much more easy to automaticaly place those simple 2D points (hooks) than vectors:)

manongjohn:
I posted a PR for Frame Range enhancenents before this one. At request from another user, I added Frame Range option to the Hook Tool to do what you described.

DarrenTAnims:
Will this only inbetween on the drawing numbers shown in the timeline or across all drawings for the level between the 2 selected drawings?

For example, if you've got a level with 10 drawings, numbered 1 to 10.
And you've drawn your keys on drawing numbers 1 and 10.

Scenario 1:
If you've exposed the drawings in this order: 1, 2, 3, 4, 5, 10
Will only the inbetweens appear ONLY on drawings 2, 3, 4 & 5 or on all drawings from 2 to 9?

Scenario 1:
If you've exposed the drawings in this order: 1, 5, 4, 3, 2, 10
Will only the inbetweens appear on drawings 5, 4, 3, 2 in that order or in numeric order?

manongjohn:
> Will this only inbetween on the drawing numbers shown in the timeline or across all drawings for the level between the 2 selected drawings?

It will only use drawings exposed in the timeline and fill them in the order the frames are exposed.

> Scenario 1:
> If you've exposed the drawings in this order: 1, 2, 3, 4, 5, 10

ONLY on drawings 2, 3, 4 & 5

> Scenario 1:
> If you've exposed the drawings in this order: 1, 5, 4, 3, 2, 10

appear on drawings 5, 4, 3, 2 in that order

manongjohn:
@papio-pl After a discussion on the Discord, I've added hook movement tweening to this PR.

PR #1446, opened 2024-04-17
Add horizontal scrollbar to file browser 

### LINKED ISSUE #222: File Browser missing horizontal scroll bar
Here I've resized the File Browser window to squash it horizontally, you can see that the left panel shows a horizontal scroll bar as is expected when its contents cannot be fully shown. However, the right panel shows very little as most is cut off, yet there is no horizontal scroll bar present at the bottom of the panel. It is still possible to horizontally scroll via the mouse-wheel to show the rest, but it feels oddly inconsistent that it is missing.
image

mb4c:
Closes #222 

Screenshot from 2024-04-17 03-09-28


artisteacher:
Works great on macOS. Thanks for the fix!

manongjohn:
LGTM. Thanks for fixing this!

PR #1465, opened 2024-05-09
Raster/MyPaint style brush Smooth option

manongjohn:
This adds the `Smooth` brush setting to the Raster Brush tool.

Smooth setting is also available for MyPaint styles in Raster and Smart Raster tools. This smooth setting is independent of MyPaint style's `Slow Tracking Position` setting and is applied before it.


beeheemooth:
Great! This option is very necessary for raster.

However, some styles leave a straight tail after lifting the stylus.

Screen_1262

manongjohn:
I know with smooth, when you stop drawing it immediately completes the line from wherever it stopped drawing at to wherever the pen was last at.  That is likely what you are seeing.  The line might be because it's going from some pressure to 0 pressure very quickly.  Will review to see what I can do about that.

PR #1471, opened 2024-05-15
Onion Skin Opacity control

manongjohn:
This adds a simple control to allow you to adjust the opacity of an onion skinned frame:

image

To use, right-click a relative or fixed onion skin marker to open the marker's opacity control.  Adjust the slider or enter the desired opacity %.  To reset and resume auto-calculation of opacity either check `Auto` or remove and re-add the onion skin marker.

Normal (auto-calculated) onion skin colors are solid colors (red/green/blue).  Onion skin's who opacity have been explicitly set will have a darker center.


ArchieLinux:
Neat feature tweak. Thanks!

RodneyBaker:
Enjoying this update a lot.  So nice to be able to adjust onion skinning at this level.

A future enhancement to consider might be to allow the user to save current onion skinning assignments to file and to load/reload those.  In this way onion skin settings would reusable and not require manual setup in every case.  I looked at the code and confess I'm not sure what form that saving/loading of Onionskin settings would take.  Having the save/load option on every Right Click dialog seems a bit of an overkill.  Perhaps others can suggest a more effective way.

Perhaps a lesser form of that save/load option might be a copy/paste option that stores the current onion skin opacity and allows pasting into selected onionskin marker.  We can't select multiple onion skin markers at the same time so that seems unnecessarily complicated compared to an option that would save values to a text file and allow them to be loaded again.

Issue #1466, opened 2024-05-10
Question about the air brush in Tahoma 


Codeoloketuyi:
### Bug Description

I don't know if this is normal in Tahoma but the airbrush looks very weird in nature compared to other airbrushs in other software.
For example krita airbrush:

https://github.com/tahoma2d/tahoma2d/assets/137768661/c778acc0-34a3-4b33-ba8e-92aad9c7bbe6

Example 2 Tahoma airbrush:

https://github.com/tahoma2d/tahoma2d/assets/137768661/4b635a7d-85a4-4986-a7fb-577ec88fef51

This are the video I recorded to show the problem.

### Software Version

Nightly (Specify below)

### Software Version Information

1.4.3

### Operating System

Window 11 

### Graphics Card

Nividia rtx 4050

RareFox2003:
I have the same problem with Airbrush. Didn't find a solution?

PR #1480, opened 2024-05-22
Fix pen tap leaving max size dot

manongjohn:
This fixes an issue on Raster levels where dots are drawing using the max size when tapping on the canvas with a stylus. 

Added similar logic found in other brushes to use the smallest size.

This started happening after #1459 which now always draws a dot, sometimes too faint to see, whenever tapping the canvas.


Codeoloketuyi:
It similar to the same problem I faced when using the air brush at max size.
https://github.com/tahoma2d/tahoma2d/issues/1466#issue-2289398662

PR #1486, opened 2024-05-24
Vector Inspector Initial Release

TomDoingArt:
# Vector Inspector

Vector Inspector is a new panel and viewer for vector-level stroke details.

## Opening the Panel

**To Open:**

`Panel > Vector Inspector`

Requires a Vector level to be the current active level.

Works with the Select Tool to highlight selected strokes and allow stroke selection from the Vector Inspector stroke list.

### The Identity of the Stroke

- **\#**: The ordinal index of the stroke in the full list of strokes.
  
- **Group Id**: The identifiers of the groups this stroke belongs to, comma-separated. 
  - The active group is the leftmost.
  - The current entered group is shown within parentheses. Example: `(5).3.1` indicates that the stroke belongs to groups with IDs of 5, 3, and 1. The current status is that group 5 has been entered.
  
- **Id**: The stroke identifier, a per-session value assigned on opening the level.
  
- **StyleId**: The palette style index value currently assigned to this stroke. 
  
- **Self Loop**: Indicates if the stroke is a closed shape. If `true`, the first control point matches the last.

### The Points of the Stroke

- **Quad**: The identifier of a quadratic equation along the stroke. Also known as a "chunk".
  
- **P**: One of the three "P" values of the quadratic equation. The types are `P0`, `P1`, and `P2`.
  
- **x**: The x-axis value of the Cartesian coordinates for this specific point.
  
- **y**: The y-axis value of the Cartesian coordinates for this specific point.
  
- **Thickness**: The line thickness at this point.

### Filter the List

A default filter is set on the `P` column, value `2`. This reduces the list to 1 row per stroke as each stroke has only one point with a `P2` value listed. Remove the `2` value to see a full unfiltered list of stroke points.

### Sort the List

Click the sort triangle icon on the `\#` column to alternate between ascending and descending sort order.


toonz/sources/toonz/menubar.cpp
```diff
@@ -669,6 +669,7 @@ void TopBar::loadMenubar() {
   addMenuItem(windowsMenu, MI_OpenComboViewer);
   addMenuItem(windowsMenu, MI_OpenLevelView);
   addMenuItem(windowsMenu, MI_OpenFileViewer);
+  addMenuItem(windowsMenu, MI_OpenVectorInspectorPanel);
```
manongjohn, line 672: Please move this new entry down to the bottom section of the Panels menu list. This isn't an actual image viewer and is more a utility feature.

toonz/sources/toonz/CMakeLists.txt
```diff
@@ -416,7 +419,12 @@ endif()
 qt5_wrap_cpp(SOURCES ${MOC_HEADERS})
 
 if(BUILD_ENV_MSVC)
-    add_executable(Tahoma2D WIN32 ${HEADERS} ${SOURCES} ${OBJCSOURCES} ${RESOURCES} ${RC_FILE})
+    add_executable(Tahoma2D WIN32 ${HEADERS} ${SOURCES} ${OBJCSOURCES} ${RESOURCES} ${RC_FILE}
+        vectorinspector.h
```
manongjohn, line 423: Why are .h and .cpp files listed here?  They are already mentioned above.

toonz/sources/toonz/vectorinspector.cpp
```diff
@@ -0,0 +1,678 @@
...
+
+    QObject::disconnect(TApp::instance()->getCurrentScene(),
+                        &TSceneHandle::sceneChanged, this,
+                        &VectorInspectorPanel::onSceneChanged);
+
+    setSourceModel(new MultiColumnSortProxyModel);
+  }
+
+  m_field->setText("Level type: " + QString::fromStdString(levelType) +
+                   " at: " + utcNow.toString(Qt::ISODate));
```
manongjohn, line 309: Is there a particular reason for showing current time?  I don't think it provides any value.

As this is for vectors only, I don't particularly see the need for Level Type information.  If you are not on a Vector level, the panel information should be blank.  If anything should be displayed, perhaps the Level Name and Frame number might be of more value than Type and Time.

toonz/sources/toonz/vectorinspector.cpp
```diff
@@ -0,0 +1,678 @@
...
+
+  filterSyntaxComboBox = new QComboBox;
+  filterSyntaxComboBox->addItem(tr("Regular expression"), QRegExp::RegExp);
+  filterSyntaxComboBox->addItem(tr("Wildcard"), QRegExp::Wildcard);
+  filterSyntaxComboBox->addItem(tr("Fixed string"), QRegExp::FixedString);
+  filterSyntaxLabel = new QLabel(tr("Filter &syntax:"));
+  filterSyntaxLabel->setBuddy(filterSyntaxComboBox);
+
+  filterColumnComboBox = new QComboBox;
+  filterColumnComboBox->addItem(tr("#"));
```
manongjohn, line 99: I suggest changing this to "Stroke #" to make it clear to a user what this number represents.

manongjohn (changes requested):
The information on this panel is kind of meaningless and most likely very confusing to those who don't understand what goes into a vector stroke.

Perhaps add some Tool Tips to column headers, if possible, to help explain what it is they are looking at.  If not, perhaps a button somewhere on the panel that can pop up a small information box explaining what the columns mean.

manongjohn:
Some things I notice at a quick glance

- The panel doesn't update...
  - When new strokes are drawn
  - When you undo/redo a drawn stroke
  - Delete a stroke
  - Cutting stroke (either with Cutter tool or Eraser)
  - Create a Blank frame in the vector level.
  - I'm assuming Duplicate doesn't refresh the panel also
  - Stroke's style ID changes

- Selecting a stroke in the panel doesn't highlight the stroke in the viewer unless you are in Selection Mode.  To make it more useful to other vector tools, I suggest highlighting the stroke, red/magenta perhaps.

toonz/sources/toonz/CMakeLists.txt
TomDoingArt: I'll resolve it.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: Timestamp so I could tell if the panel was updating. If the timestamp was "frozen" at a time in the past I would recognize it was not updating. In a screen capture as part of discussion troubleshooting Vector fill issues for example, the timestamp could be there. Similarly, the level type is to show what the panel is currently pointing to.
I'll look at putting Level Name and Frame number instead.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: It was Stroke originally, but that kept overlapping the sort order triangle. I'll take another look at this. Qt 5.15.2 seems to have a more prominent triangle and I may learn the way to control spacing to include space for the triangle.

TomDoingArt:
> Some things I notice at a quick glance
> 
> * The panel doesn't update...
>   
>   * When new strokes are drawn
>   * When you undo/redo a drawn stroke
>   * Delete a stroke
>   * Cutting stroke (either with Cutter tool or Eraser)
>   * Create a Blank frame in the vector level.
>   * I'm assuming Duplicate doesn't refresh the panel also
>   * Stroke's style ID changes
> * Selecting a stroke in the panel doesn't highlight the stroke in the viewer unless you are in Selection Mode.  To make it more useful to other vector tools, I suggest highlighting the stroke, red/magenta perhaps.

The primary intended use case is to get insight into how the strokes are currently structured "behind the scenes" to make some of the vector fill issues explainable. To have some real-time facts for discussion. It was thought that the limitation of being in Select Tool is acceptible to this use case. I considered putting a refresh button to handle those cases where stroke changes are not automatically reflected in the display, but since it is possible to refresh by pressing the a shortcut key to jump to another frame and back or to click another level and back. 
Accomodating all the situations listed here can be done going forward. The technical costs include: tracking more events, performance (updating takes time, we've seen how some users have hundreds of points on a single stroke in an image, the SVG imported stuff from VPaint), code complexity/maintainability. 
This list can be seen as a roadmap for enhancements to this panel as the panel proves its usefulness. If nobody uses it actively, it may remain just a limited use troubleshooting/diagnostic tool.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: The first column:
Stroke #, can be confused with the Id column which is the Stroke ID. The Stroke ID persists across movement up/down the list of strokes making it a way to track a stroke in discussion. 

It is an ordered Position value in the list of strokes from Back (0) to Front (highest value). A label of Position, or Pos. #. , Order #, or no label at all may be clearest. It is similar to the row number on a spreadsheet border in that the numbers are ascending or descending representing the stroke data in the position from front to back, or back to front.
Here is what I looks like with no label on that column:
image

manongjohn: `Stroke Order` is probably most appropriate

I believe I know what `Quad` is supposed to be, though not sure those who haven't looked at the code will know what those are supposed to be. What is `P`?

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: I can update the window title in a manner similar to other windows or panels have done.
Do you have a preference for separator?
I see several in use. Here I am using vertical bars, it looks clean. 
I see single or double colon used in other windows. 
image

manongjohn: If you are going to put it in the title bar area, make sure the title bar doesn't hide if the pane is docked and locked otherwise the information is lost.  If I had a choice, follow the Scene Viewer separator.

TomDoingArt: I trimmed things quite a bit getting the panel to not have extra "stuff" left over from its origin as a mash-up between HistoryPanel and an example from Qt. So, currently it is not dockable. I will bring that functionality back and make sure the title and info doesn't disappear when docked.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: P0, P1, P3 are the symbols used in the Quadratic equation formula. 
image

FYI: I added a markdown formatted write-up describing VectorInspector to the first message of this Pull Request.

TomDoingArt: With the heading Stroke Order and the sort indicator:
image

toonz/sources/toonz/menubar.cpp
TomDoingArt: Completed.

TomDoingArt:
> The information on this panel is kind of meaningless and most likely very confusing to those who don't understand what goes into a vector stroke.
> 
> Perhaps add some Tool Tips to column headers, if possible, to help explain what it is they are looking at. If not, perhaps a button somewhere on the panel that can pop up a small information box explaining what the columns mean.

Completed.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: This has been implemented.

toonz/sources/toonz/CMakeLists.txt
TomDoingArt: Completed.

toonz/sources/toonz/vectorinspector.cpp
TomDoingArt: Completed.

TomDoingArt:
> > Some things I notice at a quick glance
> > 
> > * The panel doesn't update...
> >   
> >   * When new strokes are drawn
> >   * When you undo/redo a drawn stroke
> >   * Delete a stroke
> >   * Cutting stroke (either with Cutter tool or Eraser)
> >   * Create a Blank frame in the vector level.
> >   * I'm assuming Duplicate doesn't refresh the panel also
> >   * Stroke's style ID changes
> > * Selecting a stroke in the panel doesn't highlight the stroke in the viewer unless you are in Selection Mode.  To make it more useful to other vector tools, I suggest highlighting the stroke, red/magenta perhaps.
> 
> The primary intended use case is to get insight into how the strokes are currently structured "behind the scenes" to make some of the vector fill issues explainable. To have some real-time facts for discussion. It was thought that the limitation of being in Select Tool is acceptible to this use case. I considered putting a refresh button to handle those cases where stroke changes are not automatically reflected in the display, but since it is possible to refresh by pressing a shortcut key to jump to another frame and back or to click another level and back I was not hindered when using the tool, so far. Accomodating all the situations listed here can be done going forward. The technical costs include: tracking more events, performance (updating takes time, we've seen how some users have hundreds of points on a single stroke in an image, the SVG imported stuff from VPaint), code complexity/maintainability. This list can be seen as a roadmap for enhancements to this panel as the panel proves its usefulness. If nobody uses it actively, it may remain just a limited use troubleshooting/diagnostic tool.

Let me know if any or all of these should be implemented before the initial release of Vector Inspector can be done. I can work on them for that initial release if necessary, or as follow-on work.

changes.patch
manongjohn, line 1: Please delete this file

TomDoingArt: Done. A leftover from the cleaning up process to get the branch to being only my changes.

manongjohn (approved):

PR #1456, opened 2024-05-03
Qt 5.15

manongjohn:
This PR makes the following changes:

- Windows builds have been migrated to Qt 5.15.2. (macOS and Linux already use 5.15)
  - This uses a custom Qt 5.15.2 version built with WinTab support by @shun-iwasawa (see below)
  - Build instructions have been updated
- Changed the C++ standard from C++11 to C++17 (required by Qt6)
  - Fixed some code that did not meet C++17 standards
- Removed all macros for compiling with older versions of Qt.
  - This change requires building on a minimum of Qt 5.15.2 going forward
- Replaced deprecated Qt classes and methods as of 5.15.2 (preparation for Qt6)
- Replaced almost all obsolete Qt classes and methods through the latest 5.15 version. (preparation for Qt6)

------
### Local Windows Builders
For those using local builds or developing for T2D, you will need to redo/update your build environments as follows after this is merged:
- Git pull tahoma2d/master into your fork's master branch.
  - I suggest rebasing any active branches in your fork against the updated master
- Download and install Shun-iwasawa's custom Qt5.15.2 library from https://github.com/shun-iwasawa/qt5/releases/tag/v5.15.2_wintab. See build instructions for more information.
- Redo the CMake step in the build instructions

PR #1492, opened 2024-05-27
Fix RGB Picker Tool's Pick Screen

manongjohn:
This fixes `RGB Picker Tool`'s `Pick Screen` not working after updating for Qt5.15 (#1456)


artisteacher:
This was already working as expected on macOS and continues to do so here.

PR #1527, opened 2024-07-14
Fix stylus event processing (Part 2)

manongjohn:
Qt5.15 is generating stylus move (TableMove) events differently on 2 different Windows laptop-tablet than it used to when we were on Qt5.9.  Using the customized version of Qt5.15 seems to also be playing a small role in some situations as well.  This appears to be an issue with how the WinTab drivers is generating events, and some of these issues may not be present with users that have separate graphics tablets.

This PR will address the following issues which fixes it in my system but shouldn't impact graphics tablet users:

- **Single tap causing 2 taps (base Qt5.15 issue; Windows only)**

Doing a simple tap on the canvas produces a TabletPress+TableRelease followed immediately by a MousePress+MouseRelease instead of these events being interleaved like i was i Qt5.9.  Because of this, it acts like it's tapped 2x (not double-click) and causes issues with `arc`, `multi-arc` and `polyline` modes. 

Added logic to ignore the MousePress if we just processed a TabletRelease.

- **Loss of pen pressure when using WinTab (Custom Qt5.15 issue; Windows only)**

On my 1 laptop-tablet, it appears the TabletMove events randomly are missing that the button is pressed (active drawing) and as such immediately turns off stylus drawing.   Then a MousePress event shows up and reinitializes drawing from mouse instead of stylus and ignores all pressure coming from valid stylus events.

Added logic to ignore TableMove events that has pressure but no button press.  I determined I cannot use the pressure in these bad stylus events because it incorrect. 

- **Drawing circles starts off with a flat line (Not a Qt related issue, impacts all OS)**

The current logic for processing stylus moves only processes TabletMove events every 20ms due to the volume of these events being generated.  This logic was added many years ago to address an extreme lag issue with scaling Raster images, however the way it was implemented affected all tools. It is also the source of some users complaints about drawing experience.

Recently introduced/modified logic for Windows TabletMove events corrects this issue and made drawing feel more fluid.  I've unified the logic so it applies to all OSes.  I've verified it helps with Linux builds but I don't have a mac to check.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -748,7 +748,12 @@ void SceneViewer::mousePressEvent(QMouseEvent *event) {
   // so m_gestureActive is the marker for rejecting touch events
   int source = event->source();
   // if this is called just after tabletEvent, skip the execution
-  if (m_tabletEvent) return;
+  if (m_tabletEvent
+#ifdef _WIN32
+    // See SceneViewer::doQuit() for why this is here
+    || m_tabletState == Released
+#endif
+    ) return;
   // and touchscreens but not touchpads...
   if (m_gestureActive && m_touchDevice == QTouchDevice::TouchScreen) {
     return;
@@ -1044,12 +1049,29 @@ void SceneViewer::doQuit() {
   m_mouseButton = Qt::NoButton;
   // Leave m_tabletEvent as-is in order to check whether the onRelease is called
   // from tabletEvent or not in mouseReleaseEvent.
-  if (m_tabletState == Released)  // only clear if tabletRelease event
+  if (m_tabletState == Released) {  // only clear if tabletRelease event
+#ifdef _WIN32
+    // If this was a mouse event and tablet was just relased, clear tablet state now
+    if (!m_tabletEvent) m_tabletState = None;
+#endif
     m_tabletEvent = false;
+  }
   // If m_tabletState is "Touched", we've been called by tabletPress event.
   // Don't clear it out table state so the tablePress event will process
   // correctly.
-  if (m_tabletState != Touched) m_tabletState = None;
+  if (m_tabletState != Touched
+#ifdef _WIN32
+      // On some Window systems and perhaps tablets, when you tap with a stylus,
+      // it submits a TabletPress+TabletRelease immediately followed by a
+      // MousePress+MouseRleaese, rather than interleaved. This results in
+      // tapping 2x (not a double-tap), and causes an issue with stuff like
+      // arc, multi-arc and polyline modes.
+      // To fix, keep the Released state until we get the mouse release
+      && m_tabletState != Released
+#endif
+      )
+    m_tabletState = None;
+
   m_mouseState                                = None;
   m_tabletMove                                = false;
   m_pressure                                  = 0;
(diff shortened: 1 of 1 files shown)


artisteacher:
With a macOS/ipad combo, most cases are working well with a stylus. Drawing is really smooth and responsive!

I'd still love to see 3 preexisting issues improved:

- 3rd point for the arc should let you click & drag to keep adjusting the apex of the arc. As the 3rd point where you click is not the actual top of the arc, it is challenging to get the actual apex you want.
- Small, closed vector shapes still can end up distorted with bigger curves suddenly jutting out than were actually drawn

https://github.com/user-attachments/assets/29110db5-448c-4b41-ae23-a2a72f4bc3cd

- Default brushes don't handle pressure changes as well as the Mypaint brushes - Ie. for the default vector brush, I need either a huge amount of pressure to get line variation or to drastically increase the maximum brush size.

manongjohn:
Thanks for testing the mac side.

I need to review and handle the issues you brought up separately.

As for the issues...

> * 3rd point for the arc should let you click & drag to keep adjusting the apex of the arc. As the 3rd point where you click is not the actual top of the arc, it is challenging to get the actual apex you want.

For the "arc" determination, its based on where the stylus is hovering relative to the 2 points that were set. It is visualized by a guiding arc. You should be able to adjust the arc using the guiding arc before committing which happens as soon as you press the canvas. The video doesnt seems to show this.  Seems like the arc is getting set and you're trying to move it afterwards?  I cant quite tell when you pressed the stylus.

> * Default brushes don't handle pressure changes as well as the Mypaint brushes - Ie. for the default vector brush, I need either a huge amount of pressure to get line variation or to drastically increase the maximum brush size.

I do notice the the amout of pressure required to increase the width of the brush is based on the max size of the brush. Might actually be based on the difference between min and max values.  I know for this reason, i normally keep the max around 20 (my min is usually 1).

artisteacher:
>  You should be able to adjust the arc using the guiding arc before committing which happens as soon as you press the canvas. The video doesnt seems to show this. Seems like the arc is getting set and you're trying to move it afterwards? I cant quite tell when you pressed the stylus.

There's no guiding arc until you tap the 3rd time. So if it's supposed to be visible/controlled by hovering once you've set the 2nd point, that's just not happening. And as soon as you can see the guiding arc, that 3rd point is set and can't be adjusted.

>  Might actually be based on the difference between min and max values. I know for this reason, i normally keep the max around 20 (my min is usually 1).

This definitely seems to be the case. I guess part of the issue is that a mouse needs a lower default compared to a stylus. So the way things are now, there's no way to win.

manongjohn:
I think there were weird hovering issues in macos in the past? I will take a look at the missing guiding arc when hovering.  Question, when using polyline, do you see the guiding line between points as you move?

artisteacher:
With polyline there's no guiding line unless you click and drag to make a curved segment. In that case you can see/adjust the handles as expected.

manongjohn:
What about with mouse?

artisteacher:
The guidelines show as expected for both polyline and arcs.

artisteacher:
As requested, viewer event logs for arc and polyline:
[multi-arc_stylus-only.txt](https://github.com/user-attachments/files/16241839/multi-arc_stylus-only.txt)
[polyline_stylus-only.txt](https://github.com/user-attachments/files/16241840/polyline_stylus-only.txt)

PR #1530, opened 2024-07-16
Fix crash opening PLI (Vector) files and folders

manongjohn:
This fixes 3 issues regarding opening files/folders with a .PLI extension

- **Prevent crash opening partially corrupted PLI files (missing frames)**

Sometimes a PLI file is partially corrupted in such a way that it's missing the last few frames.  This leads to a crash when it tries to work with the number of frames indicated in the file but not all were loaded.

Added logic to detect when the number of frames read is less than what was indicated in the file.  Rather than failing, it will keep whatever it was able to read. Usually the corrupted frame is the last frame missing an image. A warning is displayed and the level is immediately flagged as dirty so that the user is forced to deal with it when saving.

- **Prevent crash opening folder named with .PLI extension**

When the browser or any dialog that accesses the user's drive encountered a folder that ends with `.PLI`, it would crash when it failed to open and read it to get image information.  Theoretically this could happen with any supported image file, but some handle the error properly.

Logic added to only open browser items to get image information if it is not a folder.

- **Prevent crash loading region information in PLI file (strokeless edge)**

There are some scenarios where the region information in the PLI file contains edge information with no stroke.  In some cases this cause a crash.

Added logic to ignore edge information if it is missing stroke information.  This primarily seems to impact loading files that have this scenario.  I don't know what normally causes this scenario so I am not sure what impact the change will have during normal vector drawing/filling operations.

### DIFF
--- a/toonz/sources/common/timage/tlevel.cpp
+++ b/toonz/sources/common/timage/tlevel.cpp
@@ -11,7 +11,8 @@ TLevel::TLevel()
     : TSmartObject(m_classCode)
     , m_name("")
     , m_table(new Table())
-    , m_palette(0) {}
+    , m_palette(0)
+    , m_partialLoad(false) {}
 
 //-------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


RodneyBaker:
That for the fix @manongjohn 
This appears to resolve the issue with T2D/OT just seeing/encountering a corrupted vector file.
Still other issue with corrupted vector files but... you are aware of that.
Just want to say thanks for the fix as that will save many users from crashing.

PR #1535, opened 2024-07-29
OpenToonz Patches thru 7/27/2024

manongjohn:
The following have been ported from OT with some adjustments for T2D. This list includes some stuff I was waiting for migration to Qt5.15

opentoonz/opentoonz/#4217 - Zoom in/out and fit floating panel geometry commands  by @shun-iwasawa
opentoonz/opentoonz/#4848 - Fix update sound after editing cells  by @shun-iwasawa
opentoonz/opentoonz/#4872 - Fix the dialogs to be closed when clicking the buttons, instead of pressing.  by @shun-iwasawa
opentoonz/opentoonz/#4878 - quick fix  by @shun-iwasawa
opentoonz/opentoonz/#4863 - Studio Palette Tweaks  by @shun-iwasawa
opentoonz/opentoonz/#4843 - Unify preview and camstand visibility toggles option by Contrail Co., Ltd.  by @shun-iwasawa[^1]
opentoonz/opentoonz/#4874 - Disable drag-moving cells option by Contrail Co.,Ltd.  by @shun-iwasawa
opentoonz/opentoonz/#4966 - Fix missing initialization  by @shun-iwasawa
opentoonz/opentoonz/#5302 - Macos undo freeze  by @Funinja
opentoonz/opentoonz/#5317 - Refactor FFmpeg format check on launch  by @shun-iwasawa
opentoonz/opentoonz/#5356 - [ui] Decouple x-sheet text colors  by @konero[^1]
opentoonz/opentoonz/#5320 - Update cleanup_default.tpl to be empty  by @Astr01d
opentoonz/opentoonz/#5361 - Fix typo  by @RoboSchmied[^1]
opentoonz/opentoonz/#5379 - [ui] Extra x-sheet colors  by @konero[^1]
opentoonz/opentoonz/#5416 - Correct Minor Typos by @RodneyBaker
opentoonz/opentoonz/#5419 - windows configure ThirdParty  by @janimatic[^2]
opentoonz/opentoonz/#5284 - Remove TSmartObject from TProject  by @otakuto[^1]
opentoonz/opentoonz/#5523 - Modify CMakeLists.txt for Windows  by @shun-iwasawa[^2]

[^1]: Modified some commits for T2D changes
[^2]: Added post commit for T2D's tiff version

PR #1537, opened 2024-07-30
Fix sound column issues

manongjohn:
This fixes the following issues related to sound columns:

- Fix excessive lag moving sound cells

If you have many sound levels and/or lengthy sound levels, moving a sound cell immediately causes the overall audio to be regenerated.  This happens with every cell movement made, resulting in lag.

Modified the logic so the audio only updates when switching scenes, when the `Play` button is pressed, or when scrubbing audio.  Essentially the update occurs just before you actually want to hear it. With this change, updating audio may cause a slight delay just before actually playing.  Note: Updates only occurs when an action flags the audio to be regenerated. Currently this happens with cell movements, but will be expanded to other cell operations in PR #1535

- Fix converting format from 32bit PCM to 32bit FLOAT

When audio is mixed, if a 32bit PCM formatted audio is converted to 32bit FLOAT, it results in no sound.  The reverse didn't have this problem.

Corrected the calculation that was using Integer division when it should have been using Floating point division.

- Update audio with Insert/Remove Frame

### DIFF
--- a/toonz/sources/common/tsound/tsop.cpp
+++ b/toonz/sources/common/tsound/tsop.cpp
@@ -747,7 +747,8 @@ void TSop::convert(TSoundTrackP &dst, const TSoundTrackP &src) {
     dst_bits  = dst->getBitPerSample();
 
     if (src_chans == dst_chans && src_bits == dst_bits)
-      if (src->isSampleSigned() == dst->isSampleSigned())
+      if (src->isSampleSigned() == dst->isSampleSigned() &&
+          src->getSampleType() == dst->getSampleType())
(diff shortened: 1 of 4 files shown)


melieconiek:
Tested this on a very long animatic with lost of audio files. Seems to work fine!

An update isn't triggered after copy/cut/paste actions or moving audio after making a region selection first.

manongjohn:
Thanks for testing.  With the OT patch merged, copy/cut/paste of cells should now trigger audio updates in the nightly.

Merging this now.  Will handle any issues related to audio updating separately, as needed.

PR #1540, opened 2024-08-03
Fix Qt::AA_UseDesktopOpenGL setting (Windows only)

manongjohn:
The Qt::AA_UseDesktopOpenGL attribute must be set before the instantiation of the application but since OT went opensource, it has always been set incorrectly afterwards.  As such T2D (and OT) have apparently been using ANGLE rather than OpenGL.

I've moved the setting of this attribute to the correct position.

I'm not sure of the full impact of this change, however when I compared against the recent nightly I did notice my preview renders of a scene composed with only vector levels were faster.

Additional testing is needed to check for any negative effects of this change.  Possible this change may affect users in different ways depending on their system.

EDIT: The change of location doesn't appear to have any impact when building on Qt5.9, still used by T2D's fix releases.  As such, this will only be applied to v1.5.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -333,6 +333,12 @@ int main(int argc, char *argv[]) {
   // constructed. Available from Qt 5.6.
   QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 
+#ifdef Q_OS_WIN
+  //	Since currently Tahoma does not work with OpenGL of software or
+  // angle,	force Qt to use desktop OpenGL
+  QApplication::setAttribute(Qt::AA_UseDesktopOpenGL, true);
+#endif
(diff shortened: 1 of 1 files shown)


manongjohn:
FYI,  I may make this a Preference setting in case people run into issues with the OpenGL on their machine.

manongjohn:
I've decided to cancel this change.

After more testing, I think the improved rendering I was seeing was due to some weirdness happening on my system at the time and now I am not seeing any difference with or without this.

After additional reading, this setting merely forces the use of OpenGL rendering over anything else.  It does not appear to be necessary, since the default behavior is to try to use Open GL rendering and only defaults back to ANGLE if OpenGL is not found or the version provided is not suitable.  I found forcing the use of ANGLE actually caused the canvas to be blank.

PR #1541, opened 2024-08-05
Clip ZeraryFx Columns

manongjohn:
This enhancement allows user to apply a Clipping Mask against any Zerary Fx (column based fx), like checkboard, gradients, clouds, etc.

Per normal fx viewing, you'll need to be in `Preview` mode in order to see the effect of the clipping mask on the fx.

### DIFF
--- a/toonz/sources/include/toonz/txshzeraryfxcolumn.h
+++ b/toonz/sources/include/toonz/txshzeraryfxcolumn.h
@@ -88,6 +88,8 @@ Return \b TZeraryColumnFx.
    * icon. */
   void updateIcon() {}
 
+  std::vector<TXshColumn *> getColumnMasks();
+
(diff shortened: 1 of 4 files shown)


Codeoloketuyi:
How can I find this feature in Tahoma

manongjohn:
Download the latest nightly.

Codeoloketuyi:
I have the latest nightly build already

manongjohn:
- Turn on Preview.
- Add a column based Fx like Checkboard
- Put a Vector column beneath it (left if in xsheet)
- Draw and fill a shape
- Convert the column into a Clipping Mask by checking the option in the column's config (Cog wheel button)

Codeoloketuyi:
But I can't seem to find it

PR #1547, opened 2024-08-09
Change eraser tools resize binding to CTRL+ALT in Normal Mode only.

melieconiek:
Change eraser tools resize binding to CTRL+ALT in Normal Mode only. 

This should help reduce accidentally triggering a resize when attempting to execute another shortcut. The CTRL+ALT resize binding is consistent with the brush tools (when enabled.)


manongjohn:
LGTM.  Thanks for this improvement!

PR #1556, opened 2024-08-27
Relative Drawing Onion Skin Mode

manongjohn:
Current Onion Skin has 2 basic modes, Fixed and Relative.  This new feature will essentially split the Relative Onion Skin into 2 modes:
- Frames (current; default)
- Drawings (new)

You can switch between these modes using in the Frames context menu like so:
image

The relative marker shapes indicate the mode you are in, Circle = Frame, Square = Drawing.

In Frame Onion Skin Mode, the relative marker will show the frame that the mark is actually on. Relative Drawing Onion Skin Mode is different  in that setting a relative skin marker will display a specific drawing before/after the current frame..  The distance away from the current frame determines which drawing away from the current frame to show, i.e. 1 away-1st drawing, 2 away-2nd drawing, etc

image

In the image, the current drawing is Drawing 4.  The selected markers means that the onion skins will be shown for drawings 1, 2, 3 and 5, 6, 7 even though they are spaced apart.

### DIFF
--- a/toonz/sources/include/toonz/onionskinmask.h
+++ b/toonz/sources/include/toonz/onionskinmask.h
@@ -46,11 +46,6 @@ public:
     ENABLED_WITHOUT_GHOST_MOVEMENTS
   };
 
-  enum RelativeSkinMode {
-    FRAMES,
-    DRAWING
-  };
-
 public:
   OnionSkinMask();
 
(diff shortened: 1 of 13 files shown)


melieconiek:
The new onion skin setting isn't restored after an application restart.

Everything else seems to work great!

melieconiek:
Vector guided tweening seems to cause a crash in this PR.

manongjohn:
Please provide a crash log or steps to follow to reproduce.  I can't seem to cause a crash with the guided tweening controls.

manongjohn:
I found the problem. I inadvertently modified something in fixed onion skin logic to look for a relative marker and it crashes if there are no relative markers set.

PR #1562, opened 2024-09-02
[WIP] Block codec DLL files that cause crashes

manongjohn:
This should prevent crashes when T2D attempts to search for an AVI codec to use and encounters specific DLLs known to cause crashes at startup:

- C:\Windows\System32\lvcod64.dll     (Logitech codec)
- C:\Windows\System32\ff_vfw.dll        (ffdshow codec)

The logic actually ignores path and only look for the actual filename.

As I cannot verify this works, this ticket will remain in WIP status until someone with the problem can verify the artifact does resolve the problem.

### DIFF
--- a/toonz/sources/image/avi/tiio_avi.cpp
+++ b/toonz/sources/image/avi/tiio_avi.cpp
@@ -19,6 +19,8 @@
 #include "timageinfo.h"
 #include "trasterimage.h"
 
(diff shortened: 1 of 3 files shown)


manongjohn:
Cancalling this. This was tested by someone who was having issues with lvcod64.dll but it didn't help. 😔

PR #1575, opened 2024-09-14
Fix pasting raster selection in xsheet/timeline cell

### LINKED ISSUE #1573: Pasting into a folder cell causes a crash
### Bug Description

After copying part of a drawing, accidentally pasting into a folder cell crashes T2D.

### Steps to reproduce

1. With the selection tool, select part of a drawing (ie. on a raster level)
2. Click on a folder cell (not a level cell)
3. Try to paste

### Relevant logs/screenshots

Screenshot 2024-09-12 at 7 21 33 PM
[folder-paste-crash.txt](https://github.com/user-attachments/files/16986651/folder-paste-crash.txt)

### Software Version

Nightly (Specify below)

9/8/24

### Operating System

macOS 14.5

### Graphics Card

_No response_

### Drawing Tablet

_No response_

manongjohn:
This fixes issues related to pasting image selections in certain xsheet/timeline cells.

This is done by selecting a part of an image and then clicking on a xsheet/timeline cell in order to either add it to that image or create a new frame with the selection. 

The issues correct are as follows:
1. This fixes #1573 
Since cells are not allowed in folder columns, this will throw an error message to the user and cancel the paste action.
2. Fixes pasting selection in a non-image cell (camera, sound, palette, mesh, etc) that causes a crash similar to the issue mentioned above
3. Fixes an issue where pasting selection fails, usually pasting raster into vector/smart raster, but creates a new frame that is not undoable
4. Block pasting selection in a sub-scene cell which clears an existing frame that cannot be undone.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -9,6 +9,7 @@
 #include "toonz/imagepainter.h"
 #include "toonz/tapplication.h"
 #include "tools/cursors.h"
+#include "tpalette.h"
 
 // TnzCore includes
 #include "tcommon.h"
@@ -549,6 +550,8 @@ transformation.
   void setAlignMethod(int method) { m_alignMethod = method; }
   int getAlignMethod() { return m_alignMethod; }
  
+  void removeTouchedImageIfNeeded(const TPaletteP &oldPalette);
+
 public:
   struct CellOps {
     int r0;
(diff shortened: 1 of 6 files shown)


artisteacher:
Thanks for such a quick fix! It works great.

Just an FYI, pasting vector and smart raster data into the folder cell silently fails. For consistency, they should probably have the error message too.

manongjohn:
> Just an FYI, pasting vector and smart raster data into the folder cell silently fails. For consistency, they should probably have the error message too.

Not possible. If you click from any non-Raster cell to a folder cell, most of the cell commands like paste are disabled. Because of this I can't display anything.  Not sure why, but when clicking from a raster cell to a folder cell, the commands become active when it should really be inactivated.  Will need to look into that.

PR #1584, opened 2024-09-26
Undo/Redo Touch Gestures

manongjohn:
Discovered that Tahoma2D has Undo/Redo Touch Gestures for touchscreens (windows) and touchpads (macs). The original method uses 3-finger swipe Left (Undo) or Right (Redo).

This will add and additional option for how to invoke Undo/Redo using a 2-finger Tap (Undo) and 3-finger Tap (Redo) matching the undo/redo gestures in other apps and can be changed in Preferences:
image

Mac Users:  The 2/3-finger tapping methods do not currently work on macOS using either touchpad or touchscreen.

PR #1597, opened 2024-10-05
[WIP] macOS 2/3-fingered Touchpad Tap Gesture

manongjohn:
This is an attempt to allow Undo/Redo actions using 2/3-finger tap gesture on macOS touchpads, similar to the one introduced in #1584 

Normally 2/3-fingered tap Undo/Redo requires Qt's Touch events, Gesture events and Mouse Press/Release events. Tap gestures are not generated on macOS touchpads, for some reason, so it works without them as follows:

- 2-fingered tap:
  - macOS touchpad automatically triggers a mouse Press and Release after a 2-fingered tap.  Logic will flag the touch as a gesture from a touchscreen to allow the existing logic to handle the Undo
- 3-fingered tap:
  - macOS touchpad doesn't do anything with this, so the logic will not only flag the touch as a gesture from a touchscreen, but also trigger a mouse Press and Release event to allow the existing logic to handle the Redo

### DIFF
--- a/toonz/sources/toonz/preferencespopup.cpp
+++ b/toonz/sources/toonz/preferencespopup.cpp
@@ -2451,12 +2451,6 @@ QWidget* PreferencesPopup::createTouchTabletPage() {
   if (winInkAvailable) insertFootNote(lay);
   widget->setLayout(lay);
 
-#ifdef MACOSX
-  // Can only support 3-finger swipe undo/redo gestures for now
-  m_controlIdMap.key(gestureUndoMethod)->setEnabled(false);
-  m_controlIdMap.key(gestureRedoMethod)->setEnabled(false);
-#endif
-
   bool ret = true;
   ret = ret && connect(touchGesturesBox, SIGNAL(clicked(bool)), touchAction,
                        SLOT(setChecked(bool)));
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -1309,6 +1309,19 @@ void SceneViewer::touchEvent(QTouchEvent *e, int type) {
         panQt(centerDelta.toPoint());
         m_touchPoints = 100;  // This will block undo/redo action
       }
+#ifdef MACOS
+      else {
+        // For macs: Might be using 2-finger touchpad tap to undo
+        if (Preferences::instance()->getGestureRedoMethod() ==
+            Preferences::TwoFingerTap) {
+          // Make it look like a gesture from touchscreen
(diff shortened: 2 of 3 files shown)


artisteacher:
So the 2/3 finger taps are registering as touch events in the log when using the touchpad (though sometimes 2 fingers is registering as a right-click), it is just that they still aren't triggering undo/redo.

manongjohn:
Please provide Viewer Event Log.

artisteacher:
Thanks for continuing to pursue this! This is just with the touchpad.

1. 3 strokes
2. 2-finger tap
3. 3-finger tap

[mac-touchpad-2-3.txt](https://github.com/user-attachments/files/17271556/mac-touchpad-2-3.txt)

manongjohn:
hmm.  Are you sure you're running the artifact from this PR and verified your settings are for 2/3-fingered tapping?

The 2 fingered tap didn't generate a Move Press/Release.  I wonder why it generates it in some cases and not others. That's problematic if it doesn't do that consistently.

The 3-fingered tap should have generated a Move Press/Release since I have the code automatically creating it, but I see no evidence of it, which is why I wonder if you ran a prior version or are still set to 3-fingered swiping.

manongjohn:
As I need more information and will likely need additional work, I've moved this to WIP. This will most likely not make v1.5 and will need to wait until v1.6.

artisteacher:
Sorry about the delay!

Here's some updated logs - this is definitely the most recent action build.
Screenshot 2024-11-10 at 6 13 09 PM

Touchpad-only:
1. 2 finger tap (often registers as mouse button press, but ocassionally not)
2. 3 finger tap
3. Pan - 2 finger drag
4.  Zoom - 2 finger pinch
5. Undo/Redo - 3 finger drag
[11-10-touchpad-only.txt](https://github.com/user-attachments/files/17693461/11-10-touchpad-only.txt)

Ipad Mirroring:

- 1, 2, and 3 finger taps don't seem to be recognized on the touchscreen at all in the Viewer Event Log
- Panning gestures are recognized as pinch gestures - which explains why the 3-finger drag doesn't currently work to undo/redo in this context
- Since taps aren't recognized, you currently have to draw in the viewer to get the right focus for any touch gestures to work

This log uses the same general steps as the previous, with some drawing thrown in to make the Viewer active. All capture events but Key Press/Key Release were enabled.
[11-10-mirrored.txt](https://github.com/user-attachments/files/17693499/11-10-mirrored.txt)

manongjohn:
Closing for now. Will revisit when I am able to get a local build environment to work on.

PR #1622, opened 2024-10-20
Allow to run packaging script from any dir

Zloi-KoteikA:
Previously you can only run packaging script when your current working directory is a root of repository.

This change allows  to run packaging script from any dir.


manongjohn:
@Zloi-KoteikA Thank you for this update.   I know I'll find it useful. :)

PR #1625, opened 2024-10-21
Fix pinching at end of line

### LINKED ISSUE #1601: Pinch tool often does not grab end of line.
### Bug Description

When attempting to grab the endpoint of a line the pinch tool will often grab just short of the endpoint. The tool cursor will change indicating it has automatically changed modes because you're grabbing an endpoint but not work as expected.

### Steps to reproduce

Draw a number of lines and attempt to grab the endpoints with the pinch tool. The bug will crop up often.

In the attached vector level the right endpoint of this line will cause the error: [pinch-tool-endpoint-error.zip](https://github.com/user-attachments/files/17270321/pinch-tool-endpoint-error.zip)

https://github.com/user-attachments/assets/cab71e87-26d0-4647-96cb-b5c376bdf489

### Relevant logs/screenshots

_No response_

### Software Version

Nightly (Specify below)

2024-10-05

### Operating System

Windows 10

### Graphics Card

_No response_

### Drawing Tablet

_No response_

manongjohn:
This fixes #1601 

Due to an infinitesimal fractional calculation difference at the end of some lines, the Pinch tool thinks it's not quite at the end and so pinches the line really close to the end instead of the end itself

Added check that if we're close enough to the end, we'll treat it as the end and pinch accordingly.

### DIFF
--- a/toonz/sources/tnzext/LinearPotential.cpp
+++ b/toonz/sources/tnzext/LinearPotential.cpp
@@ -22,7 +22,10 @@ void ToonzExt::LinearPotential::setParameters_(const TStroke *ref, double par,
   leftFactor_ = lengthAtParam_;
 
   // lunghezza dal pto di click alla fine
-  rightFactor_ = strokeLength_ - lengthAtParam_;
+  if (areAlmostEqual(strokeLength_, lengthAtParam_, 0.001))
+    rightFactor_ = 0.0;
+  else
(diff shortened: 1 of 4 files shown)


melieconiek:
If you hold down CTRL or ALT to switch modes the bug will still be present.

manongjohn:
> If you hold down CTRL or ALT to switch modes the bug will still be present.

Fixed

PR #1656, opened 2024-11-24
Fix edit missing controlpoint crash

manongjohn:
Not sure what causes it to happen, but a stroke can be created with a missing end control point as shown here:
image
 (left - incomplete stroke isolated from a user's image, right - corrected stroke)

A crash would occur with the `Control Point Editor Tool` whenever you clicked between the last control point and the rest of the stroke.  This change prevents the crash by not allowing the selection of any part of the stroke beyond the last control point (segment or handle).

Sample image: [badcp.zip](https://github.com/user-attachments/files/17894470/badcp.zip)

### DIFF
--- a/toonz/sources/tnztools/controlpointselection.cpp
+++ b/toonz/sources/tnztools/controlpointselection.cpp
@@ -971,7 +971,7 @@ ControlPointEditorStroke::PointType ControlPointEditorStroke::getPointTypeAt(
 
   if (minDistance2 < distance2)
     distance2 = minDistance2;
-  else if (strokeDistance > distance2) {
+  else if (strokeDistance > distance2 || nextIndex(precPointIndex) == -1) {
     distance2 = strokeDistance;
     index     = -1;
     type      = NONE;


darkhog:
Not sure if I understand, but this just prevents Tahoma from crashing and not restore missing end point?

manongjohn:
> Not sure if I understand, but this just prevents Tahoma from crashing and not restore missing end point?

Correct. This is just a stopgap to prevent a crash. I'd rather it not be selectable than crash.

Not quite sure how to fix the control point, yet.  Normally, further line manipulation readjusts the line and removes the extra length at the end.

darkhog:
It's not "extra length", it's just how the line is supposed to be like, just missing the end point (but somehow having the bend handle?).

An "ugly hack," but working, solution would be on every modification to the path (such as after dragging the point) iterate over the points and check if the end point is there or not. If there isn't, create the endpoint and link it to the "phantom" bend handle. Again, would be an ugly hack and wouldn't solve the root of the issue (end points going missing in the first place), but should work.

manongjohn:
Forgive my rather general comment of "extra length" as I was not intending on getting into a lot of specifics regarding it.  In short, the internal datapoints for the stroke falls short of what is expected.  While displaying control points is not an issue, iterating the datapoints to find out what control point a mouse pointer is trying to capture is a problem due to certain assumptions in the code.

I did look at trying to detect this scenario and add the missing control point automatically, but in the interest of time with how close I was to the 1.5 release, I just wanted to put in a quick stopgap to prevent the crash. I do want to eventually go back and find a proper way to detect and fix a missing control point.

It did also cross my mind that the logic for simplifying control points on a line may be causing this.

PR #1660, opened 2024-12-04
High DPI Scaling preference

manongjohn:
This adds a new `Interface` preference setting `Enable High DPI Scaling`.  This setting is enabled by default.

T2D has always forced High DPI Scaling, however, on lower resolutions (i.e. 1920x1080 at 150%), this causes the UI to be extremely huge.

I've exposed this setting as a preference item for those who have the huge UI issue.  This way they can disable the DPI scaling and bring the UI down to a size appropriate for their lower resolution.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -252,6 +252,10 @@ public:
     return getBoolValue(viewerIndicatorEnabled);
   }
 
+  bool isHighDpiScalingEnabled() const {
+    return getBoolValue(highDpiScalingEnabled);
+  }
+
   // Visualization  tab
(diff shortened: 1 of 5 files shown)


artisteacher:
That should help a lot of people!

manongjohn:
Merging. 

I've decided to move this into the 1st 1.5 Fix release (1.5.1 - Feb).  This is in part because the Qt attributes `Qt::AA_EnableHighDpiScaling` and `Qt::AA_DisableHighDpiScaling` have no effect in Qt6 due to improved High Dpi support. As such, this change/option will be pointless if the next release migrates to Qt6.  At least 1.5.x users can benefit for a while.

PR #1661, opened 2024-12-05
Fix relative drawing onion skin guided drawing

manongjohn:
This fixes an issue with Vector Guided Drawing while using Relative Drawing onion skin.

When using this new onion skin mode, guided drawing worked fine when all the frames were on 1s.  Fixed the logic to account for held cells, implicit or explicit.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -731,7 +731,8 @@ public:
     invalidateAll();
   }
 
-  void getGuidedFrameIdx(int *backIdx, int *frontIdx);
+  void getGuidedFrameIdx(TXsheet *xsh, int row, int col, int *backIdx,
+                         int *frontIdx);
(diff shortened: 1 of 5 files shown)


motionmonster:
In the short term, I believe that there is a place in stopmotioncontroller that looks for Canon cameras and starts the ball rolling to call the open session command.  With T2D using libgphoto2 as well, we could temporarily bypass the Canon call and just use libgphoto for now.

manongjohn:
> In the short term, I believe that there is a place in stopmotioncontroller that looks for Canon cameras and starts the ball rolling to call the open session command. With T2D using libgphoto2 as well, we could temporarily bypass the Canon call and just use libgphoto for now.

I think this is in the wrong place

motionmonster:
Oops, sorry. . . my train of thought was going and didn't think about what page I was on.

PR #1666, opened 2024-12-08
Column Alpha Lock

manongjohn:
This adds a new column setting `Alpha Lock` to any Raster, Smart Raster and Vector column.

image

This is similar to Clipping Masks, where the alpha locked column will only show where parts of the referenced drawing is visible.

image

How to use:
- The reference column is below (timeline)/before (xsheet) the Alpha Locked column.
- Other Alpha Locked columns and Mesh Columns between the Alpha Locked column and the reference column are ignored.
- Multiple Alpha Locked columns can be applied to 1 reference column.
- If there are any other column types below/before the Alpha Locked column, the reference column is hiddne or, there is no exposed cell in the reference column, the alpha locked cell will not be displayed.
- Alpha Lock can be applied to Raster, Smart raster, Vector and, Sub-Scene columns.
- A column can only have `Alpha Lock` or `Clipping Mask` enabled, not both.

### DIFF
--- a/toonz/sources/common/tfx/binaryFx.cpp
+++ b/toonz/sources/common/tfx/binaryFx.cpp
@@ -696,11 +696,13 @@ class ColumnMaskFx final : public TBaseRasterFx {
   FX_DECLARATION(ColumnMaskFx)
 
   TRasterFxPort m_source, m_mask;
+  TBoolParamP m_isAlphaLockMask;
 
 public:
-  ColumnMaskFx() {
+  ColumnMaskFx() : m_isAlphaLockMask(false) {
     addInputPort("Source", m_source);
     addInputPort("Mask", m_mask);
+    bindParam(this, "isAlphaLockMask", m_isAlphaLockMask);
     setName(L"ColumnMaskFx");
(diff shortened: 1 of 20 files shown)


papio-pl:
Great feature. It works fine in viewport but I've noticed it renders only first cropped layer.

https://github.com/user-attachments/assets/369cdf5f-47d6-487a-a499-305fe5462d44

manongjohn:
@papio-pl Thanks for pointing that bug out. I've fixed it with the latest commit.

artisteacher:
Should the opacity of the column itself make a difference?

manongjohn:
> Should the opacity of the column itself make a difference?

No.  This uses the Clipping Mask logic to confine what is rendered to wherever the reference drawing has a pixel with an Alpha > 0. It's an all or none situation.

If anything, the Clipping Mask may be enhanced to factor in the Opacity Filter so it can adouble as an Opacity Mask when the Opacity Filter is < 100%.  This would make more sense once the other PR goes in to change the Clipping Mask stacking from bottom to top.

PR #1667, opened 2024-12-08
Reverse Clipping Mask Stacking Placement

manongjohn:
To follow more in line with the logical stacking order (and other software), this PR changes the placement and stacking order of Clipping Masks.

Clipping Masks now need to be placed **above** (timeline)/**right** (xsheet) of the clipped column.
Clipping is done in order from bottom->top/left->right.

image

Note: Due to change in stacking order, `Render Mask` has been removed.

### DIFF
--- a/toonz/sources/include/toonz/stageplayer.h
+++ b/toonz/sources/include/toonz/stageplayer.h
@@ -121,7 +121,6 @@ public:
 
   bool m_isMask;
   bool m_isInvertedMask;
-  bool m_canRenderMask;
 
   bool m_isAlphaLocked;
 
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -85,7 +85,7 @@ protected:
     eMasked                = 0x10,
     eCamstandTransparent43 = 0x20,  // obsoleto, solo per retrocompatibilita'
     eInvertedMask          = 0x80,
-    eRenderMask            = 0x100,
+    eRenderMask            = 0x100, // Obsolete but don't remove or it may impact older scens
     eAlphaLocked           = 0x200
   };
 
@@ -204,7 +204,6 @@ Return true if column is a mask.
 */
   bool isMask() const;
   bool isInvertedMask() const;
-  bool canRenderMask() const;
 
   bool isAlphaLocked() const;
   /*!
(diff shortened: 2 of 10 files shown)


artisteacher:
I think this definitely will make it easier going back & forth between different programs.

Is the "render mask" option working as expected?

manongjohn:
> Is the "render mask" option working as expected?

It should be. Was working for me in my tests.

I suppose the difference here is that when the clipping masks were below the clipped column, the rendered mask was below it.  Now it's rendered above it.

artisteacher:
So I have 2 levels.  This vector star above:
Screenshot 2024-12-10 at 10 06 57 PM
And this raster image below:
Screenshot 2024-12-10 at 10 06 49 PM

With the clipping mask enabled + render mask, the mask completely obscures the clipped area:
Screenshot 2024-12-10 at 10 06 21 PM
Which makes the raster level pointless. If you keep the render mask option, shouldn't you be able to see both?

With the clipping mask inverted + render mask, the mask isn't visible:
Screenshot 2024-12-10 at 10 06 02 PM
Though if it were, I guess there might not be a point in using the mask to begin with...

manongjohn:
@artisteacher Render mask was only visible when using Invert, even before this change.  Seems to be a bug with invert and render.  Odd because I know I had that working at one point when I was testing the reversal.   Will fix

manongjohn:
Upon further reflection, I believe you are correct and `Render Mask` is no longer necessary.

Clipping + Render Mask = essentially hides the image below it behind the drawn mask.
Clipping + Invert Mask + Render Mask = the same as just drawing the mask on top of the image without clipping.

Unless someone else disagrees with this, I can completely remove the `Render Mask` option.

artisteacher:
I think it could still be helpful for a default clipping mask, as long as the mask displays under the clipped area. Completely agree on removing for an inverted mask.

manongjohn:
> I think it could still be helpful for a default clipping mask, as long as the mask displays under the clipped area. It's the equivalent of the "visible matte in" fx. Completely agree on removing for an inverted mask.

Displaying a mask under the clipped area breaks the stacking order of things and complicates the rendering to a degree. They can just as easily accomplish this by placing a copy of the column, without mask enabled, below the clipped column.

artisteacher:
Was just hoping to save a step or 2, but that would certainly work too.

manongjohn:
Updated and slightly fixed PR due to conflicts with the Alpha Lock PR after it was merged.  With that, updated Alpha Locked columns to ignore clipping masks found between it and the reference column.

I'll leave this up a few extra days in case anyone wants to retest.

artisteacher:
Working just fine on my end.

senoworks:
thats how it should be the clip must be on the top, like software painting did

PR #1669, opened 2024-12-14
Column Looping

manongjohn:
This adds a new column property `Loop Frames` to Raster, Smart Raster, Vector and, Sub-Scene columns.

image

When enabled, all the frames between the 1st exposed frame and the last exposed frame will be repeated automatically.

image

By default, it will loop the frame sequence infinitely.  You can select to stop the loop at any time by enabling `Loop To Frame` and specifying the last frame to stop the loop.

Additional Notes:
- Looping begins wherever the 1st frame in the sequence is positioned.
- If the last frame in the sequence is a `Stop Frame`, it is ignored. The looping will actually begin on the last stop frame.
- When enabled, looped cells are uncolored (vs tinted for implicit holds) and will show frame numbers/stop frames.
- A looping icon will appear in the thumbnail and will take on the filter color if there is one.
- `Next Drawing` and `Previous Drawing` commands will loop around the original sequence only.
- Looped cells can be onion skinned, though currently no actions can be done with them.

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -147,7 +147,8 @@ enum class PredefinedRect {
   CLIPPING_MASK_AREA,
   FOLDER_INDICATOR_AREA,
   FOLDER_TOGGLE_ICON,
-  BUTTONS_AREA
+  BUTTONS_AREA,
+  LOOP_INDICATOR_AREA
 };
 enum class PredefinedLine {
   LOCKED,              //! dotted vertical line when cell is locked
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -73,6 +73,9 @@ class DVAPI TXshColumn : public TColumnHeader, public TPersist {
   QStack<int> m_folderId;
   int m_folderSelector;
 
+  bool m_loopLimitEnabled;
+  int m_loopToFrame;
+
 public:
 private:
   int m_colorFilterId;
@@ -86,7 +89,8 @@ protected:
     eCamstandTransparent43 = 0x20,  // obsoleto, solo per retrocompatibilita'
     eInvertedMask          = 0x80,
     eRenderMask            = 0x100,
-    eAlphaLocked           = 0x200
+    eAlphaLocked           = 0x200,
+    eLooped                = 0x800
   };
 
   TRaster32P m_icon;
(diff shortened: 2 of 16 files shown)


artisteacher:
Just tried out a simple test & this is working great so far! Will spend some more time experimenting over the weekend.

What do you think about automatically updating the stop marker to match the "Loop to Frame" value, if the playback range doesn't already extend that far?

artisteacher:
Ooh, what about adding ping-pong?

manongjohn:
> What do you think about automatically updating the stop marker to match the "Loop to Frame" value, if the playback range doesn't already extend that far?

Wouldnt be good if you are looping for a bit in the middle of the scene but the scene is longer.

> Ooh, what about adding ping-pong?

As intriguing as that idea is, i think it's already easy enough to ping-pong (`Swing`) a sequence.  It would just imbed it in the main sequence of the looped column. I dont want to add too much to the menu.  Starting to look cluttered.

pedrobbs00:
That's great! Although, maybe instead of just being there as a column property, it could be more accessible similarly like TVPaint. By right clicking the last frame / after the last frame and choosing a "Frame Looping Behaviour".
image
Also, this is looping the frames non-destructively, correct? Because the Ping-Pong behaviour we have right now is "destructive", meaning that it directly copies the frames we have and inverts them at the end and if we change the animation this inversion is not affected. And if we wanted the Ping-Pong behaviour to be continuously looping, we need to first "Swing" the frames, then loop them. In TVPaint, for example, the Ping-Pong behaviour refers to continuously ping-pong the frames, not just once. I think this approach would be very interesting keeping the "Loop to Frame:" structure controlling for how long any frame looping behaviour would be held to.
I made a small mock-up to help visualise it:

Here it would be as default with no Looping.
ex1

Here the normal Loop behaviour has been toggled, similarly as the "Aut-Creation" and other toggles.
ex2

papio-pl:
Works fine so far. Very useful for me. Thank you.

manongjohn:
> That's great! Although, maybe instead of just being there as a column property, it could be more accessible similarly like TVPaint. By right clicking the last frame / after the last frame and choosing a "Frame Looping Behaviour".

Thanks for the suggestion. It's along the lines of what I originally would have liked to have done.  My original idea was more robust and versatile but is significantly more complicated from a development standpoint and i had some concerns on how best to handle some of the UI/X aspects of the feature.

This was a rather quick solution. Forgive my laziness. 🥴

Having said that, my original idea is not entiely off the table.

melieconiek:
At the end of the loop forward drawings don't seem to show up in the relative onion skin.

manongjohn:
Cancelling this as I have developed an alternate cell-selection based looping feature which I will be submitting shortly.

PR #1673, opened 2024-12-18
Fix Gap Closure results (Tape Tool)

manongjohn:
This PR attempts to filter the "noise" of the results using the Tape Tool in `Rectangle` mode and View Gap Check.

The logic for what gap segments has been updated.  The results of the list of possible closing segments are filtered as follows:
- All line-2-line segments are discarded.
- Any endpoint-2-line or line-2-endpoint segments that form an acute angle < 45 degrees are discarded.
- Any gap segments that cross other existing lines are discarded.
- The shortest allowable segment from an endpoint to another endpoint or line is kept.
- endpoint-2-endpoint segments are favored over shorter endpoint-2-line segments if the endpoint-2-line segment is close enough to the end of the stroke.

Additionally:
- The Fill Tool's `Maximum Gap` toolbar option has been renamed to `Auto Close Gap`. It not longer influences the Tape Tool's rectangular mode.
- Added a `Min` distance setting to the Tape Tool's rectangular mode.  Default value is 1.15 (the default for Auto Close Gap).

Tape tool example before and after results using `Distance` = 100
image

### DIFF
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -1836,20 +1836,20 @@ void getClosingSegments(TL2LAutocloser &l2lautocloser, double facMin,
 
   double w;
   if (isCloseEnoughP2L(facMin, facMax, s1, 0.0, s2, w, forAutoClose) &&
-      w != 0.0 && w != 1.0)
+      (forAutoClose || (w != 0.0 && w != 1.0)))
     segments.push_back(std::pair<double, double>(0.0, w));
 
   if (isCloseEnoughP2L(facMin, facMax, s2, 1.0, s1, w, forAutoClose) &&
-      w != 0.0 && w != 1.0)
+      (forAutoClose || (w != 0.0 && w != 1.0)))
     segments.push_back(std::pair<double, double>(w, 1.0));
 
   if (s1 != s2) {
     if (isCloseEnoughP2L(facMin, facMax, s2, 0.0, s1, w, forAutoClose) &&
-        w != 0.0 && w != 1.0)
+        (forAutoClose || (w != 0.0 && w != 1.0)))
       segments.push_back(std::pair<double, double>(w, 0.0));
 
     if (isCloseEnoughP2L(facMin, facMax, s1, 1.0, s2, w, forAutoClose) &&
-        w != 0.0 && w != 1.0)
+        (forAutoClose || (w != 0.0 && w != 1.0)))
       segments.push_back(std::pair<double, double>(1.0, w));
   }
 }
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -1862,8 +1862,10 @@ struct SegmentData {
   int startIdx;
   double startValue;
   bool isStartStrokeStart;
+  bool isStartSelfLoop;
   int endIdx;
   bool isEndStrokeStart;
+  bool isEndSelfLoop;
   double endValue;
   double distance;
   int segType;  // 1 = e2e, 2 = e2l, 3 = l2l
@@ -1884,12 +1886,21 @@ void addToSegmentList(std::vector<std::pair<double, double>> segments,
     TPointD p1 = s1->getPoint(segments[k].first);
     TPointD p2 = s2->getPoint(segments[k].second);
 
+    if (!rect.isEmpty() && (!rect.contains(p1) || !rect.contains(p2))) continue;
+
+    if (segmentAlreadyPresent(strokeList, p1, p2)) continue;
+
     bool isP1EndPt = !s1->isSelfLoop() &&
                      (segments[k].first == 0.0 || segments[k].first == 1.0);
     bool isP1StrokeStart = !s1->isSelfLoop() && segments[k].first == 0.0;
+    bool isP1SelfLoop    = s1->isSelfLoop();
     bool isP2EndPt       = !s2->isSelfLoop() &&
                      (segments[k].second == 0.0 || segments[k].second == 1.0);
     bool isP2StrokeStart = !s2->isSelfLoop() && segments[k].second == 0.0;
+    bool isP2SelfLoop    = s2->isSelfLoop();
+
+    // Discard line-2-line segments
+    if ((!isP1EndPt && !isP2EndPt)) continue;
 
     std::vector<TPointD> pts;
     pts.push_back(p1);
@@ -1930,9 +1941,6 @@ void addToSegmentList(std::vector<std::pair<double, double>> segments,
     double angle =
         m2angle > m1angle ? (m2angle - m1angle) : (m1angle - m2angle);
 
-    // Discard line-2-line segments
-    if ((!isP1EndPt && !isP2EndPt)) continue;
-
     // Discard segments that create acute angles that make no sense
     // Only apply to end-2-line or line-2-end segments
     if (!(isP1EndPt && isP2EndPt) && (angle < 45.0 || angle > 315.0)) continue;
@@ -1944,22 +1952,17 @@ void addToSegmentList(std::vector<std::pair<double, double>> segments,
       TStroke *testStroke = strokeList[z]->m_s;
       int y = intersect(segStroke, testStroke, intersections, false);
       for (int a = 0; a < intersections.size(); a++) {
-        if (intersections[a].first == 0.0 || intersections[a].second == 0.0 ||
-            intersections[a].first == 1.0 || intersections[a].second == 1.0)
-          y--;
-      }
-      if ((y >= 1 && z != i && z != j) || (i == j && y > 2) ||
-          (y >= 2 && i != j && (z == i || z == j))) {
+        if (areAlmostEqual(0.0, intersections[a].first, 0.0001) ||
+            areAlmostEqual(0.0, intersections[a].second, 0.0001) ||
+            areAlmostEqual(intersections[a].first, 1.0, 0.0001) ||
+            areAlmostEqual(intersections[a].second, 1.0, 0.0001))
+          continue;
         intersectsOther = true;
         break;
       }
     }
     delete segStroke;
     if (intersectsOther) continue;
-
-    if (!rect.isEmpty() && (!rect.contains(p1) || !rect.contains(p2))) continue;
-
-    if (segmentAlreadyPresent(strokeList, p1, p2)) continue;
 
     // If line-2-end, normalize to end-2-line
     SegmentData data;
@@ -1968,12 +1971,16 @@ void addToSegmentList(std::vector<std::pair<double, double>> segments,
         (!isP1EndPt && isP2EndPt) ? segments[k].second : segments[k].first;
     data.isStartStrokeStart =
         (!isP1EndPt && isP2EndPt) ? isP2StrokeStart : isP1StrokeStart;
+    data.isStartSelfLoop =
+        (!isP1EndPt && isP2EndPt) ? isP2SelfLoop : isP1SelfLoop;
 
     data.endIdx = (!isP1EndPt && isP2EndPt) ? i : j;
     data.endValue =
         (!isP1EndPt && isP2EndPt) ? segments[k].first : segments[k].second;
     data.isEndStrokeStart =
         (!isP1EndPt && isP2EndPt) ? isP1StrokeStart : isP2StrokeStart;
+    data.isEndSelfLoop =
+        (!isP1EndPt && isP2EndPt) ? isP1SelfLoop : isP2SelfLoop;
 
     data.distance = tdistance(p1, p2);
     data.segType  = (isP1EndPt && isP2EndPt)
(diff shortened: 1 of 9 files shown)


melieconiek:
It's definitely better. Even if it leaves a few undesired closures, it's very much usable now.

There seem to be some instances where's doubling back across lines even though those should be discarded.

I know you haven't looked at this part of the tool, but there are a lot of instances where it jumps to another point or line when there's something else much closer. This causes it to not close some regions. This is much more common than I realized.

You can see some examples below where instead of bridging the obvious gap it jumps backwards across the line or to another point, even if it's farther away.

image
image
image
image

manongjohn:
I think a number of those issues are related to the original logic that generates all the gap segments. 

Even in my test image, I noticed at least 1 case where it should have created at most 3 segments from the same endpoint but only 2 were create. The 3rd one that was not created would have been the shortest segment.

I also noticed that if the endpoint goes beyond the midline of the stroke, it may attempt to connect to other lines beyond it rather than treat it as being connected.

If possible, can you isolate all the strokes involved in your images, above, into a new vector file and send to me/attach it here?  I dont need the whole image, just the strokes involved showing the issues so i can see what's being generated and displayed.

Maybe i can tweek the logic a bit to fix some of this.

manongjohn:
Latest push revises my original logic which helps to prioritize the shorts gaps over longer ones better and also does a better job of limiting endpoints to 1 gap closure.

It will solve most of the issues that were pointed out, however, there are still underlying issues with end-2-line gap detection.   There are some end-2-line gaps that are not being found by the current logic.   This results in obvious (to us) end-2-line gaps that will not close automatically. 

The case where a segment appears to be going backwards along the stroke, it's actually closing the beginning and ending of the same stroke. Not sure why it does that in some cases but not others yet either.

Need to look further into at least the missing end-2-line issue.  Moving this PR to draft until I can sort it out, but feel free to play with the artifacts in the meantime.

manongjohn:
Just realized this also reduces the noise when using `View` -> `Gap Check`, though the lines are currently hard to see.

melieconiek:
Attached an example of some issues with the latest. I ran into many instances where fills no longer worked when they worked fine in the stable release. In the instance I included there appears to be an issue where two endpoints are right on top of each other.

Thank you for working on this.

image

[GAP-TEST_12-24-24.zip](https://github.com/user-attachments/files/18240962/GAP-TEST_12-24-24.zip)

manongjohn:
For the 3 gap issues that would not close, I was able to close them even when `Distance` was 10.

Will look to see why the one shape isn't filling.  I suspect it's because I am not applying the logic used to ensure 1 gap closure per endpoint and if there are too many at that upper left point, it's probably confusing the region determination logic.

manongjohn:
> For the 3 gap issues that would not close, I was able to close them even when `Distance` was 10.

Correction...the one issue where it doesn't close above 35 (middle left example) is a bit of a bug.  If you limit the select area to not include the stroke endpoints, it will close.  Need to look at why it is behaving that way.

Sometimes selecting a wide area to close may bring back more possible gap closing segments which will influence the decision on which one to ultimately take.  I can only apply so much logic to this.

melieconiek:
> Correction...the one issue where it doesn't close above 35 (middle left example) is a bit of a bug. If you limit the select area to not include the stroke endpoints, it will close. Need to look at why it is behaving that way.

Yes, that's what I'm seeing then. Excluding the problem line from the selection works same as deleting it. I was trying to apply gap closures to the entire drawing and looking at the gap check to see how it would behave. I didn't know the selection area mattered.

I've also noticed that if you have two points right on top of each other, it will not join them when you have `Join Vectors` enabled. The easiest way to test this is to cut a line with the cutter tool then try to join it again with the tape tool in `Rectangular` mode.

I've experienced a few crashes when saving in the latest for this PR. I don't think it generated a crash log. I haven't experienced saving crashes in stable, not recently anyways.

melieconiek:
This is what I am seeing. There's something nearby causing the gap to not close properly even though the expected gap closure should still be the shortest. Excluding the problem from the selection works.

https://github.com/user-attachments/assets/c8af92f9-e33e-42bf-9289-ee32d6f7eae7

https://github.com/user-attachments/assets/740f670f-eaa3-463e-bf43-bb0add1350f6

melieconiek:
A little more detail -- the few times when I experienced a save crash it was on a scene that had been created in stable.

manongjohn:
Let me know how the latest changes fares.  Slight improvement to the Fill Tool's `Maximum Gap` logic as well.

manongjohn:
Cleaned up logic and increased the sensitivity of gap detection, similar to what auto gap does in the Fill Tool. Tape tool now closes the tiniest of gaps, completely ignoring what might be auto closed.  This can result in very tiny strokes.

Additionally fixed an issue with intersection determination logic when a small enough gap was closed it prevented fill from working.

I was able to determine that the general vector logic appears to be finding stroke intersections correctly, but as I have stated in the past, the logic for detecting fillable regions is a bit flawed and needs to be fixed.  This PR may bring this issue more to light, so I am not sure if I can separate the Fill region detection into its own PR or if it must be fixed with this one.  Any feedback while testing this will help.

melieconiek:
> Tape tool now closes the tiniest of gaps, completely ignoring what might be auto closed. This can result in very tiny strokes.

If this was changed due to my comment about it not connecting two points for overlapping points, it might be best if the minimum gap size is restored. In that case I was trying to merge those lines back together. You wouldn't want the tape tool creating a bunch of tiny, unnecessary gap closures otherwise. Maybe if it only ignored the minimum gap size when `Join Vectors` was enabled.

I probably just need to get used to how the rectangle mode tape tool works or switch back to point-to-point mode when needed instead.

---

I realized a bit into testing that I was running into issues on drawings done in stable. Loading them into this PR build caused them to sometimes not fill correctly. Editing the problematic intersections and then *undoing* the change fixed the fill issue. I thought maybe this problem was isolated to old drawings but I was able to reproduce this issue a few times with fresh drawings done in this PR. Editing the problem line and then undoing the change would fix the fill issue.

You can replicate this issue with the drawing file I sent for the last version. When first opened the regions will not fill as expected.
1. Attempt to fill the regions (should not work correctly.)
2. Move the line work around with the selection tool.
3. Undo your edits.
4. Attempt to fill again (it should work correctly now.)

This is a new bug. If you're having trouble replicating it I will send a video later.

---

Gap closure seems to work ok except in some instances where three lines meet. I mentioned this one before. Sometimes doing second pass on it will close the gap but not always.

I'll have to try using it on an actual project to test this any more. I'll try to do this later this week. Thank you for working on this!

manongjohn:
Auto gap, used by Fill Tool technically closes even the tiniest of gaps, but you can't see when it does this.  The problem is the old Tape Tool uses that setting without you realizing it.  This may result in gaps not appearing to be taped when you are explicitly trying to join them with the rectangle option because it is technically covered by Auto Gap. 

The only solution to avoid that confusion and prevent tiny strokes is to expose the Fill Tool's `Maximum Gap` setting on the Tape Tool as well.  For consistency, I'd probably rename Fill Tool's `Maximum Gap` to `Auto Gap Distance` or something.

-----

Seems like I will need to fix the Fill Region determination logic with this PR then.  The old Fill Tool would walk paths to find closed areas but can sometimes go off in the wrong direction and either get stuck in a dead end, causing an area not to fill , or it inadvertently combines multiple areas as 1 fill area. The flaw in the logic is it sometimes doesn't know to go back to a prior decision point and try something else.

With the old Fill tool, it created so many unseen gap closures (recall the nightmare of a web the rectangular tape tool creates?) that it managed to find a path even if it wasn't the most optimal path.  This would explain why some areas not fully filling because the invisible gap cut across it. Now with the new gap closing logic, only direct gap closures that make sense are actually made, but now it doesn't have a fallback path to follow if it gets confused/lost.

-----

Trying to connect multiple points with 1 closing segment is a bit problematic right now. So far, attempts to correct for it results in the spidering effect re-appearing or odd gap closures that make no sense.  Will still try find a way, but it's not a priority for me atm.

Rectangle mode is influenced by what strokes endpoints and potential gaps are in it.  It is also somewhat oddly influenced by the stroke order.  In the one example of the 3 strokes provided, I would get different results depending on which 2 strokes I tried to close 1st.

artisteacher:
It will be amazing for fills to be improved too. One issue is that the fills seem to keep recalculating. Moving a shape with a fill to another area, shouldn't impact the fill, but sometimes it does.

melieconiek:
> The only solution to avoid that confusion and prevent tiny strokes is to expose the Fill Tool's `Maximum Gap` setting on the Tape Tool as well. For consistency, I'd probably rename Fill Tool's `Maximum Gap` to `Auto Gap Distance` or something.

As long as we can re-enable it using the fill tool's max gap as the tape tool's minimum so we don't get a bunch of micro lines when we don't want or need them. These things were problematic when I was playing with this yesterday. Unless you're trying to join some lines there isn't any purpose in closing a gap that will already be closed automatically when filled.

It wouldn't be possible to isolate the changes to the tape tool and release that only? What you have now (sans micro gap closures) works quite well.

manongjohn:
I tried to revert Fill Tool changes so only Tape tool was impacted, however, both tools use a couple of the same procedures that I had to update for the Tape Tool.  I can try separating it again, even if it means duplicating stuff. Was trying to avoid that but may take me some time to get the fill region determination logic worked out.

melieconiek:
> I tried to revert Fill Tool changes so only Tape tool was impacted, however, both tools use a couple of the same procedures that I had to update for the Tape Tool. I can try separating it again, even if it means duplicating stuff. Was trying to avoid that but may take me some time to get the fill region determination logic worked out.

Whatever you think would work best. Looking forward to all of this, but would love to just be able to close some gaps for the time being. Happy to continue testing in the mean time. Whatever will help.

manongjohn:
Ok. I've done the following
1. Reverted the auto close changes I made to the Fill tool so it behaves like it used to.  New logic is saved for future use once the fill region detection logic is fixed.
2. The Fill Tool's `Maximum Gap` toolbar option has been renamed to `Auto Close Gap`.
3. The Fill Tool's `Auto Close Gap` setting will now cause the Tape Tool (Rectangular Mode) to filter out segments that are normally auto closed.

Please keep in mind the `Auto Close Gap` is per frame, default being 1.15.

melieconiek:
Small gaps that should be bridged at the current `Auto Close Gap` setting are not closing. I have to set the value higher than in stable to get them to close. The attached example will fill with the 1.15 setting in stable but not in this PR. Increasing the setting will cause it to close.

[_GAP-TEST_1-18-25.zip](https://github.com/user-attachments/files/18464594/_GAP-TEST_1-18-25.zip)

melieconiek:
I'll be able to spend more time with this next week, but it's working well so far.

One problem. If you hold down the keybinding for the tape tool it'll lock one corner of the box to the center of the table for some reason. It shouldn't be doing that.

A lasso mode would be a lot easier to use, btw. That's a wishlist item though.

Looking forward to the fill fixes on top of this!

melieconiek:
I'm still getting some instances where the rectangle tape tool isn't closing very small gaps. The fill tool will not close these gaps either. Adjusting the fill tool gap setting down will fix the tape tool.

This isn't working correctly, but it might be nicer if the tape tool had it's own lower setting instead anyways.

[1-24-25_gap-test.zip](https://github.com/user-attachments/files/18540511/1-24-25_gap-test.zip)

manongjohn:
Yes, I've been playing around with the changes was seeing that also and found it frustrating.  I have gone back and reworked the filtering logic trying to account for some things that were being filter out. It would result in small segments that cannot be closed without lowering the Fill Tool's Auto Gap setting due to other small gaps between multiple strokes in the same area.

The current iteration will ensure there is a closing segment from a strokes endpoint to the nearest stroke.  This can result in multiple closing segments from different strokes terminating at the same stroke.   The caveat to this is if you set the maximum gap to a high value, you may get a "webbing" effect, though not as extreme as before. 

Since The View -> Gap Check uses the same exact logic as Tape Tool, you can use it to get an idea of what will be closed before using the tool. Adjust the distance to minimize the webbing effect.

I will add a new option for Minimum Gap distance that is unrelated to the Fill Tool's Auto gap logic.

manongjohn:
FYI, I still plan on including this in the next fix release (Feb) barring any major issues/concerns.

Will plan on merging by the end of this week unless something comes up.

melieconiek:
I was able to use this for some cleanup work and ran into a few things.

There seems to be an issue with very long lines. Instead of closing to the line it will close to the closest endpoint of it. Problem intersections are circled in white.
[__1-28-25_GAP-TEST.zip](https://github.com/user-attachments/files/18578707/__1-28-25_GAP-TEST.zip)

The check gap lines are a bit too thick. When you're drawing with 1px lines it can be hard to see what you're looking at. The gap check lines will often obscure linework and other, nearby gap check lines.

This ones a minor issue -- it seems like each endpoint is forced to have a unique connection even if it isn't the shortest. You can see this at three point intersections. You'd expect two of the endpoints to share one of the connections and the odd one out to get it's own but there are almost always three connections. I wanted to mention it but trying to change this might not work as well as I'm imagining -- and it's working great otherwise so maybe we just leave it.

The only other thing on my wishlist would be a lasso mode. It'd be more accurate that way. But, really, this is great without it.

**This is working a lot better! It closes everything you'd expected to with few unneeded connections. If I'm closing the whole drawing at once I'll just use 0 thickness lines. The extra connections don't matter -- what's important is that everything can be filled without issue. I don't have to convert to smart raster to to do fills. This has already saved me a ton of time. Any other issues I've encountered were the same old fill bugs. Nothing to do with the tape tool.**

manongjohn:
> There seems to be an issue with very long lines. Instead of closing to the line it will close to the closest endpoint of it.

Seems the information I use to determine proximity to end of the line doesn't quite reflect the actual distance to the end.   I have a small change that I will push in a bit that will check to see if the segment created by changing it from end-2-line to end-2-end creates a segment that is over 2x longer than the original end-2-line segment.  If so it won't move it. Otherwise it will change it.

> The check gap lines are a bit too thick. When you're drawing with 1px lines it can be hard to see what you're looking at. The gap check lines will often obscure linework and other, nearby gap check lines.

Before it was using thickness 0 and I could barely see it on my screen.  Now it's using the equivalent of thickness 0.5.  What do you recommend?

> This ones a minor issue -- it seems like each endpoint is forced to have a unique connection even if it isn't the shortest. You can see this at three point intersections. You'd expect two of the endpoints to share one of the connections and the odd one out to get it's own but there are almost always three connections. 

Yes, the currently logic somewhat forces unique connections.  It builds a list of possible closing segments and filters out some obvious ones (line-2-line, acute angles, etc).  It then applies the segments from smallest to largest, dropping a segment if either of the segment's endpoint is the same starting endpoint of a smaller segment that was previously applied. 

With this, a line's endpoint can only be be the start of a closing segment to another stroke 1x but at the same time it can be the ending of other closing segments from other strokes.  This isn't noticeable unless you can see the direction the strokes are drawn in. 

In your provided examples, I have run into 3 line endpoints that get closed in a triangle of closing segments.  It may be technically correct, but not necessarily ideal.  I've attempted to add logic to try and filter that, but usually ends up with only 1 closing segment and potentially 2 other gaps that cannot be closed.  For now I'll leave it as is as I'd rather have more than less. 

> The only other thing on my wishlist would be a lasso mode. It'd be more accurate that way. But, really, this is great without it.

Would need to handle that as a new feature request.

melieconiek:
> Before it was using thickness 0 and I could barely see it on my screen. Now it's using the equivalent of thickness 0.5. What do you recommend?

0.5 thickness is fine then.

As for the other issues, yeah, don't worry about them. This is fantastic as is. Just wanted to put the feedback out there. The only real issue was the long line thing.

Thank you for taking the time to sort this and dive into the fill issues.

artisteacher:
The thickness is perfect on my high res mac. There was a point back in the day where it gap check was nearly invisible, and a point where the lines were so thick it was hard o see what was going on.

So if I rectangular tape with style 0, in this case:
Screenshot 2025-01-30 at 8 17 06 PM
This one gap doesn't go away until I tape a 2nd time to  target it individually...
Screenshot 2025-01-30 at 8 24 04 PM

manongjohn:
@artisteacher I'd be curious to see what segments were closed using non-0 thinkness lines

artisteacher:
Do you mean like this?
Screenshot 2025-01-30 at 9 48 52 PM

Actually, it looks like it all closed, it's just that the one gap still displays.

artisteacher:
Where does the tape tool get its thickness from?

manongjohn:
> But you probably meant like this:

Yes. Thanks.  Can you send me that image?  I'd like to see why it doesn't recognize the segment that is already there.  I'm guessing it's because it's an end-2-line segment and due to fractional calculations it's seeing this gap as something different.

> Where does the tape tool get its thickness from?

The thickness of the strokes at either end of the closing segment determines the thickness at each endpoint and then naturally transitions the segments thickness between them..

artisteacher:
Here you go:
[A.pli.zip](https://github.com/user-attachments/files/18624126/A.pli.zip)

manongjohn:
Although there are still a few things I'd still like to address, I think what I've done helps quite a bit so I will move forward with this.

Thanks for the patience and feedback.  One step closer to a better experience with Vector levels...

Cryobus:
> I was able to use this for some cleanup work and ran into a few things.
> 
> There seems to be an issue with very long lines. Instead of closing to the line it will close to the closest endpoint of it. Problem intersections are circled in white. [__1-28-25_GAP-TEST.zip](https://github.com/user-attachments/files/18578707/__1-28-25_GAP-TEST.zip)
> 
> The check gap lines are a bit too thick. When you're drawing with 1px lines it can be hard to see what you're looking at. The gap check lines will often obscure linework and other, nearby gap check lines.
> 
> This ones a minor issue -- it seems like each endpoint is forced to have a unique connection even if it isn't the shortest. You can see this at three point intersections. You'd expect two of the endpoints to share one of the connections and the odd one out to get it's own but there are almost always three connections. I wanted to mention it but trying to change this might not work as well as I'm imagining -- and it's working great otherwise so maybe we just leave it.
> 
> The only other thing on my wishlist would be a lasso mode. It'd be more accurate that way. But, really, this is great without it.
> 
> **This is working a lot better! It closes everything you'd expected to with few unneeded connections. If I'm closing the whole drawing at once I'll just use 0 thickness lines. The extra connections don't matter -- what's important is that everything can be filled without issue. I don't have to convert to smart raster to to do fills. This has already saved me a ton of time. Any other issues I've encountered were the same old fill bugs. Nothing to do with the tape tool.**

Hello , I was wondering how can you choose the thickness of the lines the tape tool is going to trace? You said you would close the drawing with 0 thickness lines but it doesn't work for me, can you explain your method?

melieconiek:
> Hello , I was wondering how can you choose the thickness of the lines the tape tool is going to trace? You said you would close the drawing with 0 thickness lines but it doesn't work for me, can you explain your method?

Vector thickness can be set via color by using the Constant effect under the vector tab.

1. Create a new color index.
2. Select Constant under the vector tab.
3. Set thickness to 0 under the settings tab.
4. Select this style before applying the tape tool. The lines it creates will use the currently selected style and will be zero thickness.

Cryobus:
> > Hello , I was wondering how can you choose the thickness of the lines the tape tool is going to trace? You said you would close the drawing with 0 thickness lines but it doesn't work for me, can you explain your method?
> 
> Vector thickness can be set via color by using the Constant effect under the vector tab.
> 
>     1. Create a new color index.
> 
>     2. Select Constant under the vector tab.
> 
>     3. Set thickness to 0 under the settings tab.
> 
>     4. Select this style before applying the tape tool. The lines it creates will use the currently selected style and will be zero thickness.

I'm sorry I tried what you suggested but it doesn't seem to work... obviously I'm doing something wrong, I'm switching to the second style with 0 thickness but the lines keep closing applying the first style:
testConstant0thickLines

melieconiek:
> I'm sorry I tried what you suggested but it doesn't seem to work... 

You have `Join Vectors` enabled. Check the tool options bar.

Consider joining the discord server if you need help with something.

Cryobus:
Wow what an eye :) thank you so much!

PR #1694, opened 2025-01-17
Fix missing Theme selector (macos/linux)

manongjohn:
This fixes #1690, the missing `Theme` selector on macOS and Linux builds.

This was due to adding a preference setting check for High DPI Scaling (#1660) before everything was fully initialized at startup.

Changed the location of where everything was initialized which primarily impacts how macOS and Linux builds determines the current working directory.

PR #1705, opened 2025-01-28
Fix macOS working directory path

manongjohn:
Fix fixes the reported problems introduced by changes made in #1694 (a fix for #1660):
- Task rendering would start but never complete.
- The installed version would not start

This was due to how the working directory path was being determined but not used correctly in other areas of T2D after the prior fix.  The working directory path is now the application's root folder (Tahoma2D.app) instead of the directory it is run from.


artisteacher:
For the portable build, Task rendering now works perfectly. 

The installer itself fails if there are any portable versions of T2D, otherwise the installed version opens as expected.

PR #1576, opened 2024-09-15
Fix ffmpeg MP4/MOV color space

manongjohn:
This fixes an issue with MP4/MOV renders through ffmpeg having their colors look muted (or desaturated, or something other than actual).

Thanks to @RodneyBaker 's investigation and some testing, he found right ffmpeg parameters to correct the color of MP4 renders through ffmpeg.
```
  postIArgs << "-pix_fmt";
  postIArgs << "yuv420p";
  postIArgs << "-b:v";
  postIArgs << QString::number(finalBitrate) + "k";
  postIArgs << "-vf";
  postIArgs << "scale=" + QString::number(outLx) + "x" + QString::number(outLy) + ",";
  postIArgs << "colorspace=all=bt709:iall=bt601-6-625:fast=1"; // Adding color space conversion
  postIArgs << "-color_primaries";
  postIArgs << "bt709";
  postIArgs << "-color_trc";
  postIArgs << "bt709";
  postIArgs << "-colorspace";
  postIArgs << "bt709";
```

@konero also provided a slightly different set of ffmpeg parameters to use.
```
  postIArgs << "-pix_fmt";
  postIArgs << "yuv420p";
  postIArgs << "-vf";
  postIArgs << "scale=" + QString::number(outLx) + "x" + QString::number(outLy) + ":in_color_matrix=bt709:out_color_matrix=bt709";
  postIArgs << "-b:v";
  postIArgs << QString::number(finalBitrate) + "k";
```
I wound up using @RodneyBaker's version, but have added both of them as co-authors in the commit. Thank you both for this fix.

This fix, though appears to be good for Windows renders, doesn't quite fix macOS renders.  It does, however, seem shift the color closer to what T2D displays in macOS.  Further investigation for parameters better suited for macOS needs to be done.

PR #1713, opened 2025-02-07
Fix MOV transparency

manongjohn:
This fixes the loss of MOV transparency after implementation of colorspace adjustments (#1576) 

The ffmpeg parameters to adjust the colorspace provided by Rodney appears to remove the transparency from MOV renders. As of yet I have not found any settings to overcome this.

I've switched the logic for MOV renders to use Konero's ffmpeg parameters which keeps the transparency.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ff_mov.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ff_mov.cpp
@@ -94,13 +94,13 @@ TLevelWriterFFMov::~TLevelWriterFFMov() {
   postIArgs << "yuva444p10le";
   postIArgs << "-profile:v";
   postIArgs << "4";
-/* KONERO_VERSION
+// KONERO_VERSION
   postIArgs << "-vf";
   postIArgs << "scale=" + QString::number(outLx) + "x" +
                    QString::number(outLy) +
                    ":in_color_matrix=bt709:out_color_matrix=bt709";
-*/
-// RODNEY VERSION
+//
+/* RODNEY VERSION
   postIArgs << "-vf";
   postIArgs
       << "scale=" + QString::number(outLx) + "x" + QString::number(outLy) +
@@ -114,7 +114,7 @@ TLevelWriterFFMov::~TLevelWriterFFMov() {
   postIArgs << "bt709";
   postIArgs << "-colorspace";
   postIArgs << "bt709";
-//
+*/
   postIArgs << "-b:v";
   postIArgs << QString::number(finalBitrate) + "k";
 


artisteacher:
On macOS there is definitely an obvious color shift - which is very clear when bringing the mov back into T2D.

Screenshot 2025-02-11 at 8 09 03 PM

The top diagonal shows the original image - a raster clipped by a vector level.
The bottom diagonal shows the imported mov - using another clipping mask to show both.

Some colors are more dull, some are brighter.

manongjohn:
Yeah, I noticed that reimporting the MOV back into T2D brightened certain colors but that's not new for me on Windows at least.  I seem to recall it did that with MP4s too. I am uncertain as to why it does that.

I am guessing something is encoded in T2D generated MOVs that ffmpeg is causing some color shifting on import. Havent tried with MOVs created from elsewhere.

Outside of T2D, however, are the results the same, better, worse than before this change?

artisteacher:
In a recent nightly, there's no perceivable color shift when bringing the render (with a white background) back into T2D but there is a shift outside of T2D.
Screenshot 2025-02-11 at 8 30 37 PM
Top is Preview in T2D Nightly, bottom is macOS Preview - but the frame number is different.

Now the same comparison with this action build:
Screenshot 2025-02-11 at 8 36 05 PM

artisteacher:
The colors tend to be duller outside of T2D.
Nightly build first. Action build second.
Screenshot 2025-02-11 at 8 55 21 PM
I guess I'm not noticing an obvious change.

manongjohn:
> The colors tend to be duller outside of T2D.
> Nightly build first. Action build second.

So outside of T2D, no obvious changes, which is somewhat a good thing I suppose.  I think it was still slightly better than before the whole colorspace changes were made. 

Unfortunately, to restore transparency back into MOV renders, i need to keep this new ffmpeg configuration.  Is this change detrimental for macs?

I'll need to see if I can figure out why reimports are different, but I think that should be handled separately.

manongjohn:
> I think it's ok to move forward with.

Thanks!

> Any chance QT6 has colorspace improvements?

Interestingly enough, in Qt5.14 it appears a `QColorSpace` class was introduced and is also in Qt6.  Would need time to explore this new class and see how to integrate into the application.

artisteacher:
> Interestingly enough, in Qt5.14 it appears a QColorSpace class was introduced and is also in Qt6. Would need time to explore this new class and see how to integrate into the application.

That is exciting news!

konero:
@artisteacher What color depth did you use in T2D (8bit, 16bit, 32bit)? I notice FFmpeg's intermediate type is written with QImage_ARGB32 to PNG (?), wouldn't it need to be 64bit to handle ProRes? I wonder if color is lost somewhere or gamma not matching, I think supporting ProRes at the moment might be redundant. Also ProRes4444 (especially YUVA444P10LE) has low compatibility, best for people who know what they're doing with the specific format.

It might be better to do PNG in MOV: `ffmpeg -i seq.%04d.png -c:v png -compression_level 9 out.mov`

It's the container I use for my work at the moment. You shouldn't need to worry about colorspace, transfers, etc.

We could add a bunch of codec options (I can give commands) but the UI needs to update based on a enum, either toggling visibility of some properties or disabling them since some won't be compatible with certain codecs, and the UX would be strange if for example we had bitrate shown that can't apply to a certain codec, or alpha transparency toggle. I tried this myself, but I don't think the implementation is there for it? There's listeners in tproperty.cpp but setting it up nicely to toggle visiblity of other properties a bit complicated for me. 😂

artisteacher:
@konero - I just used whatever the default was. I think I've previously experimented with the different bit settings, but none of them prevented the color shift at the time.

> It might be better to do PNG in MOV: ffmpeg -i seq.%04d.png -c:v png -compression_level 9 out.mov
It's the container I use for my work at the moment. You shouldn't need to worry about colorspace, transfers, etc.

If that would stop the colorspace issue, that would be amazing.

It does seem a little limiting atm, not having any codec options but I do get how that can be complicated.

PR #1721, opened 2025-02-12
Fix canceling cell/column selection

manongjohn:
When drag bars are not enabled, to clear an active multi-cell/column selection, you currently need to tap an unselected cell/column. 

This fix allows the canceling a cell or column selection by effectively tapping any part of the active selection without moving the selection.  Similar behavior can be found in Krita/CSP.

"Effectively tapping" means either quick tap or a click, pause or movement within the selected cell/column, and release. 
Clicking, moving the selection away and back to the original position, and then releasing will not reset the selection.


artisteacher:
This is so much better!

PR #1735, opened 2025-02-13
Fix Excessive Frame Range Tape lag

manongjohn:
This fixes an issue with excessive lag experienced when using the Vector Tape Tool's Frame Range option across many frames.

When the Frame Range option was added to the Vector Tape tool (Rectangular Mode), some logic was incorrectly applied causing every cell to be visited as many times as the number of frames in the range and potentially increasing the number of gap closures.


melieconiek:
Used this today. Everything seems to be working now. Thank you for the fix!

PR #1737, opened 2025-02-16
Scripts to automate compilation on Linux

charliemartinez:


linux_build.sh
```diff
@@ -0,0 +1,387 @@
...
+# File:       linux_build.sh
+# Author:     Charlie Martínez® <cmartinez@quirinux.org>
+# License:    https://www.gnu.org/licenses/gpl-2.0.txt
+# Purpose:    Script para compilar Tahoma2D
+# Supported:  Multi-distro (originally for Quirinux 2.0)
+# ======================================================================
+
+#
+# Copyright (c) 2019-2025 Charlie Martínez. All Rights Reserved.  
+# License: https://www.gnu.org/licenses/gpl-2.0.txt
```
manongjohn, line 13: Can you publish this under BSD or at most, LGPL?  To allow T2D-OT sharing, i don't allow anything licensed with GPL.

manongjohn:
FYI, there are already build scripts found under `ci-scripts\linux`.  They all have the format `tahoma-*.sh`.  They are also used by Github actions for building and packaging.   Maybe integrate your dependency and build scripts into those instead of creating brand new ones?  Please be aware of impacting current Ubuntu builds.

charliemartinez:
> FYI, there are already build scripts found under `ci-scripts\linux`. They all have the format `tahoma-*.sh`. They are also used by Github actions for building and packaging. Maybe integrate your dependency and build scripts into those instead of creating brand new ones? Please be aware of impacting current Ubuntu builds.

Oh, I’m sorry, I wasn’t aware of those scripts. I created these for Quirinux, though they also work for Ubuntu, Debian, Redhat, Arch, Fedora, OpenSUSE, and their derivatives. For Ubuntu/Debian, the for loop ensures that if a dependency is not found, it continues with the next ones (including alternative dependency names for each distribution). I shared it as an alternative because I found it useful, but feel free to reject it if you already have a working solution. Cheers!

linux_build.sh
charliemartinez: > Can you publish this under BSD (preferred) or at most, LGPL? To allow T2D-OT sharing, i don't allow anything licensed with GPL.

Sure, I’ll modify it.

charliemartinez:
> FYI, there are already build scripts found under `ci-scripts\linux`. They all have the format `tahoma-*.sh`. They are also used by Github actions for building and packaging. Maybe integrate your dependency and build scripts into those instead of creating brand new ones? Please be aware of impacting current Ubuntu builds.

I just reviewed the scripts in ci-scripts, and I think the approach is somewhat different. My scripts, linux_depends.sh and linux_build.sh, fully automate the instructions published at https://github.com/tahoma2d/tahoma2d/blob/master/doc/how_to_build_linux.md, providing information to the user in multiple languages with some interaction and dialogues.

After reviewing tahoma-build.sh, I see that it is quite simple. I believe what I did is an extension where that part is handled by my _build() function in linux_build.sh

Regarding the part executed by tahoma-buildlibmypaint.sh, I had already noticed when trying to follow the documentation that it is not working. That’s why, in my linux_depends.sh, I only included dependencies from official packages.

manongjohn:
It's fine. No need to integrate them.  Was just a thought. The scripts will definitely help local builders regardless.  It's nice that they account for other distros.

The only scripts I would expected to have modified would have been `tahoma-install.sh` and `tahoma-build.sh`.  The others are specific to official builds because I can't use certain pre-built libraries when compiling due to licensing concerns.  That is why they are separated from the main build scripts so that non-official builds can use whatever packages are in their package managers.

charliemartinez:
Includded in https://github.com/tahoma2d/tahoma2d/pull/1739

PR #1742, opened 2025-02-17
Update Build Cache Management

manongjohn:
This updates the Github Action build scripts to manage the build caches more efficiently because they are currently creating so many for the same PR and causing the master caches to be deleted when we exceed our cache quota.

Now each PR or branch (i.e. master) should have at most 4 caches (Windows, macOS, Linux-clang, Linux-gcc) and it will use and update (delete/save) the caches with each subsequent push.

Once a PR is merged/closed, a new workflow has been added to delete all caches related to it.


manongjohn:
Due to some permissions issues with PRs, needed to change the approach.  PR details updated to reflect the change.

Merging to test the rest of this PR

PR #1750, opened 2025-02-20
Increase Size of Step Lines in Motion Path

charliemartinez:
The lines representing the steps in the motion path were quite small, making it difficult to see them clearly on small screens or from a certain distance. This often required excessive zooming in on the preview, causing a loss of reference to the overall composition of the photograph. The size has been increased to improve visibility and usability.

imagen


charliemartinez:
@manongjohn 

WARNING: 
**webcam** didn't work when testing this artifact. I don't know if it's a pre-existing error, although the same issue is happening **with the local builds I make from the master branch**.

[ WARN:0] global /home/runner/work/tahoma2d/tahoma2d/thirdparty/opencv/modules/videoio/src/cap_v4l.cpp (893) open VIDEOIO(V4L2:/dev/video0): can't open camera by index
error: frame not read from webcam

PR #1754, opened 2025-02-23
Fix Delete Style Color

manongjohn:
An issue with deleted styles was fixed in 1.3.  With that the change anything drawn with that style that was not deleted would to turn black.  Prior to that fix, it used to turn red.

At a suggestion, this PR changes the color of strokes drawn in a delete style to turn red instead of black.


beeheemooth:
I checked, it works. Thanks!

PR #1755, opened 2025-02-25
New linux local builds

charliemartinez:


charliemartinez:
It's quite advanced, but I don't want to cause any issues. I'll keep improving it in my repository, and if you're ever interested, you can take it from there.

https://github.com/charliemartinez/tahoma2d/tree/scripts-linux

PR #1756, opened 2025-02-25
Adjust parallelism and exit codes for build scripts

limitedAtonement:
It's difficult to detect failures if they're ignored and the build keeps going. Adding "|| exit" after a command will

1. Check the return code of the pipeline
1. If the return code is non-zero, "||" runs the second command, "exit"
1. "exit" will exit the script returning the exit status of the last command executed (see man bash; "exit [n]")

Also, for parallelism, it's best to query the platform for its parallel capacity. Rather than using all available cores, this commit uses one less so that the computer is still mostly responsive. Of course, if there is only one core available, we should avoid -j0, so we check for that situation first.

### DIFF
--- a/ci-scripts/linux/tahoma-build.sh
+++ b/ci-scripts/linux/tahoma-build.sh
@@ -1,17 +1,21 @@
 #!/bin/bash
-pushd thirdparty/tiff-4.2.0
-CFLAGS="-fPIC" CXXFLAGS="-fPIC" ./configure --disable-jbig --disable-webp && make
-popd
 
-cd toonz
+# Leave one processor available for other processing if possible
+parallel=$(($(nproc) < 2 ? 1 : $(nproc) - 1)) || exit;
+
+pushd thirdparty/tiff-4.2.0 || exit;
+CFLAGS="-fPIC" CXXFLAGS="-fPIC" ./configure --disable-jbig --disable-webp && make -j "$parallel" || exit;
+popd || exit;
+
+cd toonz || exit;
 
 if [ ! -d build ]
 then
    mkdir build
 fi
-cd build
+cd build || exit;
 
-source /opt/qt515/bin/qt515-env.sh
+source /opt/qt515/bin/qt515-env.sh || exit;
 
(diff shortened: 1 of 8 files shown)


manongjohn:
I believe all you really needed to do was `set -e` right after `#!/bin/bash` and get the same effect 

The github actions scripts actually run these scripts as `/bin/bash -e <script>` which is why it would have failed if any statement failed during the build.

Added `|| exit` after every statement makes it ugly to read and every time I need to modify these scripts, I need to remember to add that to every new line that's added.

manongjohn:
I know these scripts can be useful for building locally. I know i use them (sometimes modified) when testing changes and have to recompile a lot.

I'd rather not have all these `|| exit` appended to every single statement. That's just unnecessay.  If you want to redo this PR and put `set -e` starements instead, please do so. Otherwise, please close.

Either way, I do appreciate the effort and hope this doesnt discourage you from contributing in the future. 🙂

limitedAtonement:
Thank you for the well wishes! I have already developed a considerable battery of scripts, the most useful of which builds the App Image in a 20.04 docker container. It actually works on my arch linux machine! <3

limitedAtonement:
Oh yeah, regarding the `-j` changes, `nproc` on my machine returns 48, so it's much faster if I compile with a higher parallel count. I assume this might also improve build times in Microsoft Github Actions? And potentially on any machine that compiles this code.

manongjohn:
> Oh yeah, regarding the `-j` changes, `nproc` on my machine returns 48, so it's much faster if I compile with a higher parallel count. I assume this might also improve build times in Microsoft Github Actions? And potentially on any machine that compiles this code.

Oh. I forgot about the nproc changes.  Can't hurt to have that, right? 😄

Building Linux version, without a ccache, does cause it to take about 30+ minutes to build all of T2D. With ccache, much less than that.  Any improvement on building T2D faster is always welcome.

Now if i can do that for Windows and macOS builds too! Ccaches help, but still a bit slow in my opinion.

PR #1760, opened 2025-02-26
New deb-creator from Appimage script

charliemartinez:
Improvements made to previous attempts:

-Added function to calculate and print size for the installed app in .deb DEBIAN/Control
-Simplified deb-template to extract as much information as possible from the AppImage
-Added informative messages for debugging
-Improvements in DEBIAN/control and minor fixes.

### DIFF
--- a/toonz/installer/linux/deb-creator/deb-template/DEBIAN/postinst
+++ b/toonz/installer/linux/deb-creator/deb-template/DEBIAN/postinst
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -e  
+
+# Function to adjust permissions of directories and files
+adjust_permissions() {
+  local dir=$1
+  if [ -d "$dir" ]; then
+    echo "Adjusting permissions in: $dir"
+    find "$dir" -type d ! -perm 755 -exec chmod 755 {} +
+    find "$dir" -type f ! -perm 644 -exec chmod 644 {} +
+  else
+    echo "Directory $dir does not exist."
+  fi
+}
+
+# Function to add alias to each user's .bashrc
+add_alias_to_bashrc() {
+    local bashrc_file="$1/.bashrc"
+    local alias_command="alias tahoma2d='tahoma2d; tahoma-permissions'"
+
+    if [ ! -f "$bashrc_file" ]; then
+        echo "Creating $bashrc_file..."
+        touch "$bashrc_file"
+    fi
+
+    if ! grep -qF "$alias_command" "$bashrc_file"; then
+        echo "Adding alias to .bashrc in $bashrc_file..."
+        echo "$alias_command" >> "$bashrc_file"
+    fi
+}
+
+# Copy files from /etc/skel and adjust permissions
+SRC="/etc/skel/.local/share/Tahoma2D"
+if [ ! -d "$SRC" ]; then
+  echo "Source directory $SRC does not exist. Aborting."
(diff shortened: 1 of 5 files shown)


toonz/installer/linux/deb-creator/deb-template/DEBIAN/postinst
```diff
@@ -0,0 +1,81 @@
...
+        [ -f "$USER_TAHOMA_DIR/$file" ] && chmod +x "$USER_TAHOMA_DIR/$file"
+    done
+    find "$USER_TAHOMA_DIR/ffmpeg" -type f ! -perm /u+x -exec chmod +x {} +
+  else
+    echo "Skipping $user_dir: not a valid user directory."
+  fi
+done
+
+# Adjust common permissions
+adjust_permissions "$HOME/.local/share/Tahoma2D/tahomastuff/profiles"
```
limitedAtonement, line 68: What is the point of this? What happens if you skip this step?

toonz/installer/linux/deb-creator/deb-template/DEBIAN/postinst
```diff
@@ -0,0 +1,81 @@
...
+adjust_permissions "$HOME/.local/share/Tahoma2D/usr/share/doc"
+adjust_permissions "$HOME/.cache/Tahoma2D"
+adjust_permissions "$HOME/.local/share/Trash"
+
+# Set execution permissions for necessary binaries
+chmod +x /usr/local/bin/tahoma-permissions
+chmod +x /usr/local/bin/tahoma2d
+
+echo "Postinst successfully completed."
+exit 0
```
limitedAtonement, line 81: This "do not enter" sign is a good warning. Text files should end with a newline: https://stackoverflow.com/questions/729692

charliemartinez: I am human. Anyway, I would prefer not to make more commits that trigger the entire artifact generation mechanism (consuming bandwidth) unless it's absolutely necessary, or @manongjohn  will kill me...
What I can say for now is that on Debian, everything I've uploaded here has worked very well for me. Best regards.

toonz/installer/linux/deb-creator/deb-template/DEBIAN/postinst
charliemartinez: > What is the point of this? What happens if you skip this step?

From what I can see, some files won’t even exist at the time the .deb is installed. I now remember that we updated the .deb when we were notified about these permission issues, with the idea that it would help fix them on users' installed systems when updating from our repository (and from that point, we incorporated our tahoma-permissions script). So I think some of these lines could be removed—I’ll review which ones now.

In any case, they don’t negatively affect the installation.

charliemartinez: This postscript was designed to fix permissions in already installed versions on Quirinux users' systems. Later, we implemented the "tahoma-permissions" script.

I think these two lines could be removed, as they won’t exist at the time of installation:

```
adjust_permissions "$HOME/.cache/Tahoma2D"  
adjust_permissions "$HOME/.local/share/Trash"  
```

In any case, the installation can proceed without any issues, even without making any changes.

manongjohn:
Merging so I can start integrating it into the build packaging

PR #1761, opened 2025-02-26
Build concurrently to reduce overall build time

limitedAtonement:
For parallelism, it's best to query the platform for its parallel capacity. Rather than using all available cores, this commit uses one less so that the computer is still mostly responsive. Of course, if there is only one core available, we should avoid -j0, so we check for that situation first.

### DIFF
--- a/ci-scripts/linux/tahoma-build.sh
+++ b/ci-scripts/linux/tahoma-build.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
+# Leave one processor available for other processing if possible
+parallel=$(($(nproc) < 2 ? 1 : $(nproc) - 1))
 pushd thirdparty/tiff-4.2.0
-CFLAGS="-fPIC" CXXFLAGS="-fPIC" ./configure --disable-jbig --disable-webp && make
+CFLAGS="-fPIC" CXXFLAGS="-fPIC" ./configure --disable-jbig --disable-webp && make -j "$parallel"
 popd
 
 cd toonz
@@ -24,4 +26,4 @@ cmake ../sources  $CANON_FLAG \
     -DWITH_GPHOTO2:BOOL=ON \
     -DWITH_SYSTEM_SUPERLU=ON
 
-make -j7
+make -j "$parallel"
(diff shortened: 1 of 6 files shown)


manongjohn:
I'll be curious to see how quickly the linux artifacts from this PR builds compared to other recently opened PRs.

manongjohn:
Building w/ ccache wasnt a significant gain though comparing details to another PR, there were some gains in some areas, moreso on the `clang` compile side. Managed to save 2 minutes off the overall build time for `clang` but remained the same for `gcc`

It seems to have helped a little more noticably reducing ffmpeg compile time, but not as much with the other compiles.

That said, i wonder how it would have compared with no ccache to use.  It should be helpful but wont be able to tell unless i deleted our current caches.

Still a potential improvement for people using these scripts to build locally.

limitedAtonement:
I'm not sure what `nproc` returns on the Microsoft Github servers, but the speed increase is pretty minimal here. Oh yes, I forgot about the caching mechanism. Of course, a cached build won't benefit from populating the cache faster ^_^

I ran the build with and without this change and got a total of 74m39s without this change and 13m3s with. My concurrency level is 48 (so the build should be using 47 concurrency). This timing includes downloading Ubuntu, installing all the packages, and failing to `tar` at the end (I'm not sure why, but it doesn't seem like a material part of the timing).

image

manongjohn:
@limitedAtonement Impressive gain.  I don't think I've ever seen T2D build that fast without a cache :)

Sadly, the VM I use for fixing linux issues only has a nproc = 2 so I wouldn't get much from this myself.  I don't know that I've ever seen an nproc = 48 before.  Didn't think that was even possible. lol

Can I ask a favor to also update the `ci-scripts/osx/tahoma-*` scripts with the `nproc` changes?

manongjohn:
Will move forward and merge this. I can take care of the macOS scripts as at a later time.

Thanks for the improvement!

manongjohn:
Oh...sorey, wasnt expecting you to build an environment, just updating the scripts and letting Actions test it. 🙂

limitedAtonement:
> Oh...sorey, wasnt expecting you to build an environment, just updating the scripts and letting Actions test it. 🙂

Well that would make a lot of sense.

PR #1763, opened 2025-02-27
Disable mouse wheel in stopmotion control panel

charliemartinez:
I still need to test it with a DSLR camera; it worked fine with a webcam.

### DIFF
--- a/toonz/sources/stopmotion/stopmotioncontroller.cpp
+++ b/toonz/sources/stopmotion/stopmotioncontroller.cpp
@@ -67,6 +67,8 @@
 #include <QDomDocument>
 #include <QHostInfo>
 #include <QDesktopServices>
+#include <QEvent> 
+#include <QWheelEvent> // To disable the mouse wheel
 
 #ifdef _WIN32
 #include <dshow.h>
@@ -297,6 +299,8 @@ StopMotionSaveInFolderPopup::StopMotionSaveInFolderPopup(QWidget *parent)
 
   m_autoSubNameCB      = new QCheckBox(tr("Auto Format:"), this);
   m_subNameFormatCombo = new QComboBox(this);
+  m_subNameFormatCombo->setFocusPolicy(Qt::StrongFocus); // To disable the mouse wheel
+  m_subNameFormatCombo->installEventFilter(this); // To disable the mouse wheel
   m_subFolderNameField = new QLineEdit(this);
 
   QCheckBox *showPopupOnLaunchCB =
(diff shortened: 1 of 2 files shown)


charliemartinez:
@manongjohn Now, tested on Canon 600D. It works great too!!!

toonz/sources/stopmotion/stopmotioncontroller.cpp
```diff
@@ -1128,35 +1142,55 @@ StopMotionController::StopMotionController(QWidget *parent) : QWidget(parent) {
 
     m_apertureLabel  = new QLabel(tr(""), this);
     m_apertureSlider = new QSlider(Qt::Horizontal, this);
+    m_apertureSlider->setFocusPolicy(Qt::StrongFocus); // To disable the mouse wheel
+	m_apertureSlider->installEventFilter(this); // To disable the mouse wheel
```
limitedAtonement, line 1146: This line is indented with tab, but the rest of this file uses spaces.

toonz/sources/stopmotion/stopmotioncontroller.cpp
```diff
@@ -5113,8 +5163,8 @@ void ExportCalibrationFilePopup::showEvent(QShowEvent *e) {
 }
 
 //=============================================================================
-
-LoadCalibrationFilePopup::LoadCalibrationFilePopup(QWidget *parent)
+// To disable the mouse wheel
```
limitedAtonement, line 5166: I think this comment is incorrect. `LoadCalibrationFilePopup` is not "To disable the mouse wheel".

toonz/sources/stopmotion/stopmotioncontroller.h
```diff
@@ -164,6 +164,12 @@ protected slots:
 //-----------------------------------------------------------------------------
 
 class StopMotionController final : public QWidget {
+//-----------------------------------------------------------
+// To disable the mouse wheel
+	  protected:
```
limitedAtonement, line 169: Tabs and spaces on the same line here; just spaces is better.

toonz/sources/stopmotion/stopmotioncontroller.cpp
manongjohn: I typically code in Windows using MSVC which auto-formats code using a clang formatter. With my IDE, i can clang-format selected code (my preference) or the entire file.

When not using an IDE, there are `clang-format` binaries that you can run on a file to help standardize how code is formatted such as converting tabs to spaces and indentations. I do recommend using one.

However, 2 things to keep in mind with clang-format programs
- Different clang-format programs don't always format exactly the same way.
- As it is run against a file, it will correct formats for stuff you didnt change so when you submit your PR, there may be a ton of changes that are not yours. I typically dont mind these.

charliemartinez: Thanks

toonz/sources/stopmotion/stopmotioncontroller.cpp
charliemartinez: Great

charliemartinez: You are right.

charliemartinez: > I think this comment is incorrect. `LoadCalibrationFilePopup` is not "To disable the mouse wheel".

Thanks

PR #1768, opened 2025-03-02
Create Debian Package for Tahoma2D

manongjohn:
Modified the Debian creator script introduced by @charliemartinez and incorporated into Github Actions to generate `Tahoma2D-linux-gcc.deb` and `Tahoma2D-linux-clang.deb` Debian packages with every PR and nightly builds.

Added/modified some of the metadata information to be consistent with other public information.

Note:  Once this merges, I will need to update the website due to the additional packages being made available.


manongjohn:
@charliemartinez Please test the .deb artifacts from this PR and let me know if there are anu issues.

charliemartinez:
> @charliemartinez Please test the .deb artifacts from this PR and let me know if there are anu issues.

The installer works, although the camera does not. It does the same thing I describe [here](https://github.com/tahoma2d/tahoma2d/issues/1769) (only allowing the 600D to be selected).

charliemartinez:
OK, the problem was with my environment, the .deb is correct!! Sorry for the inconvenience.

xeiavicaflashrepository:
@manongjohn 
I strongly encourage you to collaborate with debian itself and get this added to sid because testing is freezing soon. Which will then become part of the ubuntu and linux mint repositories.

I suggest contacting them through the email provided in here:  
https://www.debian.org/devel/wnpp/

There already is a request to maintain OpenToonz for debian, but nobody has really taken an interest yet to do so, presumably because of some dependency issues.

manongjohn:
@xeiavicaflashrepository Thanks for the suggestion. To do that properly, I think I would need to actually build on a true Debian system and make sure it works properly using current libraries.  Atm, i dont have time to set up that environment and work out issues.

As it is, i know there are T2D issues with newer versions of ffmpeg and opencv libraries which is why i need to compile older compatible versions from source.

I even started out helping to set up a flatpak version of T2D, but havent gotten back to it. 🙁

xeiavicaflashrepository:
> @xeiavicaflashrepository Thanks for the suggestion. To do that properly, I think I would need to actually build on a true Debian system and make sure it works properly using current libraries. Atm, i dont have time to set up that environment and work out issues.
> 
> As it is, i know there are T2D issues with newer versions of ffmpeg and opencv libraries which is why i need to compile older compatible versions from source.
> 
> I even started out helping to set up a flatpak version of T2D, but havent gotten back to it. 🙁
I'm typing this on debian 12 bookworm on bare metal, I've compiled it just fine without issue. Do you need testing with sid?

xeiavicaflashrepository:
I'm typing on this on bare metal, debian 12 bookworm. I've compiled Tahoma2D without issue, what testing do you require?

manongjohn:
I believe one of the issues was related to newer versions of OpenCV. Something in the library impacted the Stop Motion stuff in T2D causing either crashes and/or wouldnt connect to webcams properly.

One of the reasons i also force use of specific library versions (OpenCV, libgphoto and libmypaint) and releasing AppImages is to prevent issues caused by dependency updates that suddenly make it incompatible with T2D. I had to do this for macOS builds also.

Going a little off topic here. Can discuss further on Discord or open a Discussion post if you want.

charliemartinez:
> I believe one of the issues was related to newer versions of OpenCV. Something in the library impacted the Stop Motion stuff in T2D causing either crashes and/or wouldnt connect to webcams properly.
> 
> One of the reasons i also force use of specific library versions (OpenCV, libgphoto and libmypaint) and releasing AppImages is to prevent issues caused by dependency updates that suddenly make it incompatible with T2D. I had to do this for macOS builds also.
> 
> Going a little off topic here. Can discuss further on Discord or open a Discussion post if you want.

Would it be possible for you to revert to older versions of libraries? There's little point in having updated libraries in Linux if they cause issues _(like with the new Canon SDK)._ I understand that for general use, Ubuntu is a popular distro, although [Quirinux](https://es.wikipedia.org/wiki/Quirinux) is a distribution designed exclusively for animated film. For example, in November, it was installed in all the classrooms of the [most important film school in Buenos Aires, Argentina](https://es.wikipedia.org/wiki/Instituto_de_arte_cinematogr%C3%A1fico_de_Avellaneda) (a total of 12 computers). Like most school and government distributions in Latin America, Quirinux is based on Devuan _(Debian without SystemD)_ simply because Ubuntu didn't run well on the older equipment typically used in these schools, not even Xubuntu. Best regards.

charliemartinez:
> I'm typing on this on bare metal, debian 12 bookworm. I've compiled Tahoma2D without issue, what testing do you require?

In general terms, a basic compilation of Tahoma2D works on Debian for animators who draw. With stop-motion, things get a bit more complicated. By compiling ffmpeg and opencv, I have managed to use a webcam, although the compilation of gphoto2 on Debian hasn't worked for using DSLR cameras like the Canon t3i, which does work with the .appimage or with this .deb.

PR #1775, opened 2025-03-09
OpenToonz Patches thru 3/8/2025

manongjohn:
The following fixes, refactors and minor enhancements have been ported from OT with some adjustments for T2D:

opentoonz/opentoonz/#5579 - Include functional in place of removed xstddef  by @shun-iwasawa
opentoonz/opentoonz/#5679 - Fix project reload fail  by @Abel032
opentoonz/opentoonz/#5717 - Fix Failure when updating the folder node  by @Abel032
opentoonz/opentoonz/#5718 - Optimize Adjust Current Level to This Palette  by @Abel032
opentoonz/opentoonz/#5759 - Avoid endless loop for some small fonts  by @Abel032
opentoonz/opentoonz/#5765 - Fix File Browser Model Refreshing  by @shun-iwasawa
opentoonz/opentoonz/#5760 - Fix selection in File Browser  by @Abel032
opentoonz/opentoonz/#5050 - Remove any_iterator  refactor / build  by @otakuto
opentoonz/opentoonz/#5787 - Fix Gif Export for newer FFmpeg versions  by @konero
opentoonz/opentoonz/#5799 - Cleanup: Replace deprecated endl with Qt::endl and fix operator prece…  by @andeon
opentoonz/opentoonz/#5746 - Fix Clean Up Palette Related Crash and cleanup parameters related adjustment  by @Abel032
opentoonz/opentoonz/#5788 - Update Webm Export  by @konero
opentoonz/opentoonz/#5793 - Fix Visualize Vector As Raster  by @Abel032[^1]
opentoonz/opentoonz/#5816 - [VectorSelectionTool] Fix Rasterized Vector Image update failure  by @Abel032

Aside: I've identified a number of OT features/enhancements that will be ported once OT releases them in their next version.

[^1]:  Included 2 cherry-picked commit's from opentoonz/opentoonz/#4852 - Krita-Like Assistant Tool Alternative by @blackwarthog  in order to apply this patch


melieconiek:
Tahoma is getting stuck on ' Initializing environment...' after importing my settings.

melieconiek:
Zero thickness lines are not visible when `Visualize Vector as Raster` is enabled.

andeon:
> Zero thickness lines are not visible when `Visualize Vector as Raster` is enabled.

Now I'm unsure if this is technically good or bad. I think it's correct as it is since it's simulating raster in its final form. `Show lines with zero thickness (0)` I think makes sense only in a vector context.

manongjohn:
> Tahoma is getting stuck on ' Initializing environment...' after importing my settings.
> 
> ~EDIT: It seems to be something in my layout. Copying in the layout folder causes the hang.~ Jumped the gun, it was something in my ENV file.

I'm seeing the same thing.  Happens when the Current Project is not the sandbox project.  I can open it of it's the sandbox, but if I switch to another one, existing or create a new one, and then restart, I hang again.

Will see which PR caused the problem and find out why.

melieconiek:
> Now I'm unsure if this is technically good or bad. I think it's correct as it is since it's simulating raster in its final form. `Show lines with zero thickness (0)` I think makes sense only in a vector context.

It's meant to visualize a vector level as a raster level. Even with these changes, it's not representative of a finished render. Moreover it's the only way to get around show-stopping brush lag under certain conditions. Zero thickness lines should be visible when it is active. You won't be able to edit them otherwise.

We may need an option to switch back to a lower  resolution as well. Unless there has been some significant improvements to the render speed this will not work well on lower end machines. Can someone with access to a slower machine test some complex scenes with this?

Abel032:
> Zero thickness lines should be visible when it is active. You won't be able to edit them otherwise.

I think we can't distinguish zero thickness lines or thicker lines if zero thickness vector lines are rasterized.
But I think the make the behavior of whether to rasterize zero thickness line to sync with the global zero thickness line visibility would be good.

melieconiek:
> But I think the make the behavior of whether to rasterize zero thickness line to sync with the global zero thickness line visibility would be good.

I agree with this, yes. In the current implementation, zero thickness lines are visible when the visibility option is disabled. This can be equally irritating.

As for a resolution option, a 'shrink' option similar to the render options might be nice.

Abel032:
> As for a resolution option, a 'shrink' option similar to the render options might be nice.

Now it's seems that the 'shrink' is controlled by current camera's dpi,
if we choose a smaller dpi, program would use bigger pixel when rasterizing a vector level.

melieconiek:
> Now it seems that the 'shrink' is controlled by current camera's dpi, if we choose a smaller dpi, program would use bigger pixel when rasterizing a vector level.

Changing this would be primarily for performance reasons. Think like changing the preview resolution for quicker feedback. You'd want the `Visualize Vector as Raster` scaling to be independent from the camera or preview scaling setting.

Abel032:
I mention dpi as it's like a bug, we face dpi problems, so it worthes a talk.

I believe if we don't rasterize zero thickness lines, keep them as vector and show them over the rasterized image would be good.(Because it's also functioning as the preview of exporting.)

And I agree that what done to vector image should also be shown on rasterized image.

I'm not about going to implementing these now, just suggest.

manongjohn:
I fixed the stuck Initializing problem when not starting with the sandbox project. 

Trying to catch up on the conversation so I'm not sure where it stands....does something need to change regarding 0 thickness vector lines showing in `Visual Vector as Raster` mode?

In the old logic, the 0 thickness lines were shown in Visual Vector as Raster regardless if you were showing them normally or not.

melieconiek:
> I fixed the stuck Initializing problem when not starting with the sandbox project.
> 
> Trying to catch up on the conversation so I'm not sure where it stands....does something need to change regarding 0 thickness vector lines showing in `Visual Vector as Raster` mode?
> 
> In the old logic, the 0 thickness lines were shown in Visual Vector as Raster regardless if you were showing them normally or not.

Thank you!

I need zero thickness lines to be visible in `visualize vector as raster` mode so that they can be edited while the mode is active. If this can respect the zero thickness lines visibility setting, all the better.

manongjohn:
> I need zero thickness lines to be visible in `visualize vector as raster` mode so that they can be edited while the mode is active. If this can respect the zero thickness lines visibility setting, all the better.

I agree that it should show, only if you enabled it to show in Preferences.  I am guessing it might have been an oversite in the original logic to rasterize 0 thinkness based on preference.

@Abel032: As you originally made the changes to Visual Vector as Raster, do you have plans on addressing 0 thickness lines in OT?

manongjohn:
I updated the `Visualize Vector as Raster` render params so now 0 thickness lines will show based on preference setting.

PR #1782, opened 2025-03-28
Fix on-canvas widget/indicator visibility

manongjohn:
This change attempts to improve the visibility of on-canvas widgets, guides and visual indicators for High DPI monitors.

With Full HD monitors on-canvas widgets, guides, and indicators were drawn knowing the device pixel ratio (DPR) was 1:1.  With high DPI monitors (i.e. 4K), the device pixel ratio is 1:2.  As such, a normal 1 pixel line appears thinner/fainter on a 4K monitor.

Added logic to get the DPR of the monitor where the canvas is at and adjust the thickness of lines accordingly.

### DIFF
--- a/toonz/sources/common/tvrender/tglregions.cpp
+++ b/toonz/sources/common/tvrender/tglregions.cpp
@@ -191,7 +191,7 @@ static void drawFirstControlPoint(const TVectorRenderData &rd,
 
   glPushMatrix();
   tglMultMatrix(rd.m_aff);
-  if (!rd.m_animatedGuidedDrawing) glLineWidth(2.0f);
+  if (!rd.m_animatedGuidedDrawing) glLineWidth(2.0f * rd.m_devPixRatio);
   glColor3d(0.0, 1.0, 0.0);
   if (!rd.m_animatedGuidedDrawing) {
     drawArrows(stroke, false);
@@ -235,7 +235,7 @@ static void drawFirstControlPoint(const TVectorRenderData &rd,
     glEnd();
   }
 
-  glLineWidth(1.0f);
+  glLineWidth(1.0f * rd.m_devPixRatio);
   glPopMatrix();
 }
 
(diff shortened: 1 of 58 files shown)


melieconiek:
Using `Perspective Grids` under `View` to enable prospective grids without first using the prospective grid tool will crash Tahoma.

melieconiek:
FYI I didn't notice any difference in widget appearance in the PR. I am using a 4k drawing display. Is there something I should be looking for?

manongjohn:
> FYI I didn't notice any difference in widget appearance in the PR. I am using a 4k drawing display. Is there something I should be looking for?

Some differences are more subtle than others. For me, I noticed a difference when I compared to v1.5.1.

May also depend on your resolution  and scale.  I use 3480x2400 at 225%

artisteacher:
This has been an issue for forever, but the Skeleton Tool's animate mode widget is still super tiny on High DPI. It should be bigger than the drawing origin...
Screenshot 2025-04-01 at 9 53 18 PM

Otherwise, this seems fine on macOS.

manongjohn:
> This has been an issue for forever, but the Skeleton Tool's animate mode widget is still super tiny on High DPI. It should be bigger than the drawing origin...

Yeah.  There are a few other on-canvas widgets that are really tiny.  The Pinch Tool's manual widget is so tiny that it's barely noticable

image

Debating if I want to handle that with this or as a separate ticket.  Would need an assessment of all the widgets that needs to be enlarged

manongjohn:
I fixed the Skeleton Tool's Animate and Pinch Tool's Manual widget sizes so it accounts for high DPI monitors.

If anyone sees any other widgets that are tiny in comparison to what it is on an Full HD monitor (if you have one....), let me know and I can look at increasing the size.

artisteacher:
Much better!

manongjohn:
I think I've reviewed all the on-canvas widgets and indicators for tools and column-based Fx and they all seem fine for now.

Still have some issues with color contrasts but will handle that separately at some point.

PR #1783, opened 2025-03-29
Show image in cell tooltip

manongjohn:
With this PR a cell's image will display in the tooltip when hovering over Raster, Smart Raster or Vector cell.

image

The image shown is the same image as shown in the Level Strip.

This feature can be enabled (default)/disabled via a preference setting in  the ` Scene` tab

image

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -465,6 +465,9 @@ public:
   QString getTimelineLayoutPreference() const {
     return getStringValue(timelineLayoutPreference);
   }
+  bool isShowImagesInCellTooltipEnabled() {
+    return getBoolValue(showImagesInCellTooltip);
+  }
 
   // Animation  tab
   int getKeyframeType() const { return getIntValue(keyframeType); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -154,6 +154,7 @@ enum PreferencesItemId {
   showFrameNumberWithLetters,
   showDragBars,
   timelineLayoutPreference,
+  showImagesInCellTooltip,
 
   //----------
   // Animation
--- a/toonz/sources/toonz/preferencespopup.cpp
+++ b/toonz/sources/toonz/preferencespopup.cpp
@@ -1452,6 +1452,7 @@ QString PreferencesPopup::getUIString(PreferencesItemId id) {
       {levelNameDisplayType, tr("Level Name Display:")},
       {showFrameNumberWithLetters,
        tr("Show \"ABC\" Appendix to the Frame Number in Scene Cell")},
+      {showImagesInCellTooltip, tr("Show Images in Cell Tooltips")},
 
       // Animation
       {keyframeType, tr("Default Interpolation:")},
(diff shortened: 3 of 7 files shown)


beeheemooth:
I tested it and here is my opinion:
1. Images take a very long time to load and also take a long time to disappear when you move the cursor to another layer. If you have a scene with a lot of levels, the delay is even longer.
2. If the cursor is stationary and you scroll the timeline or scroll the animation with shortcuts, then the images do not appear or do not update, although the layers under the cursor change.
3. The black rectangle with the frame number is too big, it additionally blocks the view. Is it possible to remove it and place the frame number in the corner of the drawing?
4. Is it possible to make the size of the drawing match the size of the icon in the Level Strip that the user specified in the Preferences?
5. The cursor overlaps the name (letter) of the level. I think it is more logical to move the level name to the right corner (as in the Level Strip)

manongjohn:
> 1. Images take a very long time to load and also take a long time to disappear when you move the cursor to another layer. If you have a scene with a lot of levels, the delay is even longer.

The time to appear and how long before it disappears is based on the default QT mouse/pen movement and standard Tool Tip Hint behavior.  There is a slightly longer delay (100ms) if it needs to generate the tooltip image for the 1st time.  This likely happens on newly loaded scenes.  However,  once it's generated, it will show under normal timings. I can lower the initial delay due to initial generation, but may result in temporarily showing a broken image if it isn't ready.

> 2. If the cursor is stationary and you scroll the timeline or scroll the animation with shortcuts, then the images do not appear or do not update, although the layers under the cursor change.

This was happening before this change and not a result of this PR. Tool tip behavior is based on mouse/pen pointer movement not the movement of what's underneath it.

> 3. The black rectangle with the frame number is too big, it additionally blocks the view. Is it possible to remove it and place the frame number in the corner of the drawing?

Not easily. Would have to worry about readability of the text against certain colors which involves a lot more work than I am willing to spend on this.

> 4. Is it possible to make the size of the drawing match the size of the icon in the Level Strip that the user specified in the Preferences?

I chose a larger size because I thought you could see details better.  Unfortunately that means it requires more memory to store a separate tooltip image. I will switch it to the Level Strip icon.

> 5. The cursor overlaps the name (letter) of the level. I think it is more logical to move the level name to the right corner (as in the Level Strip)

I will move it to below and to the right

manongjohn:
> Also, this feature can be very useful in some cases, and annoying in others. How do you feel about the function working from a shortcut (for example, Alt + Roll Over)?

Now that the image is smaller, it's not necessarily as obnoxious when it pops up. I don't particularly care for using modifier keys to enable this feature, but curious what others feel about it and/or this feature, in general.  

Does any other software do something similar? If so, how is it handled?

artisteacher:
You did a great job making sure the tooltip doesn't interfere with other interactions - selecting/moving keyframes or drawings which is great.

If a raster drawing or smart raster drawing is particularly small, it can still be really hard to see - just like the layer thumbnail. I kind of wish they both showed the actual drawing instead of the full canvas. I know you lose full context of placement in the camera the way, but vector works that way and it is a lot more helpful imo.

> Now that the image is smaller, it's not necessarily as obnoxious when it pops up. I don't particularly care for using modifier keys to enable this feature, but curious what others feel about it and/or this feature, in general.

I think the timing is working well, but I could see someone feeling like it is a distraction sometimes. I wonder if instead of a modifier, it might work to just have a command that toggles the feature on/off. I'd place it in the quick toolbar with the auto creation/hold commands, though users could also set a shortcut.

artisteacher:
Personally, I think the preference is fine. It's probably out of scope, but the preference tabs are getting pretty full. I wonder if some reorganization would make it easier to find things - or search capability.

manongjohn:
Yeah. Preferences are getting to be a bit cluttered, some tabs more than others.

manongjohn:
Merging. As usual, changes can be made at any time. Feel free to discuss in Discord.

PR #1784, opened 2025-03-30
Fix Mesh Animate crash of SmartVector Level

Abel032:
A Fix ported from opentoonz/[#5886](https://github.com/opentoonz/opentoonz/pull/5886)


manongjohn:
LGTM.  Thanks for bringing this fix over.

PR #1793, opened 2025-04-09
Add Raster Tool Option - Paint Behind

manongjohn:
A new `Paint Behind` tool option has been added to the raster `Brush Tool`.  Works with both standard and MyPaint brushes styles.

image

A shortcut can be assigned to it:

image

Additional change:

The `Paint Brush Tool`'s `Selective` option has been renamed to `Paint Behind`. It now uses the new shortcut entry as well.


beeheemooth:
In raster levels everything looks great! Thanks for your work!!!

manongjohn:
> Switch to Lines (Paint Brush Tool) - crash.

Should be fixed now

PR #1797, opened 2025-04-13
Smart Raster Selection - Selective mode

manongjohn:
This adds 2 new Smart Raster Selection Tool options, `Mode` and `Selective`:

image

`Mode` allows you to target pixels that were colored by the Brush tool (Lines), by the Paint Brush/Fill Tool (Areas) or both (standard selection behavior)
`Selective` allows you to only select pixels with the currently selected style.

Limitation:  When selecting `Areas`, the style will be pulled from blended pixels (has Ink and Paint) like in aliased lines and even unaliased lines where the fill tool added paint beneath it.  When this happens, the extracted style will become fully opaque as close to the center of the aliased line (or edge of solid line) creating an aliased edge of the area and will not blended back.

https://github.com/user-attachments/assets/e46ee9fa-94e8-4ea3-b8b8-4a81baafb46f

### DIFF
--- a/toonz/sources/include/tools/rasterselection.h
+++ b/toonz/sources/include/tools/rasterselection.h
@@ -38,6 +38,10 @@ class RasterImageData;
 //! Selection of TToonzImage and TRasterImage.
 //-----------------------------------------------------------------------------
 
+#define LINES L"Lines"
+#define AREAS L"Areas"
+#define ALL L"Lines & Areas"
+
 class DVAPI RasterSelection final : public TSelection {
   TImageP m_currentImage;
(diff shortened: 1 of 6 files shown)


beeheemooth:
Looks great! Is it possible to keep the selections when switching between styles?

manongjohn:
That's outside the scope of this PR. Will need to handle that separately.

manongjohn:
Per suggestion by @beeheemooth, I've allowed the selection `Mode` to work independently from the `Selective` option.  This way you can target all `Lines` or all `Areas` separately.   Targeting `Lines & Areas` is the default existing behavior.

PR #1800, opened 2025-04-15
Correct typo Frane -> Frame

limitedAtonement:


manongjohn:
Thanks for fixing the typo!

For future reference, no need to update the .ts files as I have to rebuild them and the .qm files.  I'll need to remember to generate them for the 1.5.3 fix release so this change takes effect.

PR #1805, opened 2025-04-17
New and updated shortcuts

manongjohn:
### New
- Plastic Tool
  - Tool Modifiers (Toolbar)
    - `Keep Distance`
  - Right-click Menu Commands
    - `Copy Skeleton`
    - `Paste Skeleton`
    - `Set Rest Key`  (Animate context menu)
    - `Set Global Key`  (Animate context menu)
    - `Set Global Rest Key`  (Animate context menu)

- Timeline/Xsheet
  - Onion Skin - Right-click Menu Commands
    - `Clear All Onion Skin Markers`
 
### Updated
- Plastic Tool
  - Tool Modifiers (Toolbar)
    - `Global Key` -  Now uses existing shortcut
  - Right-click Menu Commands
    - `Set Key`  (Animate context menu) - Now uses existing shortcut

- `New Output` - Removed default setting (`Alt+O`)

### DIFF
--- a/toonz/sources/tnztools/plastictool.cpp
+++ b/toonz/sources/tnztools/plastictool.cpp
@@ -12,6 +12,7 @@
 #include "toonzqt/dvmimedata.h"
 #include "toonzqt/dvdialog.h"
 #include "toonzqt/selectioncommandids.h"
+#include "../toonz/menubarcommandids.h"
 
 // TnzLib includes
 #include "toonz/tframehandle.h"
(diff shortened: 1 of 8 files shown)


beeheemooth:
If you touched Schematic, there are many options there that cannot be assigned shortcuts. The bottom panel (Focus on Current, Maximize Nodes, etc.), as well as options from the context menu of nodes. It is not necessary to assign shortcuts to them by default, it is important that the user has the ability to assign them.

manongjohn:
I think I'll handle schematic shortcuts separately at some other time.  There's potentially a lot there that would need shortcuts.

PR #1810, opened 2025-04-20
WinKey as Modifier Key

manongjohn:
The Windows Key which normally opens the Windows menu, can also be used as a modifier key (`Meta`).

This change does the following:
- Disables the `WinKey` from opening the Windows menu only when T2D is in focus
- Allows the `WinKey` to be used as a modifier key in shortcuts.  It shows as `Meta+`
- Allows for future features to use `WinKey` as a modifier much like `Ctrl`, `Shift`, `Alt`

This change only affects Windows builds.
The `Meta` key for macOS users is the `Command` key and this change should have no impact to those builds.
Exploring an option for Linux.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -95,6 +95,56 @@
 #include <QtPlatformHeaders/QWindowsWindowFunctions>
 #endif
 
+
+// Disable WinKey so we can use it as a Modifier Key
+// Not needed for macOS since Meta key = Command key
+#ifdef _WIN32
+#include <Windows.h>
+
+HHOOK g_hook = nullptr;
+HWND hMainWindow = nullptr;
+
+bool IsOurWindowFocused() {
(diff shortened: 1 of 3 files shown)


DarrenTAnims:
Is this an option to enable or disable?
I use the Windows key all the time, so I'd hate to lose it.

manongjohn:
I will add a preference for it.

I believe you will still be able to use T2D shortcuts with `Meta+` and it wont open Windows menu as it does now for winkey shortcuts. However, that might be a problem for mouse/stylus operations that may use Meta key in the future.

manongjohn:
Canceling this, for now. Seems it doesnt work for shortcuts.

manongjohn:
Just an FYI, discovered this is not needed.  Apparently you can use `WinKey` as a modifier key in T2D shortcuts, however you just cannot use combinations for existing Windows established shortcuts (i.e. `Winkey+R`, `WinKey+E`, etc.)

PR #1812, opened 2025-04-21
Contiguous cell selection shortcut

manongjohn:
In dragbar-less mode, this adds a fixed shortcut to allow selection of connected cells in the timeline/xsheet to simulate the original dragbar selection behanvior.

When you hold the shortcut, `Shift+Ctrl`, and click on a **non-selected** cell (implicit or explicit), it will select all cells (implicit or explicit) in that column that are connected...
...before it, stopping at the cell before an empty cell, a stop hold, or the beginning of the scene.
...after it. , stopping at and including a stop hold or the last exposed cell in the column. 

If you click an implicit cell beyond the last exposed cell in the column, the implicit cells between the last exposed cell and the cell you clicked will be selected as well.

If `Alt` is included to also select keys and the last exposed frame in the column is not a stop hold and there are keys beyond it, the selection will extend to the last key.

### DIFF
--- a/toonz/sources/toonz/xsheetdragtool.cpp
+++ b/toonz/sources/toonz/xsheetdragtool.cpp
@@ -142,7 +142,9 @@ public:
 
     int r0, c0, r1, c1;
     bool shiftPressed = false;
-    if (m_modifier & Qt::ShiftModifier) {
+    if ((m_modifier & Qt::ShiftModifier) &&
+        (Preferences::instance()->isShowDragBarsEnabled() ||
+         !(m_modifier & Qt::ControlModifier))) {
       shiftPressed = true;
       getViewer()->getCellSelection()->getSelectedCells(r0, c0, r1, c1);
     }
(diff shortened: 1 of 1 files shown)


beeheemooth:
I tested it and at first glance everything works fine! It's great that there is finally an option to select all cells! The logic using stop frames is well!
One nuance that I noticed: when duplicating, when you do a second Ctrl+Shift+Click on the selected frames, duplication occurs, but the selection is reset. Is it possible to make a condition that if Ctrl+Shift is still pressed, then repeated selection (or resetting of selection) would not occur, so that it would be possible to duplicate cells multiple times?

beeheemooth:
I don't think this is a good option. Sure, it would be nice to have the ability to use the "Win" button, but not for this purpose. Working with frames is almost intuitive and addictive, and can be problematic when using multiple programs. It's better to leave the current key combination you've set.

artisteacher:
Love this!

I did notice, if you put 2 different levels into the same layer, they both get selected. Not that I actually work this way - I like my levels on their own layers.
Screenshot 2025-04-22 at 4 06 14 PM


Also, if a drawing is active & then you press this shortcut, there's an explicit hold & the other drawings shift down 1.
Screenshot 2025-04-22 at 4 03 50 PM

manongjohn:
I guess the question for multiple levels in 1 column should the selection only be limited to the level selected or because they are connected, select it as well?  

As for the duplication...`Shift+Ctrl+click an active selection` triggers Copy/Insert of selection which is what you are seeing.  That is normal behavior for that sequence of actions.  That's why I specify you must click an unselected cell for this to work right.  Only other way to prevent this is to change the activation keys.

The creation of the explicit frame is something that needs to be fixed as a separate issue, however.

artisteacher:
So as far as the multiple levels go, with explicit holds, I think the dragbar only moves the active level, not multiple levels even if they are consecutive.

As far as the second goes, I really just intended to point out the explicit hold. Otherwise it is behaving as expected.

manongjohn:
Even though the original dragbar behavior only worked with the same level, we dont necessarily have to keep to that, or should we?

artisteacher:
I guess it's a question of which behavior someone might want to do more often. I certainly can see the benefit as is. If you just want to shift all the frames in the layer this is perfect - that's honestly what I envision doing the most. If you wanted to move one of the levels to a different layer, or to change the order of the levels in the layer, the explicit hold version might be better.

manongjohn:
I plan on merging this sometime later today.

I have no issues with this contiguous selection selecting multiple connected levels in the same column and will keep it like this unless someone has a reason why it shouldn't behave this way.

PR #1816, opened 2025-04-24
Switch from Git LFS to regular files

manongjohn:
Seems we've hit the LFS Bandwidth limit this month, I think.  I didn't think we were that active with so many action builds.

After some digging, it appears GitHub  changed their billing platform within the last few months and rolling it out in waves. Stuff like LFS bandwidth is now associated with $. Previously, LFS bandwidth usage was based on bytes downloaded but it is now based on $ calculated by the bytes downloaded.  It does not seem the byte-$ ratio is equatable.  Somehow in 15 days we used up all of the 10GBs allowed!

After reviewing the 98 files currently stored in LFS, none of them are larger than 100 MiB, so we are not required to use LFS.  The total current size of files in LFS is 180 MB.

This PR will convert the repo from using LFS storage to regular files.  I've also removed some libraries for building on MSVC 2013/2015 to help reduce the mount of space used by these files (117 MB).

This will avoid running into the monthly LFS bandwidth limitation that are used up by action builds in T2D's repo and any action builds run in forks of this repo.

Bonus: Saves me from having to pay for increased LFS bandwidth each month.


manongjohn:
Side note: I think I see why it appears we've hit the bandwidth limit. 

I was paying monthly for 1 LFS bandwidth Data Pack, which set the limit at 50GB.  Now that they've converted T2D to this new billing system, the default LFS bandwidth limit is 10 GB and no longer accounts for the Data Pack I purchased.

This explains why we appeared to have run out since I've managed to use up 10 GB in 9 days with the activity I was doing.

The new method is to set a $ limit for automatic billing of overages beyond the 10GB limit, which at worst, saves me $1-2/month, or $0 if there is no activity.  To LFS or not to LFS....

andeon:
Hi @manongjohn, hope you're doing well!

OpenToonz has also been experiencing this issue since last week: https://github.com/opentoonz/opentoonz/issues/5914.

I had been theorizing a lot about it, since I didn’t have access to some of the information that only the repository owner could see. Some of the details you shared actually confirmed a few of my suspicions.

Did you see anything suggesting that even forks might affect the quota? I found that really concerning, because it means forks could help drain those limits at the expense of the main repository, which makes LFS not very suitable for public open-source projects like OpenToonz or Tahoma2D in general.

manongjohn:
Yes forks absolutely use up the root repo's lfs bandwidth whenever the github actions in the fork is triggered because for at least Windows and macOS (T2D only) builds, it has to clone the repo and pull lfs files each time.  Linux builds dont use any thing from LFS so they dont use up bandwidth and are not affected by the limitation.

So unless the fork turns off the workflows in their fork, every push of branch changes for PRs causes double usage of bandwidth because they kick off actions in both the fork and the PR.   This is especially bad when people keep commit-pushing every little change they make. PRs are not meant to be used as interim development builds.  PRs should only be made when the dev feels it is ready to go live and only push changes to fix issues that are found.

With OT, it's worse since they have other actions enabled also like Appveyor. Appveyor does a debug build AND and release build, all adding up to bandwidth usage. They also have Jenkins builds. Not sure why OT bothers with all those.

Dwango/Digital Video was likely paying something for their bandwidth usage. With recent changes, it is probably ignored, like mine was, in favor of new method.  Right now they are likely stuck as 10GB bandwidth.

manongjohn:
Git LFS is required for files 100 MiB or larger, which thankfully none of our files are. LfS is useful if binary files are stored in the repo and are updated frequently.  The LFS files in T2D's repo are just DLL and LIB files that windows builds need.  They dont change often enough, in my opinion, so taking them out of LFS seems like a good idea and will no longer need to worry about LFS bandwidth.

The only other thought is some thirdparty dlls/libs might be best built each time as part of the dependency setup, since they might depend on MSVC version. Debating if I should do this now or wait. MSVC 2019 is being retired soon and will need to move to 2022.

xavier2k6:
xref: https://github.com/opentoonz/opentoonz/issues/5914#issuecomment-2828940111

manongjohn:
I've created a script to build some of the 3rd party LIB files from the source already in the repo but I've decided against it in order to not overly complicate the initial building of T2D.

I just force pushed to consolidate my commits so LFS converted files that i am getting rid of won't be in the commit history. 

Once complete, I will merge these changes so we no longer need to worry about LFS bandwidth running out and I can start pushing/merging  again.

manongjohn:
Merging.  Please update your forks and I recommend rebasing any branches you are actively working before pushing.

andeon:
Sorry for bothering you, John. But, I've been thinking about this approach, since the repository could grow significantly, although updating the libraries isn't something we do often. I noticed that these libraries are mostly included to support the Windows build process.

Let me know if I'm wrong about that. So, I'm considering removing all of them from the repository. Before doing that, I would create maybe a 'Libs' release containing all the libraries, zipped with the correct folder structure preserved. Then, I would update the Windows workflow to download and extract that 'Libs' release archive during the build process, maintaining the structure as if they were never removed.

This way, the main repository wouldn't keep growing, as the libraries would only be used in CI builds and downloaded from the release 'Libs'.

These libs are not meant for local compilation, they're intended for CI builds. That's how I understand, since there's a prj folder with .sln and .vcxproj files. Users compiling locally would still need to set up MSVC themselves, although not all libraries seem to follow this structure with the prj folder.

manongjohn:
All of them are used for local compiles, not just CI builds, otherwise you would have to build the libraries first before you can build T2D.  I actually built a script to build the libraries that can be build with MSVC using the source in the repo.

I also opted to potentially put these in a separate repo that could be downloaded as needed to keep them out of the main repo.

In the end, I chose to not over complicate the set up and compilation of T2D builds, locally or CI, so developers don't need to go and build every library dependencies or remember to pull submodules.   Setting up and building is already complicated enough.

I don't see these libraries being recompiled so often that the sudden growth of the repo is a huge concern so I'm ok with it in the repo.  The decision to leave them in the repo may come back to bite me in the future.  Will need to revisit this issue if it does.

andeon:
I understand. I was also looking into the possibility of using [Conan](https://conan.io/center/recipes?value=zlib+glew+libtiff+glut+lzo+libpng+libjpeg-turbo+superlu) in separate repo.

At first I thought it might be slow like vcpkg, but I was wrong about that. It also allows you to create patches, which is helpful since some libraries such as libtiff need adjustments. Another option is to copy the recipe into your own repository and customize it. This way, you can use export_sources to include files like tif_config.vc.h and tiffconf.vc.h, allowing you to change them before building while still taking advantage of Conan's. I just need to figure out how all these builds behave regarding the MSVC runtime configurations, but it might take me some time to figure all of this out. Working with libraries is always a bit of a headache. 

Thanks, I appreciate the discussion!

PR #1821, opened 2025-05-02
Loop Frames

manongjohn:
This new feature allows you to select 2 or more frames to implicitly and repeatedly loop the sequence on the timeline/xsheet.

The new actions, `Loop Frames` and `Remove Frame Loop`, can be found in the main `Cell` dropdown menu as well as the cell context menu. They can be assigned shortcuts.

image

Behavior
- Looping frames works whether you are using Implicit Holds or not.
- Looped frames will be implicitly looped until either a stop frame or an exposed frame is encountered.
- A looped selection is based on the scene frames they are initially set at and do not move with frame selections or frame movements.
- You can have multiple looped selections in the same column/layer but they cannot be overlapped or nested.
- Looped frames cannot be set on Sound, Note Text or Folder columns.
- Your selection can include multiple columns as targets for these actions.
- Most Cell commands will treat looped frames as empty frames.  Some Cell/Level commands may treat looped frames as real frames.

Display
- A looped selection is denoted by loop frame markers (brackets) on the start and end frames on the timeline/xsheet.
- Implicitly looped frames are uncolored (vs tinted for implicit holds) but will show frame numbers and stop frames that are looped.

Changing loop frames
- You can click and drag either end of the loop frame markers to expand/shrink the looped sequence.
- `Shift` + drag of either loop frame marker will shift both frame markers at the same time in the same direction
- `Alt` + drag of either loop frame marker will  shift both frame markers at the same time in opposite directions

Removing loops
- Select any frame within a looped selection and choose `Remove Frame Loop` to delete that loop.
- Can delete multiple loop selections, even across multiple columns, by ensuring your selection includes at least 1 frame in that loop.

Column looping
- In addition to the locations mention above, the `Loop Frames` and `Remove Frame Loop` actions can be found in the column context menu.
- You can select 1 or more columns as targets for these actions.
- Looping columns in this manner will define the sequence at the 1st expose frame and end at the last exposed frame.

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -147,7 +147,9 @@ enum class PredefinedRect {
   CLIPPING_MASK_AREA,
   FOLDER_INDICATOR_AREA,
   FOLDER_TOGGLE_ICON,
-  BUTTONS_AREA
+  BUTTONS_AREA,
+  LOOP_START_MARKER_AREA,
+  LOOP_END_MARKER_AREA
 };
 enum class PredefinedLine {
   LOCKED,              //! dotted vertical line when cell is locked
@@ -184,7 +186,9 @@ enum class PredefinedPath {
   FRAME_MARKER_DIAMOND_SMALL,
   FRAME_MARKER_DIAMOND,
   FRAME_MARKER_DIAMOND_LARGE,
-  NAVIGATION_TAG
+  NAVIGATION_TAG,
+  LOOP_START_MARKER,
+  LOOP_END_MARKER
 };
 enum class PredefinedPoint {
   KEY_HIDDEN,  //! move extender handle that much if key icons are disabled
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -73,6 +73,8 @@ class DVAPI TXshColumn : public TColumnHeader, public TPersist {
   QStack<int> m_folderId;
   int m_folderSelector;
 
+  QList<std::pair<int, int>> m_loops;
+
 public:
 private:
   int m_colorFilterId;
@@ -112,8 +114,9 @@ Constructs a TXshColumn with default value.
       , m_colorTag(0)
       , m_opacity(255)
       , m_colorFilterId(0)  // None
-      , m_folderSelector(-1)
-  {}
+      , m_folderSelector(-1) {
+    m_loops.clear();
+  }
 
   enum ColumnType {
     eLevelType = 0,
@@ -301,6 +304,28 @@ Set column color tag to \b colorTag.
 
   bool loadFolderInfo(std::string tagName, TIStream &is);
   void saveFolderInfo(TOStream &os);
+
+  bool hasLoops() const { return !m_loops.isEmpty(); }
+  QList<std::pair<int, int>> getLoops() const { return m_loops; }
+  void setLoops(QList<std::pair<int, int>> loops) { m_loops = loops; }
+  void addLoop(std::pair<int, int> loop) { m_loops.push_back(loop); }
+  void addLoop(int start, int end) { addLoop(std::pair<int, int>(start, end)); }
+  void removeLoop(std::pair<int, int> loop) {
+    m_loops.erase(m_loops.begin() + m_loops.indexOf(loop));
+  }
+  void removeLoop(int start, int end) {
+    removeLoop(std::pair<int, int>(start, end));
+  }
+  std::pair<int, int> getLoopForRow(int row);
+  std::pair<int, int> getLoopWithRow(int row); 
+  bool isInLoopRange(int row);
+
+  bool isLoopedFrame(int row);
+  int getLoopedFrame(int row, bool forOnionSkin = false);
+  TXshCell getLoopedCell(int row, bool forOnionSkin = false,
+                         bool implicitLookup = true);
+  bool loadLoopInfo(std::string tagName, TIStream &is);
+  void saveLoopInfo(TOStream &os);
 };
 
 #ifdef _WIN32
@@ -387,7 +412,8 @@ Return true if cell in \b row is empty.
 Return cell in \b row.
 \sa getCells and setCell()
 */
-  virtual const TXshCell &getCell(int row, bool implicitLookup = true) const;
+  virtual const TXshCell &getCell(int row, bool implicitLookup = true,
+                                  bool loopedLookup = true) const;
   /*!
 Set cell in \b row to \b TXshCell \b cell.
 \sa setCells() and getCell(); return false if cannot set cells.
@@ -399,7 +425,7 @@ Set \b cells[] from \b row to \b row + \b rowCount to column cells.
 \sa getCell and setCells()
 */
   virtual void getCells(int row, int rowCount, TXshCell cells[],
-                        bool implicitLookup = false);
+                        bool implicitLookup = false, bool loopedLookup = false);
   /*!
 Set column cells from \b row to \b row + \b rowCount to cells[].
 \sa setCell() and getCell(); return false if cannot set cells[].
--- a/toonz/sources/include/toonz/txsheet.h
+++ b/toonz/sources/include/toonz/txsheet.h
@@ -196,14 +196,19 @@ public:
      an empty cell.
           \sa setCell(), getCells(), setCells()
   */
-  const TXshCell &getCell(int row, int col, bool implicitLookup = true) const;
+  const TXshCell &getCell(int row, int col, bool implicitLookup = true,
+                          bool loopedLookup = true) const;
 
-  const TXshCell &getCell(const CellPosition &pos,
-                          bool implicitLookup = true) const;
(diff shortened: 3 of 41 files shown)


beeheemooth:
I don't understand, what's wrong with playback markers, which essentially do the same thing?

manongjohn:
@beeheemooth This is not a change or replacement for playback looping.  This replaces #1669 and allows you to select a sequence of frames to repeat on the timeline automatically and implicitly.

So if you want to repeat Frames 1,2,4 forever, you just select the frames, click `Loop Frames` and it repeats the sequence after it without you having to copy/paste the sequence over and over.

Should it be called something else to make it more clear?

beeheemooth:
Wow! Very cool feature! 
This is a very convincing argument to use Implicit Hold! So far I have not found any glitches. 
The only question is: do you think "phantom" frames should be merged with another level when using the "Merge Levels" option? (This is not a suggestion, just a question)

melieconiek:
This works very well.

- **The loop frames aren't exposed on the onion skin when relative drawing mode is enabled.**
- **Attempting to draw on a loop drawing will just draw on the last _real_ exposed drawing. You ought to be able to draw on the looped frames.**

Should `Next/Previous Drawing` allow you to navigate loop drawings and should be able to edit (draw on) loop drawings? This would making animating loops easier since you'd want to be able to flip between the last/first drawing in a loop and expose the onion skin around these frames. I'm not suggesting you should be able to move the loop drawing (the faded color works well to signal this, btw.)

manongjohn:
@beeheemooth 

> The only question is: do you think "phantom" frames should be merged with another level when using the "Merge Levels" option?

I could go either way with this.  Technically the frames are not really there so you would think it shouldn't, however the application knows what should be there and should be able to use what it knows.

@melieconiek 

> The loop frames aren't exposed on the onion skin when relative drawing mode is enabled.

Will look into.

> Attempting to draw on a loop drawing will just draw on the last real exposed drawing. You ought to be able to draw on the looped frames.

When I draw on a looped frame, it creates a brand new drawing for me like it usually does.  Just curious, do you have `Toggle Auto-Creation` and/or `Toggle Creation In Hold Cells` disabled?  If either of those is disabled it appears to use the last exposed frame since it thinks it's a held frame.

> Should Next/Previous Drawing allow you to navigate loop drawings and should be able to edit (draw on) loop drawings? 

In general and in some odd ways I have mixed feelings about treating these "phantom" frames as real frames, but I can see the benefit of being able to do some stuff with them as if they were real.  Kind of like my answer to beeheemooth...it's not really there but the application implicitly knows it is so it should in theory be able to.  Will look into.

melieconiek:
> When I draw on a looped frame, it creates a brand new drawing for me like it usually does. Just curious, do you have Toggle Auto-Creation and/or Toggle Creation In Hold Cells disabled? If either of those is disabled it appears to use the last exposed frame since it thinks it's a held frame.

Creation in hold cells is disabled. This is what I'm looking at. The drawing exposed via a phantom is visible in the viewer, but I'm editing the last real drawing exposed on the xsheet. Editing here either needs to be disabled or preferably phantom drawings should be editable like any other exposure.

image

> In general and in some odd ways I have mixed feelings about treating these "phantom" frames as real frames, but I can see the benefit of being able to do some stuff with them as if they were real. Kind of like my answer to beeheemooth...it's not really there but the application implicitly knows it is so it should in theory be able to. Will look into.

Yeah. I can't see any downside aside from the effort needed to implement it. Thank you for this one, it'll be really helpful!

beeheemooth:
I wonder if it is possible to make "phantom" layers in the same way, i.e. layers that will have the same set of frame numbers as the original? This would be useful in many cases where different layers need to have the same drawing order. Also, it could probably solve the problem with complex clipping masks, since the same layer could be in different "phantom" layers and different masks could be applied to them. What do you think?

artisteacher:
This is really great! Perhaps the quick toolbar should have these commands - either to replace the repeat button, or in addition to.

Is it possible for the loop to adjust with step/reframing changes? 

Also, if you have a loop but a stop frame some time later, selecting the loop frames and use one of the reframe commands - 1s, 2s or etc., it moves the stop frame that ends the loop. Before a reframe:
Screenshot 2025-05-04 at 10 29 03 PM
After reframing on 2s:
Screenshot 2025-05-04 at 10 30 03 PM
I think the stop frame should stay at the original location...

manongjohn:
I've managed to address the suggestions/request @melieconiek mentioned by making the handling of looped frames similar to implicit frames.

Because of this, it was necessary to modify many cell/level commands to treat looped frames as not being real frames.  There may be unexpected behaviors with certain operations when implicit, looped, implicit-looped frames are involved.  If so, please let me know.

Additionally, I added to modifier keys for adjusting Loop Frame Markers
- `Shift` + drag of either loop frame marker will shift both frame markers at the same time in the same direction
- `Alt` + drag of either loop frame marker will  shift both frame markers at the same time in opposite directions

melieconiek:
**Creating a loop with or modifying a loop to include empty hold frames after the last drawing in the loop will crash Tahoma.** It's not completely consistent, but I can get it to crash by wiggling the marker up and down if it doesn't.

Creating a loop with this selection will crash:
image

Attempting to extend the loop down from here will crash:
image

Crash log: [crash.zip](https://github.com/user-attachments/files/20152219/crash.zip)

---

**Creating a loop consisting of phantom frames will crash Tahoma. This shouldn't be possible.**

image

---

melieconiek:
**Many of the commands in the Cells menu work on phantom frames and have some strange behavior.**

The drawing substitution commands, and the `duplicate drawing` and `clone cells` commands work on phantom frames. All of them have odd behavior. They should ignore phantom frames and treat them as an exposure of whatever the last real frame is, correct? 

**The `merge levels` command takes into account phantom frames, I assume this intended? Other commands like the adjust, optimize, and convert commands also work. This is probably fine?**

**Loops do not survive the creation of a subxsheet, but do work inside them. This is a minor issue.**

**Loops cannot be copied with other with drawings and keyframes. This is a minor issue.**

manongjohn:
Will look into the crash you mentioned.

I have to allow loops, even on phantom cells due to the fact that you can move all exposed cells in a loop out of it, leaving an empty loop.  Keep in mind, this looping is based on scene frames, not exposed frames, much like Cell Markers are.

I purposefully allowed Duplicate frames to work on the phantom cells, but had made changes to most other cell commands to treat phantom cells as empty cells. If there is still oddity with that in mind, please let me know which command and how it is and shouldnt be behaving.

For now Drawing substitution is a bit strange.  I cant entirely prevent it because when substituting a valid cell that is preceeded by phantom cells, it becomes a phantom cell when it matches the last exposed cell in the looped range.  If a valid cell switched to a phantom cell and drawing substitution were not allowed, it would stop further substitutions. I will continue to to find a way to prevent this, but for now I am ok with the behavior.

As for merge and other commands you mentioned, i'm ok with it working if it doesnt cause phantom cells to become real cells unexpectedly and also if undo still works properly.

melieconiek:
> If there is still oddity with that in mind, please let me know which command and how it is and shouldnt be behaving.

`Clone Cells` works on phantom frames but only clones the first drawing in the selection. That's the only one that's not really working as expected.

chrymium:
I'm getting a pretty weird bug on Windows 10 where whenever i test out the new features of this build and it crashes, i lose the executable. it isn't in the recycle bin and seems to have been just outright deleted. i have to re-extract the zip file to test again.

manongjohn:
@chrymium No. Check that any anti-virus software isnt quarantining it.

manongjohn:
Not sure why it thinks that. Likely a false detection and probably should be appealed.

As you are downloading it from here, I would add T2D exe/folder to the safe/ignore list.

manongjohn:
I'm planning on merging this in the next day or so unless someone has found something that requires fixing.

I'm not planning on making loop markers move/adjust with any cell commands, at this time.

artisteacher:
I'm excited for this to move forward!

As far as automatic loop resizing with the cell commands, for now I've realized we can just place the loop in a sub-scene & do any timing refinements to the sub-scene. Or just test the timing with the loop play button before actually looping.

I do still think a stop hold (or later drawing frames) shouldn't move to the end of the "real frames" when using cell commands on the loop. Consider a 5 frame loop with a stop hold on frame 200. I can't envision ever wanting the stop hold to automatically move to frame 11 because of reframing on 2s.  Or consider a 2 frame loop on 2s that ends on frame 48. If I deliberately added new drawing frames starting at frame 49, I wouldn't want them shifted to begin back at frame 3 if I set the loop frames back to 1s.

melieconiek:
If there isn't a stop frame or new drawing to stop the loop later in a column the first few phantom exposures will not show up in the onion skin relative drawing mode.

The first two phantom drawings aren't visible in drawing relative mode. Adding a stop frame later will fix this.
image

manongjohn:
Fixed issue with onionskin not working immediately at the end of last looped range.
Fixed Reframe to not include unselected looped frames if the last frame in the selection is an exposed or implicit frame.

manongjohn:
Merging...thanks for all the feedback and patience!

PR #1829, opened 2025-05-11
Percentage spacing on pattern brushes.

melieconiek:
I changed pattern brush space setting to a percentage of the stamp width instead of an absolute value and reduced the minimum spacing. This makes the stamp spacing consistent across different stroke widths. 

Setting spacing to 0 will place the stamps end-to-end. Positive values will move the stamps away and negative values will cause them to overlap.

This change affects both vector and raster stamp images.

(Old vs New)
image image

An alpha blending issue with raster pattern brushes was also fixed.


manongjohn:
LGTM.  Thanks for the improvement!

PR #1836, opened 2025-05-15
MyPaint Brush Tilt Support

manongjohn:
This adds the new toolbar option, `Tilt` which is only available for MyPaint styles...for now.  It is available on both Raster and Smart Raster brush toolbars when a MyPaint style is selected.

Similar to `Pressure`, enabling `Tilt` will capture tilt information from pens that support it.  The tilt information is passed to the MyPaint brush engine so it can be applied to styles that are configured for Tilt sensitivity.

Example of a MyPaint style that uses Tilt:  `Classic` -> `Marker-Fat`

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -96,6 +96,8 @@ public:
   TPointD m_pos;  //!< Mouse position in window coordinates, bottom-left origin.
   double m_pressure;  //!< Pressure of the tablet pen (0.0 - 1.0) , or 1.0 for
                       //! pure mouse events.
+  double m_tiltX, m_tiltY; //! Tablet pen tilt
+  double m_rotation; //! Tablet pen rotation
 
   ModifierMask m_modifiersMask;  //!< Bitmask specifying key modifiers applying
                                  //! on the event.
@@ -114,7 +116,10 @@ public:
       , m_buttons(Qt::NoButton)
       , m_button(Qt::NoButton)
       , m_isTablet(false)
-      , m_isHighFrequent(false) {}
+      , m_isHighFrequent(false)
+      , m_tiltX(0.0)
+      , m_tiltY(0.0)
+      , m_rotation(0.0) {}
 
   bool isShiftPressed() const { return (m_modifiersMask & SHIFT_KEY); }
   bool isAltPressed() const { return (m_modifiersMask & ALT_KEY); }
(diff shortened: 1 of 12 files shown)


melieconiek:
I'm having issues drawing strokes with the pen at 45 degree angles. The angled strokes should be as wide as the up and down strokes (I'm drawing perpendicular to the angle of the pen.) This issues crops up whenever drawing at 45, 135, 225, 325 degrees. 
image

I double checked in CSP and Krita and was not experiencing this issue there.

melieconiek:
> Which MyPaint style are you using?
Fat Marker

manongjohn:
Can you try this?

Krita supports MyPaint brushes, but doesn't make it easy to import new ones, at least I didnt see an easy way to do it, however I did manage to get a new brush into it by doing the following:

1. Copy the following from `tahomastuff\library\mypaint brushes\classic`
    `marker_fat.myb`
    `marker_fat_prev.png`
2. Paste in this folder for Krita brushes: `C:\Users\yourlogin\AppData\Roaming\krita\paintoppresets`

Open Krita and switch to the newly imported brush.   Test your pen tilt and compare it with what T2D does with the same angle.

Although I get very odd and inconsistent results in Krita, when I do manage to get some consistent results with the same angle in T2D, they appear to be the same, I think.  The 1 difference I do notice is when my pen is upright, T2D draws a thin line, however in Krita it draws a wide stroke.  I'd be curious to see what you get doing the same.

FYI: None of the 7 MyPaint brushes included in Krita appear to be configured for tilt.

melieconiek:
I will try a mypaint brush in Krita. It would be nice if I know how to configure a MyPaint brush for tilt.

I _did_ test in MyPaint standalone. It does not have this issue.

manongjohn:
For some reason standalone MyPaint doesnt seem to recognize my pen on my laptop-tablet. Sees it as a mouse so I have no Pressure or Tilt when I try it there.

MyPaint properties for Tilt are listed as 
- `tilt_declination` or `Declination` : 0 = stylis parallel to tablet/90 = perpendicular
- `tilt_ascension` or just `ascension`: direction pen is pointing

manongjohn:
As far as processing tilt is concerned, I am passing tilt information from pen directly to the mypaint engine.  What it does with that is unknown.  The only thing I am doing is converting QTableEvent tilt range from -60 to 60 to MyPaint tilt range -1.0 to 1.0

manongjohn:
Merging.  Thanks for testing.

PR #1861, opened 2025-06-12
Fix raster autoclose

Abel032:
Orignal PR :[ot/5932](https://github.com/opentoonz/opentoonz/pull/5932)

- Transform input angle in autoclose to radians and correct angle application
- RasterTapeTool: Adjust angle range from 1-180 to 1-360
- Add debug micro to autoclose

#### Before

https://github.com/user-attachments/assets/bfc0422c-df00-42d4-8c10-183bccaaf3b2

#### After

https://github.com/user-attachments/assets/d2080e0d-b67e-48b9-a9ef-9d15e241d012


manongjohn:
LGTM.  Thanks for the fix!

PR #1880, opened 2025-07-11
postinst-fix

charliemartinez:
If you installed a new .deb, Tahoma2D was not being updated for users who had already installed a previous version.


charliemartinez:
And here: https://github.com/tahoma2d/tahoma2d/pull/1880/commits/9843931215b5007f6167acb85b21c22521fa74ee

I handle the case where it's a first-time installation. As it was, the find command would fail because the Tahoma2D folder wouldn't exist yet for a new user.

PR #1888, opened 2025-07-30
Stylus Settings Popup (Raster Brush & MyPaint styles)

manongjohn:
This introduces a new popup providing stylus adjustment graphs allowing for customization over how the pen responds to things like pressure and tilt.

A new configuration button provides access to the popup on the `Raster Tool` toolbar for standard styles (left), and on each MyPaint style setting (right):
image

Popups provide the following options (Left: Standard Raster Styles; Right: MyPaint Styles):
image

Some notes:
- `Tilt` (`Declination` for MyPaint) input range is 0° (horizontal, laying down) - 90° (vertical, upright). Most tablets that are tilt sensitive support up to 60°, from the vertical, which translates to 30° on the graph. 
- For standard styles, the `Pressure` and `Tilt` options in the `Size` and `Opacity` popups are independent of each other.
- MyPaint default curves are based on the original brush settings (from the `.myb` file).
- The MyPaint configurations are available anywhere a MyPaint style is used (Raster/Smart Raster) and configured for pressure and tilt sensitivity.
- Brush Presets and MyPaint style curve changes are saved in such a way that the preset/palette can still be opened in prior versions of T2D.  However, if the preset/palette is updated and saved in prior versions, the curve data will be lost.

### DIFF
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -1690,20 +1690,24 @@ QDialog #dialogButtonFrame QPushButton:focus:pressed {
 /* -----------------------------------------------------------------------------
    Preferences
 ----------------------------------------------------------------------------- */
+#StylusConfigPopup QListWidget,
 #PreferencesPopup QListWidget {
   background-color: #262626;
   alternate-background-color: #262626;
   border: 1 solid #111111;
   font-size: 13px;
 }
+#StylusConfigPopup QListWidget::item,
 #PreferencesPopup QListWidget::item {
   border: 0;
   padding: 3;
 }
+#StylusConfigPopup QListWidget::item:hover,
 #PreferencesPopup QListWidget::item:hover {
   background-color: rgba(255, 255, 255, 0.1);
   color: #e6e6e6;
 }
+#StylusConfigPopup QListWidget::item:selected,
 #PreferencesPopup QListWidget::item:selected {
   background-color: #a35293;
   color: #ffffff;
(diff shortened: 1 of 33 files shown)


melieconiek:
This is excellent. Proper brush settings and we can finally configure MyPaint bushes in Tahoma!

One thing. Linking Pressure and Tilt between the default and MyPaint brushes isn't very intuitive. If you switch from a Mypaint color to one which uses the default brush and don't want tilt applied to `Size` you'll have to enable/disable it every time you switch.

The toolbar settings shouldn't be linked between the default brush and the MyPaint brushes. It might be helpful if the MyPaint toolbar settings were saved with each palette index like the custom MyPaint settings are. _Ideally_ the MyPaint brushes should be applied via a brush settings panel and shouldn't be tied to the palette at all, but that's a much bigger change.

As for the default brush, obliqueness and angle settings would be excellent. It would allow us to take advantage of tilt with the default brush. The ability to use brush tips would be nice too. Perhaps in another PR.

manongjohn:
@beeheemooth 
I can add a button to reset all changed settings.

I think some of the settings arent meant to be sliders w/ brush dynamic graphics. I think a few od them should be checkboxes instead.  Will need to check how MyPaint handles some of these settings.

Which setting caused the crash?  I just saw 1 setting where the graph had no curve/line. Need to fix that also.

beeheemooth:
I still can't find the setting that caused the crash. I only remember that it was a checked box and I unchecked it.

manongjohn:
I think I've addressed any issues, concerns and suggestions that were brought up as well as a few issues I uncovered.

Let me know.

PR #1892, opened 2025-08-04
Smart Raster Brush Tool engine unification (Experimental - Feedback requested)

manongjohn:
_This, currently, is just an experimental PR. Whether this becomes official or not will be based on the feedback I get._

---
Similar to the Raster Brush Tool, this PR unifies the Smart Raster Brush Tool under the MyPaint brush engine, replacing the original T2D logic for handling the standard style.

The standard style is now an internally defined MyPaint style used to mimic the original standard style. For reference, it's the same configuration as the Raster standard style.  The logic for how the smart raster brush behaves is now controlled by the MyPaint engine, rather than T2D's existing logic.

`Pencil` mode adds additional settings to the internally defined MyPaint style to mimic the original `Pencil` mode.  May need a little tweeking, but primarily behaves as the `Experimental\pixel-hardink` MyPaint style.

The one difference is how `Hardness` is applied to the stroke.  I haven't quite found the best MyPaint settings to emulate the existing `Hardness` logic.  Not sure if it can be exactly replicated. Is the way `Hardness` is applied in MyPaint acceptable or is the original way preferred? Would this be a deal-breaker if it cannot be duplicated in MyPaint?

I've built this on top of the `Stylus Settings Popup (Raster Brush & MyPaint styles)` PR  so the pressure and tilt stylus settings can be configured on the Smart Raster Brush also.

Please provide constructive feedback. Thanks.

### DIFF
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -1690,20 +1690,24 @@ QDialog #dialogButtonFrame QPushButton:focus:pressed {
 /* -----------------------------------------------------------------------------
    Preferences
 ----------------------------------------------------------------------------- */
+#StylusConfigPopup QListWidget,
 #PreferencesPopup QListWidget {
   background-color: #262626;
   alternate-background-color: #262626;
   border: 1 solid #111111;
   font-size: 13px;
 }
+#StylusConfigPopup QListWidget::item,
 #PreferencesPopup QListWidget::item {
   border: 0;
   padding: 3;
 }
+#StylusConfigPopup QListWidget::item:hover,
 #PreferencesPopup QListWidget::item:hover {
   background-color: rgba(255, 255, 255, 0.1);
   color: #e6e6e6;
 }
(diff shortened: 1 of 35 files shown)


manongjohn:
Unfortunately, MyPaint, uses a line graph for curve adjustments.  I'm not sure why they went that route.  We have no option to change that from a MyPaint perspective.  It's the same way in the MyPaint application.  You would need to add more points to the curve to help "round" that out a bit.

I just tested saving and restoring a preset and it worked for me.

melieconiek:
Strokes can look blotchy when moving the pen quickly while pencil mode is on. This is not present in the current brush engine. It looks like the dabs are too far apart. This isn't apparent if pencil mode is off.
image

melieconiek:
Pencil mode seems fine now.

I don't use smart raster all that often. It appear to be work much the same with the exception of the softness setting, which will be a matter of preference. I like the look of it in this PR. If it makes adding additional features to the default brush easier than I would be ok with it.

PR #1898, opened 2025-08-23
Stylus Settings Popup (Smart Raster - Standard Brush, Vector) 

manongjohn:
Building on #1888, this PR introduces the stylus setting popup for Smart Raster (standard brush) and Vector levels.

Stylus setting allow for `Pressure` and `Tilt` customizations and is available for `Size` options only.

The 1 notable difference between Raster implementation and this PR is the graphs use segmented non-linear curves since Smart Raster (standard brush) and Vectors do not use the MyPaint brush engine.

image

Additional change:
- Modified the standard `Tilt` range from 0 .. 90 to 30 .. 90
  - QT and most tables only support up to a 60 degree tilt from vertical (30 degree from horizontal).
  - In theory, 0-30 degree from horizontal would never be possible.
- The Vector `Smooth` and `Pressure` (enabled) settings are now separated from the Smart Raster settings when saved.

### DIFF
--- a/toonz/sources/common/tproperty.cpp
+++ b/toonz/sources/common/tproperty.cpp
@@ -3,6 +3,7 @@
 #include "tproperty.h"
 #include "tstream.h"
 #include "texception.h"
+#include "tcurves.h"
 // #include "tconvert.h"
 
 void TProperty::addListener(Listener *listener) {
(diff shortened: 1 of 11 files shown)


melieconiek:
Everything seems to work as expected! Thank you for this.

PR #1915, opened 2025-09-25
Brush Tips (Raster/Smart Raster)

manongjohn:
This PR adds a new button to the Raster/Smart Raster brush tool option bar, allowing you to select and configure a brush tip to draw with. It is located on the far left and always displays the selected brush tip.

image

### Brush Tip Image Information:
- Directory: `tahomastuff\library\brush tips`. Subdirectories are supported.
- File Formats: PNG BMP TIF TIFF TGA RGB
- Dimensions: No enforced requirements, but keep them reasonably sized.  Will be desplayed as 64x64 in popup.
- Image: Any color, though usually black or white. Will be shown as-is in the popup, but colored by style.  Must have transparent background.
- Image sequences: Supported. Naming convention must adhere to T2D's image sequence filenaming format.
   
### Brush Tip Properties (disabled for default brush)
- `Spacing` - Position of next dab/stamp from the previous stamp. Default: `1`
- `Scatter` - Radius in which to place the next dab/stamp. Default: `0`
- `Rotation` - Base rotation of the stamp. Default: `0`
- `Auto Rotate` - Rotate based on pen's tip direction. Default: `Enabled`
- `Flip Horizontally` - Flip the stamp horizontally. Default: `Disabled`
- `Flip Vertically` - Flip the stamp vertically Default: `Disabled`

### Other notes:
- Raster brush no longer uses MyPaint for the standard brush styles. The original OT Raster brush engine logic was restored, fixed and updated for newer features.
- Brush tip information is saved/restored with each session and in `Presets`.
- At startup, if the previous session's selected brush tip is not found, the default brush (`Round`) will be automatically selected.
- For Smart Raster `Pencil` mode always uses the default brush tip.
- `Hardness` is disabled when not using the default brush tip.
- The brush tips that were introduced in the `FlowsPaintBrushFx` PR have been relocated to the directory noted above.

I would like to include a few more brush tips, beyond what was created for the `FlowsPaintBrushFx`. If anyone would like to submit a brush tip set they created for consideration, please let me know. Free-to-use brush tip sets can also be submitted for consideration but I need the source cited and/or license provided.


artisteacher:
This is a nice addition! I would love for each brush tip to remember its own settings. Right now when you change the settings on one tip and then switch to another, those settings are still changed. 

It would also be nice to have a preview of the changes in brush tip properties dialog. Right now you have to make a change, click out of the dialog, test the brush and then repeat to adjust from there.

artisteacher:
If you are planning on expanding this to vector, I think the functionality is already largely in place with the custom styles - it's just that you would need some better images that would actually work as tips.

manongjohn:
I think once a brush tip and the desired settings are found, I would expect users to create a preset so they can switch back to it at any time, much like with other brush setting.   With this brush tip feature and the stylus configuration stuff, maybe presets will be used more.

A preview would certainly be nice, in general, if we had a brush panel, instead of just a bar. Setting size, hardness and opacity is a guessing game right now also. :)  If I keep adding more brush settings, I may need to create a brush panel afterall.

I'm not sure this will expand to vectors, considering the custom styles already work the same way as these brush tips.  The major difference is the custom styles are not recolored.  The additional options like flipping and auto-rotation, can be added but is currently a problem due to how stroke information is stored in the PLI file. I would not add scatter.

All good thoughts though.

PR #1916, opened 2025-09-26
Minor QoL: Automate the build libtiff stage during cmake generation

parradex99:
This is an extremely minor change, but I think is convenient specially for package maintainers in other linux distros. It just builds the libtiff library for you when creating the makefiles or ninja files when you run the cmake command.

It does exactly what you would have done yourself, it be nice to see this done for you. One less step to worry about. It built without issue on my fedora box.

### DIFF
--- a/doc/how_to_build_linux.md
+++ b/doc/how_to_build_linux.md
@@ -144,17 +144,12 @@ EOF
 ```
 Note the generated file must not actually contain `$HOME`, this expands to an absolute path in the generated file.
 
-### Building LibTIFF
+<hr>
+TODO: Make it possible to use the system libtiff.
 
-TODO: make sure we can use the system libtiff instead and remove this section.
-Features from the modified libtiff are needed currently, so this isn't a simple switch.
+Tahoma2D is completely dependent on a tweaked libtiff library that's under the thirdparty directory, which is static linked. The CMake build system will compile for you, but until then, it's stuck on this modified library.
 
-```
-$ cd tahoma2d/thirdparty/tiff-4.2.0
-$ ./configure --with-pic --disable-jbig --disable-webp 
-$ make -j$(nproc)
-$ cd ../../
-```
+
 
 ### Building Tahoma2D
 
--- a/toonz/cmake/FindTIFF.cmake
+++ b/toonz/cmake/FindTIFF.cmake
@@ -1,4 +1,8 @@
 # looks for libtiff(4.2.0 modified)
+#
+# Configures and builds the modified libtiff library
+execute_process(COMMAND ./configure --with-pic --disable-jbig --disable-webp WORKING_DIRECTORY ${SDKROOT}/tiff-4.2.0)
+execute_process(COMMAND make WORKING_DIRECTORY ${SDKROOT}/tiff-4.2.0)
 find_path(
     TIFF_INCLUDE_DIR
     NAMES


parradex99:
> The script `ci-scripts/linux/tahoma-build.sh` needs to be updated to no longer build libtiff ahead of T2D build.

Did I remove too much? Or is it good?

parradex99:
@manongjohn 
It failed. Why the configure script failed, when it worked on my machine, I don't know why. Any ideas?
```
hecking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether UID '1001' is supported by ustar format... yes
checking whether GID '1001' is supported by ustar format... yes
checking how to create a ustar tar archive... gnutar
checking whether to enable maintainer-specific portions of Makefiles... no
checking for gcc... /usr/bin/ccache
checking whether the C compiler works... no
configure: error: in `/home/runner/work/tahoma2d/tahoma2d/thirdparty/tiff-4.2.0':
configure: error: C compiler cannot create executables
See `config.log' for more details
make: *** No targets specified and no makefile found.  Stop.
***** libtiff Header path:/home/runner/work/tahoma2d/tahoma2d/thirdparty/tiff-4.2.0/libtiff
***** libtiff Library path:TIFF_LIBRARY-NOTFOUND
CMake Error at /opt/hostedtoolcache/cmake/3.31.6/x64/cmake-3.31.6-linux-x86_64/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find TIFF (missing: TIFF_LIBRARY)
Call Stack (most recent call first):
  /opt/hostedtoolcache/cmake/3.31.6/x64/cmake-3.31.6-linux-x86_64/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /home/runner/work/tahoma2d/tahoma2d/toonz/cmake/FindTIFF.cmake:40 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:496 (find_package)


-- Configuring incomplete, errors occurred!
make: the '-j' option requires a positive integer argument
```

parradex99:
> The make command appears to need the `-j#` option? Compare the commands in the original instructs and build script

Right, I get and figured out the variable I deleted should have remained. I'm more puzzled why the configure step that makes the makefile failed.

ci-scripts/linux/tahoma-build.sh
```diff
@@ -26,4 +21,4 @@ cmake ../sources  $CANON_FLAG \
     -DWITH_GPHOTO2:BOOL=ON \
     -DWITH_SYSTEM_SUPERLU=ON
 
-make -j "$parallel"
+make -j$(($(nproc) < 2 ? 1 : $(nproc) - 1))
```
manongjohn, line 24: This is wrong. We dont want to build libTIFF in this script anymore, since you are trying to build it as part of CMake. The change I proposed should be in the FindTIFF.cmake file.

ci-scripts/linux/tahoma-build.sh
```diff
@@ -21,9 +16,9 @@ then
...
 
 export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
-cmake ../sources  $CANON_FLAG \
+CC="gcc" cmake ../sources  $CANON_FLAG \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DWITH_GPHOTO2:BOOL=ON \
     -DWITH_SYSTEM_SUPERLU=ON
 
-make -j "$parallel"
+make
```
manongjohn, line 24: Revert this change back to original line.

ci-scripts/linux/tahoma-build.sh
```diff
@@ -1,9 +1,4 @@
 #!/bin/bash
-# Leave one processor available for other processing if possible
-parallel=$(($(nproc) < 2 ? 1 : $(nproc) - 1))
```
manongjohn, line 3: Restore the 2 lines above

ci-scripts/linux/tahoma-build.sh
```diff
@@ -21,9 +16,9 @@ then
 fi
 
 export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
-cmake ../sources  $CANON_FLAG \
+CC="gcc" cmake ../sources  $CANON_FLAG \
```
manongjohn, line 19: As this is setting up to build T2D, this change also impacts the main build, not just libTIFF build.

Need to see if T2D build is really using ccache or not. Hard to tell from the Github logs. if so, this change would disable that causing build times to increase.

parradex99:
@manongjohn 
Okay, I think I've resolved all I can resolve. Can we test out the build to see it works? I made a test change in the yaml file, just to see if that would make a difference. I'll revert it if not.

parradex99:
@manongjohn 
Done.

manongjohn (commented):
After more thought on this approach to dealing with libTIFF, I don't think it's good practice to be compiling another library in the middle of a cmake configuration.  Cmake is meant to set up configurations in preparation of building, not build components on the fly in the middle of setup.

Sorry, but I will not be approving this.

A potential alternate approach would be to add a Cmake option to use system libTIFF, for those that do not wish to use the customized one.  This would be similar to using the included SuperLu or system SuperLu.  However, compiling with that option comes with the the understanding that doing so prevents correctly loading certain (maybe all?) TIFF files.  So if selecting this option, it must be clear in the configuration summary that TIFF files may not load properly.

parradex99:
> After more thought on this approach to dealing with libTIFF, I don't think it's good practice to be compiling another library in the middle of a cmake configuration. Cmake is meant to set up configurations in preparation of building, not build components on the fly in the middle of setup.
> 
> Sorry, but I will not be approving this.
> 
> A potential alternate approach would be to add a Cmake option to use system libTIFF, for those that do not wish to use the customized one. This would be similar to using the included SuperLu or system SuperLu. However, compiling with that option comes with the the understanding that doing so prevents correctly loading certain (maybe all?) TIFF files. So if selecting this option, it must be clear in the configuration summary that TIFF files may not load properly.

I understand your reasoning. Do you wish for me to close this or not?

PR #1917, opened 2025-09-27
Github build on macos-15-intel (Sequoia)/Fix Linux build

manongjohn:
Following has been changed

1. Homebrew has updated causing the `macos-13` builds to take several hours to install dependencies, only to fail during installation and abort the build.  Additionally `macOS-13` is being deprecated by Dec so I'm updating to the next Intel based macOS runner `macos-15-intel` (Sequoia).

Unfortunately I need to skip `macos-14` (Sonoma) as it's not Intel chip based and may require some changes somewhere to build correctly.  It was failing to link to local (not arm64) and system `superlu` library.

Will need help verifying the build as I don't have a Sequoia environment yet.

2. Fix Linux build missing Protobuf library

I will merge this one I see the builds have completed successfully.

### DIFF
--- a/.github/workflows/macOS_build.yml
+++ b/.github/workflows/macOS_build.yml
@@ -4,9 +4,15 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   macOS:
-    runs-on: macos-13
+    runs-on: macos-15-intel
     steps:
       - uses: actions/checkout@v4
+      - name: Setup cmake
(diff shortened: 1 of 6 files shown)


manongjohn:
Merging, so macOS and Linux builds will work again.

artisteacher:
Too bad about Sonoma - that's what we use at school. I'm surprised it's not intel based - the lab is all old intel macs.  

I can verify that the latest nightly runs as expected so far on Sequoia 15.4.1.

PR #1929, opened 2025-10-13
Build on macOS-14 (ARM64/M1 Runner)

manongjohn:
With a few modifications to the cmake files and build scripts, I was able to successfully build on the macOS-14 action runner which is Apple silicon based (M1).

Need help testing since I do not have a local environment to test this on.

This PR build doesn't have the Canon SDK built into it. To test if Stop Motion with Canon SDK works, please test my repo's build here: https://github.com/manongjohn/tahoma2d/actions/runs/18453887181

If the build is good, I will switch from macOS-15-intel to this for the upcoming release.

### DIFF
--- a/.github/workflows/macOS_build.yml
+++ b/.github/workflows/macOS_build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   macOS:
-    runs-on: macos-15-intel
+    runs-on: macos-14
     steps:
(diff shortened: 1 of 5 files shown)


manongjohn:
Closing this, for now. There is a PR in OT changing the way macos is built on ARM/M1 machines which might work. Will also wait to see what other changes OT might be making to compile and work more natively on the new architecture.

PR #1930, opened 2025-10-13
Existing scenes list in Startup Popup (OpenToonz Port)

manongjohn:
After a number of requests, this PR ports the `Existing scenes list in Startup Popup` feature created by @justburner  from OT.

I've had to modify the OT commit to handle differences in code between OT and T2D.  I followed up with additional commits to style the Startup Popup for T2D and fixed the coloring of the New Scene icon to change with the theme.

image

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -1306,8 +1306,8 @@ void StartupScenesList::clearScenes() {
 void StartupScenesList::addScene(const QString &name, const QString &path) {
   QPixmap pixmap;
   if (path == ":")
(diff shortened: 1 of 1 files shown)


beeheemooth:
I'm not particularly thrilled about this PR in Tahoma, and I'll explain why. From my observations, some crashes in OT (specifically, crashes that occur upon program startup) are likely related to the display of corrupted scene icons. This is similar to crashes that occur when corrupted image icons are displayed in the Level Strip. But while in the latter case you can find the corrupted level by elimination, in this case you won't even be able to launch the program or understand what's going on. What are your thoughts on this? Is there a way to disable the display of scene icons, i.e., switch to the old window interface?

PR #1931, opened 2025-10-13
Add color and fill support to raster pattern brush.

melieconiek:
This adds colorization support and the ability to use pattern brush color indexes as fills. This allows raster pattern brushes to work with column color filters and onion skins -- these did not work before.

Setting the color to black will have no colorization effect.

https://github.com/user-attachments/assets/c2a85855-dd9a-4dcf-82a3-9834b0afeac4


melieconiek:
Thank you taking a look @manongjohn.

I restored the default spacing from before, changed the id, added color saving/loading, and added logic for handling the old save data. This will break loading in previous versions, is that ok?

PR #1934, opened 2025-10-15
Fix(autoclose): Add intersection check for spotRearchTwoPoints

Abel032:
A port from [ot/#6196](https://github.com/opentoonz/opentoonz/pull/6196)
### [Add intersection check for spotRearchTwoPoints](https://github.com/tahoma2d/tahoma2d/commit/7ce36efff5b046954a09932f794a546153a42ea4)
Added intersection detection specifically for the initial two endpoints search (triangle area search).
Before drawing a connection during the two endpoints search (within the triangle area defined by current endpoint, search angle, and distance), the algorithm now checks if the new line would intersect with any existing lines.
If an intersection is detected, the line is not drawn and the point is passed to the subsequent sector search phase.
Sector search remains unchanged:
The sector search operates by rotating rays within a defined angular range to find the nearest ink pixel.
It connects to the closest suitable point found during this rotational scan, not limited to endpoints only.
| Expected Result | Old Result |
| ----- | ----- |
| 图片 | 图片 |

### DIFF
--- a/toonz/sources/toonzlib/autoclose.cpp
+++ b/toonz/sources/toonzlib/autoclose.cpp
@@ -106,6 +106,7 @@ public:
   int exploreTwoSpots(const TAutocloser::Segment &s0,
                       const TAutocloser::Segment &s1);
   int notInsidePath(const TPoint &p, const TPoint &q);
+  bool hasInkBetween(const TPoint &p1, const TPoint &p2);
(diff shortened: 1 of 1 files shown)


manongjohn:
LGTM.  Thanks for brining this fix/enhancement to T2D!

PR #1941, opened 2025-10-25
Fix bounds issue

Abel032:
A port from OT.

This PR addresses these visualization problems:
- Scene Viewer with `Preview` toggled: [1-Pixel Clipping on Right/Up Edges](https://github.com/tahoma2d/tahoma2d/commit/6c97a4d4f6d5bce0aca52821ece6fcfd46f85ff8)
- Scene Viewer in normal: [TLV Bounds not correct](https://github.com/tahoma2d/tahoma2d/commit/de0b071332f787d63aa5abe97d110f80eb72de6f)
- Scene Viewer in Level Editing Mode(Raster Level) : [Raster Background not correct](https://github.com/tahoma2d/tahoma2d/pull/1941/commits/d79a27d5e789e4495be1c90187d88636cb2d8e6f)

And add:
- Scene Viewer in Level Editing Mode(Vector Level) : [Draw background rectangle for vector level in levelEditing mode](https://github.com/tahoma2d/tahoma2d/pull/1941/commits/022df925a74102c56c76c05bf3839d39657bcf2a)
(This can't be seen by default, must change the bg color or turn on black BG check)

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -1615,6 +1615,7 @@ void SceneViewer::drawCameraStand() {
 
     TToonzImageP ti  = image;
     TRasterImageP ri = image;
+    TVectorImageP vi = image;
     if (ti) {
       TRect imgRect(0, 0, ti->getSize().lx - 1, ti->getSize().ly - 1);
       TRectD bbox = ToonzImageUtils::convertRasterToWorld(imgRect, ti);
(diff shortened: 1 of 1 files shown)


manongjohn:
LGTM. Thanks for porting these fixes over. I appreciate it very much.

Just a side thought...

Using the vector stroke bounds to show the black background, when enabled, seems weird to me as it doesn't necessarily line up with where the true center of the canvas is even though vectors have no bounds.  More of an issue depending on where the strokes are in relation to the virtual center of the canvas.  Kind of disconcerting when switching between scene-level edit modes and when adding new strokes.

I debated adding a toggle-able placeholder box to show where the camera area would be for reference.  I wonder if that might be a better area to show the black bg check.

Another thought would be when enabling black bg check on vector levels while in level edit mode, it should just override the normal level background color with black, rather than limiting it to the stroke boundaries.

Anyway...something to think about for another time....

PR #1942, opened 2025-10-25
Fix(Scene Depth/Quicktoolbar): Invisible after startup

Abel032:
A port from OT.

This issue usually happen when the first room contains a vertical xsheet panel (not timeline).
This PR replaces `visibleRegion().boundingRect()` with `geometry()` to make the bars' width can be initialized when first show.

### DIFF
--- a/toonz/sources/toonz/xsheetviewer.cpp
+++ b/toonz/sources/toonz/xsheetviewer.cpp
@@ -437,7 +437,7 @@ void XsheetViewer::positionSections() {
 
   if (Preferences::instance()->isShowQuickToolbarEnabled()) {
     m_toolbar->showToolbar(true);
-    int w = visibleRegion().boundingRect().width();
+    int w = geometry().width();
     m_toolbarScrollArea->setGeometry(0, 0, w, XsheetGUI::TOOLBAR_HEIGHT);
     m_toolbar->setFixedWidth(w);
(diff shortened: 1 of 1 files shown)


Abel032:
@manongjohn 

Timeline always works on my computer:
图片
But for xsheet, it shows part of toolbar after startup:
图片

In OT it fixed the issue but..
In T2D it just extended the width:
图片

And one clue is(Not the case above), the docked locator might also affect the initialization of toolbar's width...

I actually doesn't understand why we need to use `visibleregion()` here, the doc says it's not necessary to call this function.
ref: https://doc.qt.io/archives/qt-5.15/qwidget.html#visibleRegion

Would be appropriate if you can find the real reason.
Since this PR doesn't resolve the problem, let me close this.

manongjohn:
FYI, I arranged my `2D` room to match yours and was able to duplicate the issue with the Xsheet in T2D. 

So your fix partially worked in that you only applied the fix to the Breadcrumbs area  It also needs to be added to the xsheet toolbar, which is a few lines above it

manongjohn:
LGTM.  Thanks for porting this fix over.  Will merge soon.

I change the title of the PR just to put it in T2D terms, plus it fixed the same issue I was having with the Timeline when certain things were docked next to it.

PR #1970, opened 2025-12-03
Shift markers with Cell Insert/Delete

manongjohn:
This allows Loop Markers, Cell Marks and Play Markers to shift along the timeline/xsheet depending on how and where cells are inserted/deleted.

Actions
- Insert Frame
- Remove Frame

Actions shifting Loop Markers and Cell Marks only
- Insert Cells
- Remove Cells
- Cut Cells
- Paste Cells
- Paste as a Copy
- Reframe (all variations)
- Step (all variations)
- Each (all variations)
- Swing
- Repeat
- Time Stretch
- Auto Input Cell Number (Insert mode)
- Cell hold extension
- Shift + Drag/Drop from Level Strip (inserts cells)
- Shift + Ctrl + Click selection (inserts duplicate of selection)

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -125,12 +125,16 @@ namespace XshCmd {
 class InsertSceneFrameUndo : public TUndo {
 protected:
   int m_frame;
+  int m_playR0, m_playR1, m_playStep;
 
 public:
-  InsertSceneFrameUndo(int frame) : m_frame(frame) {}
+  InsertSceneFrameUndo(int frame) : m_frame(frame) {
+    XsheetGUI::getPlayRange(m_playR0, m_playR1, m_playStep);
+  }
 
   void undo() const override {
     doRemoveSceneFrame(m_frame);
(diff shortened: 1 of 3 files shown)


DarrenTAnims:
> > Does this move the play markers?
> 
> No. I can look into though.

That would be amazing. Someone asked me about this the other day. They were using the insert frame command on the last image in the scene and they expected the length of the scene to extend, which seemed reasonable. I'd not considered it before I read the question, but it would make sense to do so.

manongjohn:
Not sure I agree with this. Similar to Navigation Tags, Play Markers are based on overall scene frames, not individual layers.

I will definitely update Insert/Remove Frame to adjust play markers accordingly.

We can continue to discuss insert/remove cells impacting play markers on Discord.

manongjohn:
> I've noticed that when your loop starts at first frame the starting brackets don't move along with other frames

That was an intentional choice.  I view the start loop marker like a wall between cells that you push from the left to move, not pull from the right.

I'll leave it up for discussion on what is the preferred outcome. If the general consensus is it should pull when you are on the cell with a start loop marker, I will change it.

PR #1972, opened 2025-12-05
Stopmotion room changes

charliemartinez:
stop-motion-room-changes

Since the Motion Path is accessible via its tab, I suggest replacing that panel — which isn’t scalable — with the new preproduction board, which is scalable and very handy to have at hand.


manongjohn:
Out of curiosity, how does the Preproduction Board get used in the context of this room?

Normally the preproduction board requires drag/dropping files into it. It's not dynamic in that it updates automatically if the scene changes or the level changes.

If there are any StopMotion users available, I suggest providing input on this change.

charliemartinez:
@manongjohn As a stop-motion user, I find it very comfortable to create scenes with a single click, which is why I usually use that panel. It's also practical for the animator [Isabel Macías](https://quirinux.org/isabel-macias/). I've never dropped files into it — how does that work?

manongjohn:
> I've never dropped files into it — how does that work?

Just open a browser and drag/drop scene and image files into it.

I'm still very unsure about this change, as such, I will not be including this in the upcoming release so it can be further evaluated and discussed for the following release.

Personal choices and workflows aside, I am concerned switching from motion path to preproduction board will not make sense to new users to T2D and stop motion. It may actually confuse them.

Additionally, If the sole purpose is to have have access to the `Create New Scene` button, I feel it takes up space unnecessarily for something you will not potentially use as often as Motion Path when working on a scene. You can just as easily add a button to the Xsheet Quicktool Bar for `New Scene`.  We can potentially add the same button/action on the Stop Motion Panel which I would be more inclined to do.

Just my thoughts.

PR #1976, opened 2025-12-07
Fix key as last scene frame

manongjohn:
This fixes an issue where play back doesn't treat the last key as the end of the scene when it's made past the last exposed frame.

This usually happens when the key is the last thing created just before play back.  It wasn't updating the viewer's controls with the new end of scene frame.


beeheemooth:
Cool! It also works within sub-scenes.
I see this wasn't that easy to fix. 
Thanks for your work!

PR #1980, opened 2025-12-13
Update translation files for v1.6

manongjohn:
In preparation of the T2D v1.6 release, currently targeting Feb 1st, I've updated the translation files and tip files with the latest changes so they are ready for translation updates.

Here are the translation and tip files that need translations for those interested in helping to update them:
[translations.zip](https://github.com/user-attachments/files/24361592/translations.zip)

Note that the tip files still say 1.5. I will change this later with a separate PR when I update all version numbers.

You do not need to generate the QM files. I will do that just prior to the release.

Please submit translations by Jan 26th.

I will merge this PR to update the master in a day or so to allow those familiar with git to pull down the updated repo and submit PRs.

For those that don't want to go the PR route, feel free to open an Issue or add a comment on this PR with updated translation files even if this is closed and I'll make sure they get added.

Thanks in advance!

@beeheemooth @gab3d @pedrobbs00 @aweispace: if you are available to translate for us again, it would be much appreciated!


gab3d:
Also shown in the timeline layer name:

imagen

gab3d:
I now see that after renaming the layer in the timeline:
- it shows the accented character correctly.
- it only shows the name once on the node, and not twice as when the node was first created.

imagen

gab3d:
@manongjohn 
Hi,
I've adapted all the Spanish "Ino Fx" docs PDFs to the new translated strings.
Also, I've caught a pair of errors in my previous translations, that I've now fixed.
Thx in advance! :)

[__translations-Spanish_1.6 (newdocs_n_fixes).zip](https://github.com/user-attachments/files/24601211/__translations-Spanish_1.6.newdocs_n_fixes.zip)

PR #1983, opened 2025-12-21
Add marker and soft round vector styles.

melieconiek:
image

Adds two new custom vector styles, Marker and Soft Round. Marker is a simplified version of standard vectors that can overlap with itself. It's much cheaper than standard vectors, especially when semitransparent or when using onion skins. If opaque it looks identical to standard vectors in a output render. It supports both lines and fills. If you have performance issues with standard vectors give this a try.

Soft Round is meant to simulate the look of a simple soft round raster brush. The hardness and endcap fade is configurable. It looks a little odd at higher line widths but is nearly indistinguishable from a real soft round for most linework. It only supports lines like most custom vector styles. It's also cheaper than standard vectors or raster pattern brushes, so it can be used to get a soft brush look without the performance impact.

EDIT: I'd recommend enabling `Break` on the vector brush tool to improve the look of joins as well.

### DIFF
--- a/toonz/sources/colorfx/colorfx.cpp
+++ b/toonz/sources/colorfx/colorfx.cpp
@@ -66,6 +66,9 @@ void initColorFx() {
 
   add(new TMatrioskaStrokeStyle());
 
+  add(new TMarkerStrokeStyle());
+  add(new TSoftRoundStrokeStyle());
+  
 #ifdef _DEBUG
   add(new OutlineViewerStyle());
 #endif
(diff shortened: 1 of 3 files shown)


manongjohn:
> Unless it's needed, I'm not going to update soft round to be visible at 0 thickness .

Did you mean when `Show Lines with Thickness 0` is enabled or something else?

manongjohn:
Interesting.  Is that also true if only part of the line is 0 thickness, like in the middle?

Reason I ask is because on a standard stroke, when you make a portion of the stroke 0 thickness, it still renders that portion as if it had thickness.   I think it might have to do with how a standard stroke is switched to an outline and essentially filled, but somehow, even with 0 thickness it still draws (fills?) it.   Need to look into that further and see if how the pattern styles work can help me fix that.

manongjohn:
LGTM.  I've decided to add this to the 1.6 release.

Originally I had locked down for the release and wasn't accepting anything new but i had completely forgotten about doing a BETA release. With the push of the release for a number of fixes and just recently remembering to do a BETA release, I feel it's best to add this now.

Sorry about the wait.  Thanks for your patience.

PR #1988, opened 2025-12-26
Switch appimage builder from linuxdeployqt to linuxdeploy

ghost:
It's not ready yet, still needs a few things fixed.

This project should switch to linuxdeploy away from linuxdeployqt because of two reasons: Firstly, LD supports wayland (It bundled in the appimage output stuff like libwayland-egl.so, so I think it's working so far) and linuxdeployqt isn't even maintained anymore. There's still a few rough around the edges issues such as the AppRun script being replaced (The old one is renamed as a symlink), so some additional changes are necessary. I think this will be for the best, because distros like fedora are already phasing out x11 (not to mention, GNOME and KDE are stripping out the code as well), so to avoid xwayland and have native HDR and other features wayland provides, native wayland must be supported.

### DIFF
--- a/toonz/sources/scripts/AppRun
+++ b/toonz/sources/scripts/AppRun
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-export QT_QPA_PLATFORM="wayland;xcb"
 
 DIR="`dirname \"$0\"`" 
 DIR="`( cd \"$DIR\" && readlink -f $(pwd) )`"


manongjohn:
Fyi, I will not be reviewing this until after the next release.

As far as your own testing goes, you dont need to submit a PR to create a test build. When you forked T2D, the builds are automatically run in your fork whenever you create a branch in your fork and push to it.  Check your fork's Actions for builds and test the artifact these. Submit a PR when you've confirmed it is correct.

As far as wayland integration is concerned, I think its more than just including qtwayland library and requires more investigation as the preliminary work I had done on it would not work properly. Local builds sort of worked, but not when I modified the build scripts to include relevant libraries though I was manually including things and might have missed something.

splashmeganine:
> Fyi, I will not be reviewing this until after the next release.
> 
> As far as your own testing goes, you dont need to submit a PR to create a test build. When you forked T2D, the builds are automatically run in your fork whenever you create a branch in your fork and push to it. Check your fork's Actions for builds and test the artifact these. Submit a PR when you've confirmed it is correct.
> 
> As far as wayland integration is concerned, I think its more than just including qtwayland library and requires more investigation as the preliminary work I had done on it would not work properly. Local builds sort of worked, but not when I modified the build scripts to include relevant libraries though I was manually including things and might have missed something.

When I made the PR, I was convinced that I was about 95% done and was just a few very minor things to fix. Might have been due to some mental fatigue, but in the end, I think I made this prematurely as you mentioned.

I'm going to keep it open, you can merge whenever you wish to, but I rather keep it open than having more than one PR listed in the repo.

splashmeganine:
I'm happy to confirm it works. The new AppImage modified build system works as it should,, I don't see anything out of the ordinary right now.  I'll wait until february for this to be reviewed.

PR #1997, opened 2026-01-04
Fix vector fill loss due to ghost grouping changes

manongjohn:
This fixes an issue where vector fills may get lost due to grouping changes and stroke order changes.

Background:
Fillable regions are determined by stroke intersections of strokes within the same group.  Ungroup strokes are assigned a "ghost group" which is influenced by existing groups and stacking order.  When a group is made in the middle of ungrouped strokes,  the group strokes are moved to be near each other causing ungrouped strokes below it will have a different ghost group than the ones after the new group.

Issue:
The intersection logic treated ghost groups as unique groups.   Creating groups in the middle or even changing the stacking order of groups would cause ghost groupings to change, potentially breaking valid fillable regions for ungrouped strokes.

Resolution:
In order to keep valid fillable regions for ungrouped strokes, I have modified the logic to treat all ghost groups as being in the same group (the ungrouped group).  This keeps intersections among ungrouped strokes maintained regardless of where groups are in the stacking order.

### DIFF
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -2502,7 +2502,15 @@ void addIntersection(IntersectionData &intData, const vector<VIStroke *> &s,
 
   point = s[ii]->m_s->getPoint(intersection.first);
 
-  for (p = intData.m_intList.first(); p; p = p->next())
+  int gid1 = s[ii]->m_groupId.m_id[0];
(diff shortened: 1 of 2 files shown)


beeheemooth:
I tested this a bit. It seems that deleting strokes or undoing the operation after removing a stroke from a group causes a crash.

Steps:
1. Create an intersection of two vectors (X) and group them together.
2. Enter the group, select one stroke, and ungroup it.
3a. Press Ctrl+Z and you'll get a crash.
3b. Exit the group and delete all strokes—crash again.

PR #2030, opened 2026-01-30
Fix fill loss due to ghost grouping changes

manongjohn:
This restores the reverted changes made in #1997 with corrections for handling autoclose strokes.

From the original PR.....

This fixes an issue where vector fills may get lost due to grouping changes and stroke order changes.

Background:
Fillable regions are determined by stroke intersections of strokes within the same group. Ungroup strokes are assigned a "ghost group" which is influenced by existing groups and stacking order. When a group is made in the middle of ungrouped strokes, the group strokes are moved to be near each other causing ungrouped strokes below it will have a different ghost group than the ones after the new group.

Issue:
The intersection logic treated ghost groups as unique groups. Creating groups in the middle or even changing the stacking order of groups would cause ghost groupings to change, potentially breaking valid fillable regions for ungrouped strokes.

Resolution:
In order to keep valid fillable regions for ungrouped strokes, I have modified the logic to treat all ghost groups as being in the same group (the ungrouped group). This keeps intersections among ungrouped strokes maintained regardless of where groups are in the stacking order.

### DIFF
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -2502,7 +2502,17 @@ void addIntersection(IntersectionData &intData, const vector<VIStroke *> &s,
 
   point = s[ii]->m_s->getPoint(intersection.first);
 
-  for (p = intData.m_intList.first(); p; p = p->next())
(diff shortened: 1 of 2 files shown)


melieconiek:
I think this is working. I had one crash but I don't believe it's related. I wasn't able to replicate it anyways.

PR #2039, opened 2026-02-07
Fix sound extender targeting when zoomed

manongjohn:
This fixes targeting the sound level extender (yellow bars) when zoomed out at certain amounts due to the targeting area being incorrectly resized based on zoom level.

Additionally, the targeting area is extended to the edge of the cell when in no drag bar mode.


melieconiek:
Seems to be fixed now. It's still very difficult to grab the extender bar at higher resolutions, but it always works.

PR #2061, opened 2026-02-19
Updates to Animation Tool option bar

manongjohn:
This enhances the Animation Tool's option bar as follows:

image

`Set Key` button
----
It can be used to set/clear only the channels related to the animation mode if `Global Key` is not enabled.
- Position: X/Y or posPath (for motion path) depending on what is displaed, Z, SO
- Rotation: Rotation
- Scale: Scale H, Scale V, Scale
- Sheer: Sheer H, Sheer V
- All: All the above except X/Y or posPath depending on what is displayed.

Key icon will changed based on animation mode and how many channels related to that animation mode are keyed.

Color coded background field value
----
If a field is key'd or has a value that is interpolated between keys, the field's background will be colored similar to the color scheme found in the Function Editor

Interpolation dropdown
----
A dropdown to select the interpolation to use for new keys. Note: New keys are channels that are not interpolated values.  These are typically keys created after the last key or before the 1st key on that channel.  The dropdown will become disabled whenever it cannot be applied to at least 1 new key.

This dropdown also updates `Preferences` -> `Animation` -> `Default Interpolation`

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -279,8 +279,7 @@ class ArrowToolOptionsBox final : public ToolOptionsBox {
 
   // Flip buttons
   QPushButton *m_hFlipButton, *m_vFlipButton, *m_leftRotateButton,
-      *m_rightRotateButton, *m_setNoKeyButton, *m_setPartialKeyButton,
-      *m_setFullKeyButton;
+      *m_rightRotateButton, *m_setKeyButton;
 
   // enables adjusting value by dragging on the label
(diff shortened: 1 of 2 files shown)


beeheemooth:
The problem is now solved, thank you!
Is it possible to stop the flickering of the input window fields' color changes while the animation is playing? It's annoying.

manongjohn:
> The problem is now solved, thank you! Is it possible to stop the flickering of the input window fields' color changes while the animation is playing? It's annoying.

Will stop the flickering.  Are the field colors useful?  If not, I have no problem removing it.

PR #2104, opened 2026-04-02
Fixes for ease handles and interpolation management from timeline/xsheet 

manongjohn:
This addresses the following issues as it relates to timeline/xsheet ease handles and interpolation:
- Fixed issue where handles were hidden or shown when it shouldn't due to incorrect storage of ease handle information.
- Fixed issue where handles were hidden due to multiple animated channels with mixed interpolation types
- Fixed issue where some ease values were not restored on undo of an interpolation change
- Allow changing interpolation type from timeline/xsheet for mesh keys
- Show all interpolation options in the timeline/xsheet context menu instead of hiding the "active" interpolation.
- Fixes setting the correct interpolation type information at end of channel keys
- Autocorrects the last channel key on loading to have the correct internal default interpolation type (Linear)

NOTE: There are valid reasons why ease handles may not show. Here are the conditions for displaying:
- Shows only for `Speed In/Out` and `Ease In/Out` interpolation types.
- For keys with multiple animated channels
  - 1 or more `Speed In/Out` or 1 or more `Ease In/Out` will show ease handles. A mix of both will not.  Other interpolation types are ignored.
  - The `Speed Out` (First Speed Handle)  or `Ease Out` values must all match on the left key
  - The `Speed In` (Last Speed Handle) or `Ease In` values must all match on the right key

### DIFF
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -1195,6 +1195,12 @@ is >> m_imp->m_defaultValue;
     }
   }
 
+  if (!m_imp->m_keyframes.empty()) {
+    // For loader scenes, ensure the last keyframe's type is set to Linear
+    if (m_imp->m_keyframes.rbegin()->m_type != TDoubleKeyframe::Linear)
+      m_imp->m_keyframes.rbegin()->m_type = TDoubleKeyframe::Linear;
+  }
+
   m_imp->notify(TParamChange(this, 0, 0, true, false, false));
 }
 
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -804,6 +804,7 @@ void TDoubleParam::setKeyframes(const std::map<int, TDoubleKeyframe> &ks) {
 void TDoubleParam::setKeyframe(const TDoubleKeyframe &k) {
   DoubleKeyframeVector &keyframes = m_imp->m_keyframes;
   DoubleKeyframeVector::iterator it;
+  bool atEnd = false;
   it = std::lower_bound(keyframes.begin(), keyframes.end(), k);
   if (it != keyframes.end() && it->m_frame == k.m_frame) {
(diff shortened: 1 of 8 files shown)


beeheemooth:
I did a quick test, and everything seems to be working fine.

The only thing I noticed is that sometimes (in Plastic Curves), you can slightly change a parameter's handle (you have to move it up and down), but the arrows don't disappear in the timeline. And when you move the arrows, the changed parameter also changes. Is there a threshold for how much movement is acceptable for handles?

manongjohn:
> I did a quick test, and everything seems to be working fine.
> 
> The only thing I noticed is that sometimes (in Plastic Curves), you can slightly change a parameter's handle (you have to move it up and down), but the arrows don't disappear in the timeline. And when you move the arrows, the changed parameter also changes. Is there a threshold for how much movement is acceptable for handles?

Not quite sure i understand what you are seeing. Can you provide a short video?

beeheemooth:
(At first, I couldn't reproduce the glitch, but then I noticed I was trying to do it in nightli-build, lol).

The glitch happens not only with the Plastic Tool, but also with the Animate Tool.

https://github.com/user-attachments/assets/cc566e41-9412-40b5-aeff-8f0b00940f34

beeheemooth:
Here are some more nuances.


https://github.com/user-attachments/assets/624af487-7b5e-48b2-b39c-721261a40b0c

manongjohn:
Can supply scene for me to debug with?  If needed, DM them to me in Discord.

manongjohn:
Fixed a some issues with the new logic that was causing the behavior beeheemooth noted.

Also found and fixed some other cases that caused the ease handles to either not show or disappear as key and/or interpolations were manipulated.  This was mainly due to how the first and last channel keys were being updated and saved, though there was no real impact to the animation.  Management of the first and last key's information is corrected going forward. For loading older scenes, logic has been added to auto-correct the last channel key it if doesn't contain the correct information expected for being the last channel key.

Will leave this up for a bit for further testing.

manongjohn:
But of course, I found an issue with the recent changes, surrounding copy/paste..  Working on fixing.

PR #2117, opened 2026-04-12
Animate drawing substitution

manongjohn:
This PR adds a new feature that allows you to animate drawing substitution between animation keys based on interpolation type. It was originally started by @conicsectionn but I took over after they stopped development to focus on other non-T2D things.

The Animate tool now has a new `Drawing #` option which allows you to set a key for which Drawing # to show at the current frame location.  Click-drag left and right on the canvas to change the drawing that is to be shown at that key. 
image

Additionally, in `All` mode, an option was added to the Animate widget for Drawing #.  Select it and click-drag left and right to change the drawing number. 
image

When there are 2 Drawing # keys placed on the timeline, drawings between the 2 keys are automatically placed between them based on the interpolation used and the number of drawings that exist for that level.

Notes:
- To use this feature you need a drawing level with multiple sequentially numbered drawings.
- Enabling `Global Key` or using the general `Set Key` will never set a channel key for Drawing #.  Drawing # channel keys can only be created by copy/pasting existing Drawing # keys, by using the Animate Tool's Drawing # mode, or by editing the Drawing # channel in Function Editor.
- Setting Drawing # keys around already exposed frames will cause those frames to be removed due to the automatic drawing substitution logic.  Undo should restore them if done in error.
- Use of this feature will prevent the scene from opening in prior versions of T2D

https://github.com/user-attachments/assets/3cd9cb13-8961-408d-ad52-61eb18da3550

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1084,6 +1084,7 @@
   <item>W_ShearH		"Shear H"		</item>
   <item>W_ShearV		"Shear V"		</item>
   <item>W_Scale			"Scale"			</item>
+  <item>W_DrawingNumber			"Drawing #"			</item>
   
 <!------------------------------ Ino FXs ------------------------------------------->
 
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -479,7 +479,6 @@ void TDoubleParam::copy(TParam *src) {
   if (!p) throw TException("invalid source for copy");
   setName(src->getName());
   m_imp->copy(p->m_imp);
-  
   m_imp->notify(TParamChange(this, 0, 0, true, false, false));
 }
 
@@ -1201,8 +1200,6 @@ is >> m_imp->m_defaultValue;
     }
     if (!is.matchEndTag()) throw TException(tagName + " : missing endtag");
   }
-
-
   if (m_imp->m_keyframes.empty() && !is.eos()) {
     // vecchio sistema (prima 16/1/2003)
     while (!is.eos()) {
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -663,6 +663,8 @@ double TDoubleParam::getValue(double frame, bool leftmost) const {
     if (convertUnit) value = a->convertFrom(m_imp->m_measure, value);
   }
 
+  if (getName() == "W_DrawingNumber" && value < 0) value = 0;
+
   // if (cropped)
   //  value = tcrop(value, m_imp->m_minValue, m_imp->m_maxValue);
   return value;
@@ -738,6 +740,9 @@ void TDoubleParam::setKeyframe(int index, const TDoubleKeyframe &k) {
   TActualDoubleKeyframe &dst        = keyframes[index];
   TActualDoubleKeyframe oldKeyframe = dst;
 
+  int minFrame = std::min(oldKeyframe.m_frame, k.m_frame);
+  int maxFrame = std::max(oldKeyframe.m_frame, k.m_frame);
+
   (TDoubleKeyframe &)dst = k;
   dst.updateUnit(m_imp->m_measure);
 
@@ -747,7 +752,7 @@ void TDoubleParam::setKeyframe(int index, const TDoubleKeyframe &k) {
   if (dst.m_type == TDoubleKeyframe::File)
     dst.m_fileData.setParams(dst.m_fileParams);
 
-  m_imp->notify(TParamChange(this, k.m_frame, k.m_frame, true, false, false));
+  m_imp->notify(TParamChange(this, minFrame, maxFrame, true, false, false));
 
   assert(0 == index || keyframes[index - 1].m_frame < keyframes[index].m_frame);
   assert(getKeyframeCount() - 1 == index ||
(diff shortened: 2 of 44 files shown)


melieconiek:
The only way I can trigger a change is by entering an number in the `Drawing #` field. Dragging the animate tool widgets, making changes anywhere on the xsheet or function editor will not make changes. Those changes will be reflected only if a number is entered in the animate tool option bar.

manongjohn:
> The only way I can trigger a change is by entering an number in the `Drawing #` field. Dragging the animate tool widgets, making changes anywhere on the xsheet or function editor will not make changes. Those changes will be reflected only if a number is entered in the animate tool option bar.

Are you working with a brand new level or loading an existing one? I just discovered the behavior you mentioned when I started a brand new level, added a few drawings and tried to set a key and change the drawing.  Will look into that.

If it is still doing what you are seeing with an existing level, please provide video so I can see what you are doing/seeing.

melieconiek:
> Are you working with a brand new level or loading an existing one? 

Yes, I created a new drawing in a new, unsaved scene and tried fiddling with it. We're probably experiencing the same issue.

beeheemooth:
Sorry, I have very little time for testing right now, but I couldn't ignore this PR and will look into it further later.
I also noticed the manipulator widget not working (I didn't create any new drawings after applying the function).
For some reason, the feature creates a non-existent frame 0000. You can draw in it... But what about the invalid frame 0000?
The last key created by changing the "Drawing" parameter isn't taken into account when playing the animation (this has already been fixed; apparently, this is the old way of creating the key).

manongjohn:
The invalid frame 0000 hasn't been an issue for years.  The feature allows for frame sequences starting with 0 since levels imported from other software may start with that.

Until I fix the issue with this feature and newly created levels, I suggest testing this with imported levels.

manongjohn:
Ok, the issue with newly created levels and linux/macos biuld failures should now be fixed.

beeheemooth:
Wait... I'm a little confused about the practical application of this feature. It works well, and I haven't noticed any critical errors yet.

However, it seems that duplicating a drawing during work should no longer occur immediately after the original drawing  (in the Level Strip), as this breaks the order of phases (drawings), and when reusing the feature, the drawings will be placed in a different order with the same numbering.

It's likely that duplicating a drawing  after the original wasn't the best idea (I admit, it was my idea, but who knew a feature like this would appear), and duplicates are best placed AFTER all drawings already in the Level Strip. This is necessary to avoid breaking the existing frame range in the timeline. Generated frames (duplicates or blank drawings) should also, in theory, be created at the end of the Level Strip, not between other drawings.

However, even this still raises many questions. For example, this:
The fact that we can place images between two keys is great (let's assume the first problem I described above is solved; let's say it isn't).
But if you have three keys, then when you edit the middle one (that is, when replacing images), the interpolation changes in both directions (between the first and second keys and between the second and third). Theoretically, this should look smooth, but not when you have images in the Level Strip depicting different, sometimes opposing, movements or phases.

Questions arise more about the workflow and the logic of using the feature in practice than about its technical stability and performance.

manongjohn:
Good questions and thoughts. As I didn't originally develop this, I dont really have any answers for you.. I just made sure it played nicely in T2D. 😄

 I understood it would help with cutout/rigging form of animation, I think. Seemed like ab interesting feature. Made me work more with keys, interpolations and function editor and seeing of the issues with their use.

melieconiek:
This will be very useful, thank you for working on this. If you want to create a more advanced drawing replacement rig you'd need to use a controller and use a expression in the function editor to take it's x/y/z/rot values and use them select the correct drawing numbers. An explicit controller object would make that easier. I'm not sure how that would be implemented.

manongjohn:
Fyi, I plan on merging this later today unless someone found something that must be addressed.

beeheemooth:
I thought a lot about the workflow logic for this feature, but (as usual) didn't have time to finish it.

Basically, the main idea is that the feature will work more or less adequately if we reinstate the ability to create duplicates AFTER all drawings in the Level Strip (that is, at the end of the Level Strip), as it was before. The duplicate numbering should, of course, follow the sequence (regardless of its location on the timeline).

In this case, the drawings in the Level Strip will at least be separate, rather than chaotic, sequences, and they can be manipulated within a separate interpolation.

But if you need to add a duplicate drawing to one of the specific sequences (of the same level), then adding a duplicate at the end doesn't solve the problem of editing the interpolation.

In this paradoxical case, this feature should probably include an option to insert or duplicate a drawing between drawings with a number sequence on the timeline (with Autorenumber for subsequent drawings), so that when editing the interpolation, it fits into the sequence order.

You can do whatever you like (regarding merging), but in my personal opinion, this feature isn't fully developed yet. There's no clear understanding of how to work with it. Ideas supported by practical examples are needed.

manongjohn:
I wouldn't say this isn't "fully developed yet". As far as I know, it is working as intended by the original developer of this feature. Could there be more that was intended? I couldn't say, but from what I understand, this was it. How it will be used and shortcoming that come from more use will likely be exposed over time, but this is true of any new feature, to be honest.  Should I wait for a feature to be perfect before it is merged? If so, nothing my ever get merged.

Having said that, I too had concerns about adding new drawings between existing drawings once keys have been set.  Doing this would require going to all key'd drawings after it to be corrected for shifted drawing numbers.  However, if they are duplicating a drawing, consider the intent. If they are trying to create an inbetween to smooth out the movement of the interpolation by modifying an existing drawing just before it, then this duplication in the middle is probably intentional.  If not, they can always renumber the duplicate to the end.

Maybe the answer is if a new drawing is added in the middle of a sequence, any Drawing # keys would be automatically updated to account for the new drawing if drawing numbers are shifted.

I will leave this open for another day or so to allow for additional feedback on this feature and/or this discussion.

melieconiek:
> Maybe the answer is if a new drawing is added in the middle of a sequence, any Drawing # keys would be automatically updated to account for the new drawing if drawing numbers are shifted.

This feature is intended for drawing replacement rigs, not timing inbetweens. Most will be using it to select specific drawings for rig body parts (different hands, mouth shapes) or through controllers like below to select body part rotations. Automatic shifting Drawing # keys _might_ be useful in some cases (say you wanted to add an additional hand shape without shifting anything) but would not be needed in most.

I would keep it as is, at least for now. It is very much complete.

image

https://github.com/user-attachments/assets/7e4ffdec-3084-4d0a-8e0a-4cb727fe0b33

manongjohn:
Thanks for testing. Moving forward with this as is.  Discussions regarding this new feature can always be continued as needed either here or in Discord.

Many thanks to @conicsectionn for working on bringing this feature to T2D.

PR #2119, opened 2026-04-12
Tape Tool Freehand Mode & Line Extension Method

### LINKED ISSUE #1764: Tape Tool - Lasso Mode
It is difficult to make precise selections with the Tape Tool in Rectangular mode. Rotating the table can also make selections unintuitive since the rectangle remains oriented with the table not the viewer.

A Lasso mode like those available for other tools would make closing gaps faster and more precise.

### LINKED ISSUE #1765: Tape Tool - Line-to-Line should favor closing straight ahead.
If the tape tool is making a connection line-to-line it should attempt to close straight ahead if possible. If connecting straight ahead would 'miss' the connecting line then it should keep the closest connection -- the current behavior.

This would be the desired result in most cases. Often I'll have to close a line 'manually' since the tape tool would close the gap at a strange angle.

When changes are made to the region gap close logic it should probably adopt the same behavior for the same reason.

Image
Image

melieconiek:
This replaces the tape tool's closest point method with a line extension method and adds a freehand selection mode. The new method can reliably close entire drawings without creating excessive, unnecessary closures. Normal mode has not been changed. This feature was created by @TomDoingArt, I only cleaned things up and fixed a few bugs.

https://github.com/user-attachments/assets/1bf7e47a-f18b-4aef-8245-3f773acad141

https://github.com/user-attachments/assets/acc9c2b5-7331-486e-b086-edb1938e72fe

https://github.com/user-attachments/assets/6a315f71-0830-4cfc-af1b-9b661bb5b17d

Enable `View` ⟶ `Gap Check` to show gap closures and extensions. The center extension is used to find point-2-line connections. Intersections of outer extensions are used to find point-2-point connections. If you select only a few endpoints it will use the closest connections within the selection. Leave a small amount of space between endpoints for the best results.

### SETTINGS: 
- `Min` sets a lower limit to the size of possible closures. Endpoints that are just over a line this length are considered closed and ignored.
- `Max` sets the length of the line extensions and the maximum size of gap closures.
- `Line Ext. Angle` sets the angle of the outer extensions.

https://github.com/user-attachments/assets/9a971127-4e23-4971-b1c9-2b40b61aaab9

image

### KNOWN ISSUES:
- This closure method has no way of dealing with near misses when finding point-2-line closures. It always looks straight ahead. 
- Points where three or more endpoints meet will sometimes not close properly. Doing a second pass will sometimes help.
- Hooks cause the endpoint to face the wrong direction.  Hooks can be quite small so the problem may not be immediately obvious. Erasing a bit of the end of the line will fix this quickly. 

closes #1765 
closes #1764

### DIFF
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -21,39 +21,12 @@
 
 #include <algorithm>
 
-#include "tpalette.h"
 #include "tmathutil.h"
 #include "tenv.h"
 
-#include "../tnztools/controlpointselection.h"
-
-bool debug_mode_1 = false;  // Set to false to disable debug output
-#define DEBUG_LOG(x) if (debug_mode_1) std::cout << x // << std::endl
-
-bool debug_mode_2 = false;  // Set to false to disable debug output
-#define DEBUG_LOG2(x) if (debug_mode_2) std::cout << x // << std::endl
-
-// TomDoingArt ---- TapeTool Freehand - end
-
-#include "tvectorimage.h"  // make sure this include is at the top
-//#include "strokehookfixer.h"
-
-//#include "tools/tapeenv.h"
-
-//extern TEnv::DoubleVar AutocloseFactorMin;
-//extern TEnv::DoubleVar AutocloseFactor;
-//extern TEnv::DoubleVar TapeStartAt;
-//extern TEnv::DoubleVar TapeIncBy;
-
 TEnv::DoubleVar AutocloseFactorMin("InknpaintAutocloseFactorMin", 1.15);
 TEnv::DoubleVar AutocloseFactor("InknpaintAutocloseFactor", 4.0);
-
-TEnv::DoubleVar TapeDehookFactorMin("InknpaintTapeDehookMin", 0.01);
-TEnv::DoubleVar TapeDehookFactorMax("InknpaintTapeDehookMax", 0.20);
-TEnv::DoubleVar TapeDehookAngleThreshold("InknpaintTapeDehookAngleThreshold", 30.00);
 TEnv::DoubleVar LineExtensionAngle("InknpaintTapeLineExtensionAngle", 0.30);
-
-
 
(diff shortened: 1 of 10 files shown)


melieconiek:
Sorry, I thought I had fixed this. I will verify that it builds before attempting again.

melieconiek:
I fixed the build issues and rebased the PR on the latest commit.

beeheemooth:
I tested this feature a bit—it looks great. For some reason, the Line Ext. Angle option isn't available for the Rectangular Type. Is it specifically disabled for this type?

melieconiek:
> I tested this feature a bit—it looks great. For some reason, the Line Ext. Angle option isn't available for the Rectangular Type. Is it specifically disabled for this type?

No this was an oversight. `Line Ext. Angle` is now adjustable when `Rectangular` type is enabled. I also made it possible to enable/disable  `Join Vectors` when `Line to Line` mode is active -- it's always been possible to do p2p closures with `Line to Line` mode so this limitation never made sense.

@beeheemooth Thank you for testing!

manongjohn:
> This closure method has no way of dealing with near misses when finding point-2-line closures. It always looks straight ahead.

Do you have an example of this?

Beyond that LGTM.

Thanks @TomDoingArt and @melieconiek for working on this!

Aside: Perhaps we should look at extending this logic to autoclose?

melieconiek:
@manongjohn 
> Do you have an example of this?

Near miss:
image

And there line intersection issues:
image

The solution we had discussed for near misses was to add a second set of adjustable extensions for the point-2-line checks. The green lines would be looking for line intersections, the blue looking for intersections with other blue extensions for point-2-point connections.
image


> Aside: Perhaps we should look at extending this logic to autoclose?

With some improvements maybe? This works very well for the tape tool, but it's probably too finicky for something that's supposed to be automatic. I've tried setting autoclose to the minimum value and using the tape tool to close everything instead. It works pretty well, but requires some manual work to get everything closed. I do like having more control over it.

Not relevant, but setting the autoclose max to a smaller value causes that bug where regions are formed along lines to happen more often. Only real downside I've encountered.

PR #2129, opened 2026-04-19
Update OS X reference to macOS in README

C0rn3j:
It's not been OS X for quite some time, update to macOS.


manongjohn:
Thanks for the update.

PR #2138, opened 2026-04-25
Create level in empty cell or new column/layer

manongjohn:
This adds an option to the `New Level` popup allowing the user to create the new level in an empty cell or a new column.

This option is only active and applicable when on an empty cell (not explicit cell, not implicit cell, not a stop frame) in a column/layer that already contains a level.  The default is to create it in a new column.

This was added to keep new users from unintentionally creating multiple levels in the same column.

### DIFF
--- a/toonz/sources/toonz/levelcreatepopup.cpp
+++ b/toonz/sources/toonz/levelcreatepopup.cpp
@@ -56,6 +56,7 @@
 #include <QComboBox>
 #include <QPushButton>
 #include <QMainWindow>
+#include <QCheckBox>
 
 using namespace DVGui;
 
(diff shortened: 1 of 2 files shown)


beeheemooth:
How to disable this checkbox (if necessary)?

manongjohn:
Nothing to disable. Only when you are on an empty cell in a column that has a layer, it will become enabled and you can choose what you want to do.

The normal behavior, even before this change, is to always create in a new column if you are on an explicit or implicit cell. The old behavior was to create level in the same column when you are on an empty cell.  This is just a checkbox to let you choose what happens when you are on an empty cell.

beeheemooth:
I think your description should be supplemented with clarification that the option becomes available in the current column after frames are exposed (for Explicit Hold), or after Stop Frame Hold (for Implicit Hold). "Empty Cell" sounds confusing.

PR #2142, opened 2026-04-25
Fix Drawing Substitution with implicit cells

manongjohn:
This change technically restores the original intent of `Drawing Substitution` (shortcuts Q/W) when working with implicit cells. 

The original intent of the command was just to switch the currently selected frame's drawing without affecting anything else.  In Explicit mode, this causes the next frame to appear to show the original frame, which makes sense as it's an explicit hold.  When I implemented implicit mode, I forced this behavior unto implicit frames, when it really shouldn't have.  There is no frame there.

So now, when substituting a drawing, the adjacent implicit frame is no longer forced to be an explicit frame of the original drawing. Per implicit logic, it is now an implicit hold of the substituted drawing.  No other frames will be modified.

As a result of this change, the `Similar Drawing Substitution` (shortcuts Alt+Q/W) is somewhat obsolete in Implicit mode. The only difference between both sets is what happens when the next frame matches the substituted frame. 

Also, due to this change, it is possible that you will have an adjacent explicit held frame if the substitution matches the next frame.  For example, if you have frames `5 6` and substitute the `5`, it will be come `6 6` not `6 _`. Another example would be `5 _ _ _ 6` -> `6 _ _ _ 6`

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -1043,11 +1043,6 @@ public:
     TXsheetP xsh = TApp::instance()->getCurrentXsheet()->getXsheet();
     while (c <= m_range.m_c1) {
       col = c;
-      // Convert implicit cell at end of selected range into a populated cell
-      if (xsh->isImplicitCell((m_range.m_r1 + 1), col)) {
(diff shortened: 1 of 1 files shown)


beeheemooth:
Perfect!
Thanks!

PR #2143, opened 2026-04-26
Output (Render) settings enhancements

### LINKED ISSUE #579: [Minor change] Rename "Output Settings" to "Render Settings" or "Render Options"
This would make the naming of the menu item consistent to the functionality and the other options in the Render menu. Also, "Render Settings" is the name that new users look for when setting up their first render, so it would aid usability.

### LINKED ISSUE #827: Append Sequence or Timestamp To Render Output Filename
### Issue
When I output a render I want to be able to create a new versioned file not overwrite the prior file. To do this currently I have to keep manually updating the filename.

I want the ability to automatically append a sequence number or timestamp to the filename.

### Proposed Solution
There are two places I can see this feature appearing in the menu.

**Output Settings > Options > File Settings popup**

AppendToFilename_OptionsPopup

File Settings is specific to file type, so the option can be ticked/unticked per file type (I assume).

**Output Settings**

AppendToFilename_OutputSettings

Since Output Settings is not specific to file type, the option selected will apply for all file types.

The two types of information which can be appended to the filename, Timestamp and (render) Sequence:

**Timestamp**
Append the UTC timecode to the filename, separated by a '.'.
ISO 8601 timecode [wikipedia](https://en.wikipedia.org/wiki/ISO_8601)

Example:
filename.20211022T093410Z.gif
filename.20211022T095021Z.gif

On a file type that renders as a sequence, a single timecode value at the render level and sequence numbers at the frame level:
filename.20211022T093410Z.0001.png
filename.20211022T093410Z.0002.png
a later render...
filename.20211022T095021Z.0001.png
filename.20211022T095021Z.0002.png

**(Render) Sequence**
Append the next value in a four digit sequence.
The sequence number is separated from the filename by a '.'.

Example: 
filename.0001.gif
filename.0002.gif

On a file type that renders as a sequence, a sequence value at the render level, and sequence numbers at the frame level:
Example, the first render: 
filename.0001.0001.png
filename.0001.0002.png
the next run of render...
filename.0002.0001.png
filename.0002.0002.png

The options are mutually exclusive. Checking one will uncheck the other. I did not see any radio buttons being used in Tahoma2D...so I used checkboxes in these screen mockups.

### LINKED ISSUE #1405: Add a checkbox on output settings to synch the range up with the current play range.
The opportunity exists to automatically keep the output range in synch with the play range. This will help when an animator has used the play range to preview an animation and wants to then render it. Currently the output range is not updated when the the play range is changed.

image
I suggest to add a checkbox to the output settings that will match the output settings start and stop frame to the current play range which was set on the xsheet or timeline.
With the box ticked, any time the play range is changed, the output range will update to match it.

manongjohn:
Following enhancements have been made to `Output Settings`

image

1. Adds `Sync with Play Range` which will sync the render range to the play markers as they are moved.
  `Preview Settings` panel does this automatically so it was not added there.

2. Adds `Append to Name` which will add a sequence number or timestamp to the filename.
   For raster file renders, the sequence # or timestamp (formatted `yyyyMMddThhmmss`) is added between the name and the drawing sequence #:   `<filename>.<version>.<drawing #>`

4. Relabeled `Output Settings` to `Render Settings`

closes #1405 
closes #827
closes #579

### DIFF
--- a/toonz/sources/include/toutputproperties.h
+++ b/toonz/sources/include/toutputproperties.h
@@ -71,6 +71,8 @@ Can set in output to all levels, only selected levels or only animated levels.
 
   enum MaxTileSizeValues { LargeVal = 50, MediumVal = 10, SmallVal = 2 };
 
+  enum AppendVersionFormat { None = 0, Sequence, Timestamp };
+
 private:
   TFilePath m_path;
 
@@ -107,6 +109,8 @@ private:
   int m_nonlinearBpp;
 
   bool m_syncWithPlayRange;
+
+  AppendVersionFormat m_appendVersionFormat;
 
 public:
   /*!
@@ -254,6 +258,11 @@ machine's CPU).
 
   bool isSyncWithPlayRangeEnabled() { return m_syncWithPlayRange; }
   void setSyncWithPlayRangeEnabled(bool sync) { m_syncWithPlayRange = sync; }
+
+  AppendVersionFormat getAppendVersionFormat() { return m_appendVersionFormat; }
+  void setAppendVersionFormat(AppendVersionFormat format) {
+    m_appendVersionFormat = format;
+  }
 };
 
 //--------------------------------------------
--- a/toonz/sources/include/toutputproperties.h
+++ b/toonz/sources/include/toutputproperties.h
@@ -106,6 +106,8 @@ private:
   // for restoring bpp when setting the color space back to nonlinear
   int m_nonlinearBpp;
 
+  bool m_syncWithPlayRange;
+
 public:
   /*!
 Constructs TOutputProperties with default value.
@@ -249,6 +251,9 @@ machine's CPU).
   void syncColorSettings(bool sync) { m_syncColorSettings = sync; }
   int getNonlinearBpp() { return m_nonlinearBpp; }
   void setNonlinearBpp(int bpp) { m_nonlinearBpp = bpp; }
+
+  bool isSyncWithPlayRangeEnabled() { return m_syncWithPlayRange; }
+  void setSyncWithPlayRangeEnabled(bool sync) { m_syncWithPlayRange = sync; }
 };
 
 //--------------------------------------------
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -2338,10 +2338,10 @@ void MainWindow::defineActions() {
   // Menu - Render
 
   createMenuRenderAction(
-      MI_OutputSettings, QT_TR_NOOP("&Output Settings..."), "Ctrl+O",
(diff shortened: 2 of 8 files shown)


gab3d:
Frame Start/End remain editable when Sync is active?
(...so changing them here will adjust the Playback Range?)

gab3d:
I'm sure it's a good thing, but I don't get what is the Sequence option really does. 🤔

manongjohn:
> Frame Start/End remain editable when Sync is active? (...so changing them here will adjust the Playback Range?)

Yes they remain active so you can still edit it as needed.

Changing the value does not change the playback range, though.

manongjohn:
> I'm sure it's a good thing, but I don't get what is the Sequence option really does. 🤔

Just a way to version your rendered output in case you need it.

Read #827 and see if that clarifies it more.

gab3d:
> > Frame Start/End remain editable when Sync is active? (...so changing them here will adjust the Playback Range?)
> 
> Yes they remain active so you can still edit it as needed.
> 
> Changing the value does not change the playback range, though.

ok, so in that case (manually editing the values), the Sync option state will change back to disabled?

manongjohn:
> > > Frame Start/End remain editable when Sync is active? (...so changing them here will adjust the Playback Range?)
> > 
> > 
> > Yes they remain active so you can still edit it as needed.
> > Changing the value does not change the playback range, though.
> 
> ok, so in that case (manually editing the values), the Sync option state will change back to disabled?

No.  Just allows you to override what is there if you want to.  It basically does what Preview Panel does, but you may not want that all the time.

gab3d:
> > I'm sure it's a good thing, but I don't get what is the Sequence option really does. 🤔
> 
> Just a way to version your rendered output in case you need it.
> 
> Read #827 and see if that clarifies it more.

ok, thanks
yes, the intention is clear to me now.
So, the Sequence number is determined how? 
Is it incremental based on what?

manongjohn:
Sequence starts at 1 and incrementally updated based on the last numbers version it finds

gab3d:
incremental, for the same output base name?
so if I change the base name the Sequence number will return back to 1?

manongjohn:
Yep

DarrenTAnims:
I like the look of this.
I'll have to try this out, but I'd expect that when you tick the box, the start/end would be disabled and show the numbers that will be rendered, and to enter new numbers you'd have to untick the box and then they'd become editable.

manongjohn:
Consider you can disable the play range markers. In that case the from/to box needs to be editable so you can enter what you want.

Alternately you may wish to quickly change to a different range without having to go to the timeline and click-drag marks to where you want it.

For these reasons I dont think it would be good to lock the fields while the feature is enabled.

Merging. Can discuss further on Discord if desired.

PR #2144, opened 2026-04-27
Preferences popup enhancement

manongjohn:
The `Preferences` popup has been enhanced as follows:

image

- The popup is now resizeable.
- The right panel is now a scrollable area with all preferences listed.  Preferences are grouped by their categories.
- Clicking on a category in the left panel will display only that category in the right panel. Clicking in the empty space after the last category will show the full list again
- A `Search` box has been added to show only the preferences and the categories it is in that meet the criteria.  Search is based on text in labels.

### DIFF
--- a/toonz/sources/toonz/preferencespopup.cpp
+++ b/toonz/sources/toonz/preferencespopup.cpp
@@ -58,6 +58,8 @@
 #include <QListWidget>
 #include <QGroupBox>
 #include <QCheckBox>
+#include <QScrollArea>
+#include <QScrollBar>
 
 using namespace DVGui;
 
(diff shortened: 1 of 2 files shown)


gab3d:
It's great to have the Search field.
Great job! :)
Thx!


(I'm not a big fan of having everything as a continuous page, though. But, if it is what's needed for the search feature, I embrace it) 😅

manongjohn:
> (I'm not a big fan of having everything as a continuous page, though. But, if it is what's needed for the search feature, I embrace it) 😅

If this is generally preferred, I can make it so clicking on the category list only shows that group, similar to before. When searching it shows all matches regardless of category.  Clearing the search field or clicking back on any category will show that full category again.

I can go either way.  In some cases I like being able to scan through the entire list without clicking through the different windows.

Open to discussion.

manongjohn:
Actually, no further discussion needed.  I've modified it so it shows the full list by default.  When clicking on a category, it only shows that category.  To show the full list again, need to click on an empty space in the category list.

PR #2150, opened 2026-04-30
Save/Restore Onion Skin markers

manongjohn:
This will save relative Onion Skin marker information as environment information so it can be restored on startup.

As  a result of this change, relative onion skin markers are no longer cleared when switching scenes.


melieconiek:
I would save/restore relative onion skins only -- fixed onionskin markers should be cleared when loading a new scene. 

If you've set fixed onion skin markers for a scene, they're not likely to be relevant for another scene. It's the relative onion skin markers you'd expect to remain active and be restored.

manongjohn:
> I would save/restore relative onion skins only -- fixed onionskin markers should be cleared when loading a new scene.

Makes sense. I will update it later this evening.

PR #2151, opened 2026-04-30
Fix Release v1.6.1

manongjohn:
This is to mark and update the nightly to fix release version v1.6.1 in the master.

Version 1.6.1 is being build in a separate branch containing primarily fix PRs which I will be releasing tomorrow (5/1)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -40,8 +40,9 @@ body:
       label: Software Version
       description: What version of the software are you running?
       options:
-        - 1.6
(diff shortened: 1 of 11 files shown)


gab3d:
Are there any new strings to translate for this release? (I don't know when it's scheduled to happen) 🤔

manongjohn:
I have to check each PR to see if I added or modified any translatable strings.  I try not to on Fix releases if it can be helped.

I plan on releasing tomorrow morning (my morning at least). We can always include translation fixes in the next Fix Release, which I try to do every 2 months.

gab3d:
ok, let's do that. 👌

PR #2152, opened 2026-05-01
Level Edit switch level with column change

manongjohn:
In Scene Edit mode, when changing columns, the Level Strip changes level to match the level and drawing at the current cell indicator. 

This PR allows the same thing to happen in Level Edit mode.

### DIFF
--- a/toonz/sources/toonz/tapp.cpp
+++ b/toonz/sources/toonz/tapp.cpp
@@ -355,12 +355,13 @@ int TApp::getCurrentImageType() {
 
 //-----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


beeheemooth:
You solved the problem of multiple levels in a single column by displaying the current frame's drawing—a very clever solution!

Hats off to you!

It seems to work without any problems.

manongjohn:
> You solved the problem of multiple levels in a single column by displaying the current frame's drawing—a very clever solution!

Cant really take any credit for that. It was already doing that in Scene Edit mode. I mostly just let it do whatever Scene Edit mode was doing to the Level Strip. 😁

PR #2174, opened 2026-05-06
Fix changing cell during playback

manongjohn:
This restores the ability to change the location of the current cell during playback which was inadvertently removed with #2100 to fix the erratic ping pong behavior.


melieconiek:
Does what it says on the tin, no issues as far as I can see. Thank you for the fix!

PR #2185, opened 2026-05-14
Vector selection toolbar options: Show/Flip stroke direction

manongjohn:
At the far end of the Vector Selection Tool's option bar are 2 new options
- `Show Direction` - shows the direction of selected strokes
- `Flip Direction` - flips the direction of selected strokes
image

Both options can be assigned shortcuts.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -434,7 +434,7 @@ class SelectionToolOptionsBox final : public ToolOptionsBox,
   ToolOptionIntSlider *m_miterField;
 
   QPushButton *m_hFlipButton, *m_vFlipButton, *m_leftRotateButton,
-      *m_rightRotateButton;
+      *m_rightRotateButton, *m_flipStrokeButton;
 
 public:
   SelectionToolOptionsBox(QWidget *parent, TTool *tool,
@@ -453,6 +453,7 @@ protected slots:
   void onFlipVertical();
   void onRotateLeft();
   void onRotateRight();
+  void onFlipDirection();
 };
 
 //=============================================================================
--- a/toonz/sources/tnztools/selectiontool.h
+++ b/toonz/sources/tnztools/selectiontool.h
@@ -475,6 +475,7 @@ signals:
   void clickFlipVertical();
   void clickRotateLeft();
   void clickRotateRight();
+  void clickFlipDirection();
 };
 
 #endif  // SELECTIONTOOL_INCLUDED
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -1808,6 +1808,18 @@ SelectionToolOptionsBox::SelectionToolOptionsBox(QWidget *parent, TTool *tool,
     m_miterField->setEnabled(m_joinStyle->currentIndex() ==
                              TStroke::OutlineOptions::MITER_JOIN);
 
+    addSeparator();
+    if (tool && tool->getProperties(2)) tool->getProperties(2)->accept(builder);
+
+    m_flipStrokeButton = new QPushButton(tr("Flip Direction"), this);
+    int buttonWidth =
+        fontMetrics().horizontalAdvance(m_flipStrokeButton->text()) + 10;
+    m_flipStrokeButton->setFixedWidth(buttonWidth);
+    m_flipStrokeButton->setFixedHeight(20);
+    connect(m_flipStrokeButton, SIGNAL(clicked()), this, SLOT(onFlipDirection()));
+
+    hLayout()->addWidget(m_flipStrokeButton);
+
     onPropertyChanged();
   }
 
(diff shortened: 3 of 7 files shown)


beeheemooth:
Users ask if it is possible to also show the number (index) of the strokes?

manongjohn:
> Users ask if it is possible to also show the number (index) of the strokes?

It is possible as I have done it with some of my localized testing. I guess I question how is that helpful or to be used?

This is likely an option that would need to be turned on as needed. As the bar is already packed, I am not sure I want to add more to the toolbar unless it is really necessary.

beeheemooth:
I don't think we need a separate checkbox for numbers. In my opinion, numbers and stroke direction serve the same purpose: you can see the number and stroke direction in one drawing and then find it in another to change its direction or redraw it. That's probably the workflow.

RodneyBaker:
I hesitate to add this as I am  looking forward to seeing these options get into the main T2D release sooner rather than later.  Hopefully this can be addressed without too much trouble.

There are color fill issues with flipping the direction of vector strokes.
A way to test this (there are likely other ways but this I have found to be repeatable):
- Create vector level
- Draw strokes to create enclosed area
- Turn on options to display and flip stroke direction
- Select a stroke or strokes and flip their direction.
- Fill will a new color
- Now, outside of that filled shape draw a new stroke using the previous shape's fill color.
- In some cases all of the fill will disappear while in other cases part of the fill will disappear
- Attempting to refill the areas will likely fail although Undoing and then using Freehand fill can sometimes refill the shape.
- Often the portion of the shape that had the fill removed cannot be refilled.
- Importantly:  If the vector level is saved and reloaded the shape apparently can be refilled.
There are other times when the direction of the stroke prevents filling of the section controlled by the stroke that has been flipped.  Saving the vector level appears to resolve this. (To specifically test for this, draw a shape with sufficient curves and flip one stroke.  Fill the main shape and note how the part adjacent to the flipped stroke doesn't fill and cannot be filled.  Save and the unfilled are should resolve)

Short of that, just work with shapes that are filled while flipping stroke direction and continuing to draw.  Fill problems should appear.

beeheemooth:
Hmm... Indeed, when changing the stroke direction, for some reason, closed areas stop being filled. Here are the steps.

https://github.com/user-attachments/assets/c2511b1d-be4f-4a33-8248-d961af576a10

RodneyBaker:
In the 'great news' department, the ability to flip stroke direction is a major win for auto-inbetweening!

The current workflow to avoid the color fill woes would appear to be to use this feature during line drawing, save that line drawing and then do the color work.  

flipstroke

It's so nice to be able to direct auto-inbetweening by flipping the direction of strokes.

manongjohn:
I am not surprised that flipping the direction may cause fills to misbehave as direction can change how paths are "walked" to find fillable regions.

This isnt a new bug, by any means. The fill logic is just as flawed as it was yesterday. Just may bring it more to the surface as flipping vectors could likely happen more often. On the flip side, it could also help if the original direction made it unfillable before. 🙂

RodneyBaker:
All good.
This ability to flip direction of strokes is great stuff and I regret we didn't get Shun's implementation into Opentoonz as we decided not to go forward with it due to the color fill issue being potentially confusing.  In hindsight I wish we'd gone ahead and implemented the fix.  I couldn't find that PR so I think perhaps Shun deleted it.

autoinbetweening

Aside:  It would be nice to be able to assign a shortcut key to trigger xsheet/timeline auto-inbetweening but I'm just really glad we can tween there as we can add Navigation tags to let us know where breakdown frames are for major changes of strokes/shapes so we can tween different parts of a sequence accordingly.

RodneyBaker:
Since we are here I note another potential glitch with auto-inbetweening (outside the scope of this PR or course):

If we move strokes forward or back this can do some strange things with autoinbetweening.
As such I recommend leaving that push/pulling of strokes in depth until after the tweening is complete.  Save.  And then experiment further as you'll likely otherwise not be able to return to that original order easily.  This is apparently a glitch with 'direction' but not stroke direction.  Rather the stored order of strokes or control points in z-depth.

I'll try to keep an eye on that aspect of autoinbetweening and see if I don't find some patterns.

manongjohn:
> I regret we didn't get Shun's implementation into Opentoonz as we decided not to go forward with it due to the color fill issue being potentially confusing.

I dont recall there being a PR in OT that directly allowed you to manually flip stroke direction like this. Do you recall anything particular about it?

Only stroke flipping related PRs i recall and found were an option when converting to vector (by shun) and the vector guided tweening (by me).

I checked Issues for a discussion on flipping stroke direction and didn't see anything there.

As far as I know you cant delete a PR, though sometimes the search omits things if the author is ghosted.

RodneyBaker:
My memory is that it related to Tape Tool automatically slipping the direction of the stroke when connecting the end of a stroke to the middle of another.  It related to a stroke flipping bug @beeheemooth had reported.

It wasn't a direct means of selecting and flipping strokes but rather prevented the automatic flipping of strokes during certain processes.

manongjohn:
Oh. That PR is still open and marked WIP.

RodneyBaker:
Oh.  Crazy.  The last time I looked for that I couldn't find it.
REF: https://github.com/opentoonz/opentoonz/pull/2355

PR #2188, opened 2026-05-16
Drawing Mark

manongjohn:
This provides the ability to add a mark to drawings that is only visible in the Level Strip and Timeline/Xsheet.  The mark is shown in the upper left corner of the drawing icon or exposed cell.  This feature only applies to drawing levels.

image

This is an extension of the `Cell Mark` feature and uses the same mark list set in `Scene Properties`. `Drawing Mark` is applied to individual drawings and follows the drawing as it is moved around the timeline/xsheet.

Drawing marks can be set from either the Level Strip or Timeine/Xsheet. The option is found in the context menu.
image

Similar to cell marks, the options can be added to command/quick/main toolbars as well as set shortcuts for them.

### DIFF
--- a/toonz/sources/include/toonz/txshsimplelevel.h
+++ b/toonz/sources/include/toonz/txshsimplelevel.h
@@ -322,6 +322,10 @@ The oldFp is used when the current scene path change...
   bool canConvertSingleFileToSequence();
   void convertSingleFileToSequence(TXsheet *xsh);
 
+  std::map<TFrameId, int> getDrawingMarks() const { return m_drawingMarks; }
+  int getDrawingMark(const TFrameId &fid) const;
+  void setDrawingMark(const TFrameId &fid, int markId);
+
 public:
   // Auxiliary files management: hooks, tpl, etc.
   // May throw; copy and rename perform touchparentdir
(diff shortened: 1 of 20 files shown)


papio-pl:
It's so good. Thank you. I used to draw some marks next to a drawing to distinguish key frames from inbetweens. 

Do you plan to show a frame mark on looped frames? Maybe a dimmed one. I'm not sure if it's necessary but maybe it's like one line to add one, and it may be useful in big projects:

looped

manongjohn:
> Do you plan to show a frame mark on looped frames? Maybe a dimmed one.

I've just pushed a change to implement this suggestion. Will leave this open for a few more days for additional feedback.

DarrenTAnims:
This looks really useful.
Is it possible to rename the colour text?
So you can set them to: Keys, Breakdowns, Inbetweens, etc ?

manongjohn:
> This looks really useful. Is it possible to rename the colour text? So you can set them to: Keys, Breakdowns, Inbetweens, etc ?

I think this was what people were actually expecting from Cell Marks.
As for renaming the text, you can find it in Scene Settings.  Since Drawing Marks are an extension of Cell Marks, they use the same list.

Due to it's placement in Scene Settings, the list is scene specific, unfortunately, so changing text will likely become bothersome when switching scenes.  I probably need to create a default list in Preferences that new scenes use as a template if not overridden.

PR #2201, opened 2026-05-25
Fix copy-pasting first key as last key

manongjohn:
This fixes an issue where copying the 1st key and pasting it as the last key creates an invalid interpolation type (None) between the previous last key and the newly pasted key.  Additionally, the newly pasted key values are all incorrectly reset to 0.

Issue was caused by changes made in #2104

### DIFF
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -827,7 +827,9 @@ void TDoubleParam::setKeyframe(const TDoubleKeyframe &k) {
   } else {
     if (keyframes.size()) {
       if (it == keyframes.end()) {  // Adding after the last one
-        keyframes.rbegin()->m_type = k.m_prevType;
+        keyframes.rbegin()->m_type = k.m_prevType != TDoubleKeyframe::None
+                                         ? k.m_prevType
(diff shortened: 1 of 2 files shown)


beeheemooth:
Good job! Thank you!

beeheemooth:
By default, the copied first key will always have linear interpolation, as far as I understand?

manongjohn:
Internally Linear is the default after the last key. It changes based on what new key is created afterwards. On brand new keys it uses the preference setting. In the case of a copied key, it will be whatever the interpolation was before the copied key.  Since no key exists before the 1st one, I've defaulted it to Linear.

Open to discussion if we want to do something different

beeheemooth:
If a different interpolation type is selected in Preferences, such as Constant (for blocking), then copying the first key will still create linear interpolation.
In this case, is it possible to bind the copying of the first key to the interpolation type selected in Preferences, or is this difficult due to the context?
Just asking.

manongjohn:
> In this case, is it possible to bind the copying of the first key to the interpolation type selected in Preferences, or is this difficult due to the context?
> Just asking.

Not difficult.

I just checked an older version and the 1s key copied as the last key uses the interpolation set to Preferences.  Somehow I broke that too. Will fix.

PR #2203, opened 2026-05-28
Separate Auto Close from Auto Group + new shortcuts

### LINKED ISSUE #2198: Can "auto close" and "auto group" be independent of each other?
One thing that doesn't make any sense to me: why must for vector levels, the toolbar setting for "auto group" to be enabled, "auto close" has to be as well? I can't really stand auto close, because it always messes up what shape I'm trying to draw. So could the two be more independent of each other, turning one or the other off/on should I desire to?

manongjohn:
This fixes #2198.  

`Auto Close` is now an independent setting in relation to `Auto Group`.

Note: `Auto Close` is still required, along with `Auto Group` when `Auto Fill` is enabled.

Additional changes
- Renames `Brush Tool - Auto Fill Off` shortcut -> `Brush Tool - Auto Close/Group/Fill Off`
- Adds `Brush Tool - Auto Close On` shortcut
- Adds `Brush Tool - Auto Group On` shortcut

### DIFF
--- a/toonz/sources/tnztools/toonzvectorbrushtool.cpp
+++ b/toonz/sources/tnztools/toonzvectorbrushtool.cpp
@@ -2289,26 +2289,17 @@ bool ToonzVectorBrushTool::onPropertyChanged(std::string propertyName) {
   }
 
   if (propertyName == m_autoClose.getName() && !m_autoClose.getValue()) {
+    // We need autofill off if off.
     m_autoFill.setValue(false);
-    m_autoGroup.setValue(false);
-    V_VectorBrushAutoFill  = m_autoFill.getValue();
-    V_VectorBrushAutoGroup = m_autoGroup.getValue();
-    notifyTool             = true;
+    V_VectorBrushAutoFill = m_autoFill.getValue();
+    notifyTool            = true;
   }
 
-  if (propertyName == m_autoGroup.getName()) {
-    // We need close turned on if on,
+  if (propertyName == m_autoGroup.getName() && !m_autoGroup.getValue()) {
     // We need autofill off if off.
-    if (m_autoGroup.getValue()) {
-      m_autoClose.setValue(true);
-      V_VectorBrushAutoClose = m_autoClose.getValue();
-      notifyTool             = true;
-    }
-    if (!m_autoGroup.getValue() && m_autoFill.getValue()) {
-      m_autoFill.setValue(false);
-      V_VectorBrushAutoFill = m_autoFill.getValue();
-      notifyTool            = true;
-    }
+    m_autoFill.setValue(false);
+    V_VectorBrushAutoFill = m_autoFill.getValue();
+    notifyTool            = true;
   }
 
   // we need close and group on
@@ -2472,11 +2463,6 @@ void ToonzVectorBrushTool::loadLastBrush() {
       m_autoGroup.setValue(true);
       V_VectorBrushAutoGroup = 1;
     }
-  }
-
-  if (m_autoGroup.getValue() && !m_autoClose.getValue()) {
-    m_autoClose.setValue(true);
-    V_VectorBrushAutoClose = 1;
   }
 
   switch (V_VectorBrushSnapSensitivity) {
--- a/toonz/sources/tnztools/toonzvectorbrushtool.cpp
+++ b/toonz/sources/tnztools/toonzvectorbrushtool.cpp
@@ -433,11 +433,11 @@ static void addStroke(TTool::Application *application, const TVectorImageP &vi,
                        autoGroup, autoFill, sendToBack));
   }
 
-  if (autoGroup && stroke->isSelfLoop()) {
+  if (autoGroup) {
     int index             = vi->getStrokeCount() - 1;
     if (sendToBack) index = addedStrokeIndex;
     vi->group(index, 1);
-    if (autoFill) {
+    if (autoFill && stroke->isSelfLoop()) {
       // to avoid filling other strokes, I enter into the new stroke group
       int currentGroup = vi->exitGroup();
       vi->enterGroup(index);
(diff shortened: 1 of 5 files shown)


beeheemooth:
The intention was probably that with Auto Close disabled, grouping should occur with open strokes. Otherwise, what's the point of this improvement?

manongjohn:
Separating these settings has been asked by a few different people over several years.  I cant find any compelling reason why Auto Close is required for Auto Group. Unless a compelling reason is presented, I see no harm in separating them.

Aside: I am also not sure why Auto Group of a single stroke is even desired all the time. Grouping of individual strokes prevent them from being used with other strokes to fill a closed region.

melieconiek:
> Aside: I am also not sure why Auto Group of a single stroke is even desired all the time. Grouping of individual strokes prevent them from being used with other strokes to fill a closed region.

This is the reason -- so they don't affect fills. I use large, airbrush stroke as part of shading mattes. Stroke thickness affects autoclose range (as it should) but with the current fill system this causes absolute auto close chaos. Even with smaller strokes there are plenty of reasons you'd want soft or stylized strokes to not affect fills.

beeheemooth:
Flash had the ability to group each drawn stroke. The resulting closed shape of similar strokes couldn't be filled without ungrouping (Break Apart). Grouping strokes was likely necessary to select a stroke entirely if it intersected another stroke, since intersecting strokes were selected all at once (or the portion of the stroke before the intersection).

melieconiek:
- Enabling `Auto Group` on its own will not place the resulting stroke in a group. `Auto Close` must be enabled.
- The shortcut `Brush Tool - Auto Fill Off` does not disable the `Auto Group` setting.
- A request: a direct shortcut for `Auto Group On` and `Auto Group Off` would be handy.

A related issue:
- If the brush tool was already active `Brush Tool - Auto Fill Off` and `Brush Tool - Auto Fill On` brush only work every other time you've activated the brush tool. If it does not work you must switch to another tool then back for it work. This was an existing issue. Don't check who implemented this.

EDIT: I can take a look at the direct shortcut request/issue later if that would be helpful.

manongjohn:
> * Enabling `Auto Group` on its own will not place the resulting stroke in a group. `Auto Close` must be enabled.

Odd, could have sworn I checked that.  Will look into.

> * The shortcut `Brush Tool - Auto Fill Off` does not disable the `Auto Group` setting.

Turning off `Auto Fill` never disabled the other 2 settings.  Prior to this change, turning off `Auto Close` or `Auto Group` always turned off `Auto Fill`, not the reverse.

> * A request: a direct shortcut for `Auto Group On` and `Auto Group Off` would be handy.

Will look into adding

> A related issue:
> 
> * If the brush tool was already active `Brush Tool - Auto Fill Off` and `Brush Tool - Auto Fill On` brush only work every other time you've activated the brush tool. If it does not work you must switch to another tool then back for it work. This was an existing issue. Don't check who implemented this.

I will look into it...and whomever implemented it. lol

melieconiek:
> Turning off `Auto Fill` never disabled the other 2 settings. Prior to this change, turning off `Auto Close` or `Auto Group` always turned off `Auto Fill`, not the reverse.

Fair. The command actually disables `Auto Close` which at the time disabled all three settings. Disabling `Auto Close` no longer does this, so the command should be changed to disable all three and it's name changed to, I don't know, something that makes sense.

EDIT: So three direct shortcuts: `Auto Fill`, `Auto Group`, and one that disables both. I personally wouldn't need one for `Auto Close` by it'self, but that could be added as well to round it out.

manongjohn:
> The command actually disables `Auto Close` which at the time disabled all three settings. Disabling `Auto Close` no longer does this, so the command should be changed to disable all three and it's name changed to, I don't know, something that makes sense.

The Auto Close shortcut is meant only to target that specific setting on the toolbar as if you checked/unchecked it on the toolbar manually. What happens with other settings are done by the tool itself as consequences of turning off that setting and doesn't matter if you are doing that through a shortcut or from the toolbar directly.  Changing the existing shortcut should not be changed.

If you want a new shortcut to disable all 3, with an appropriate name, that can be done as a new request.

Beyond that, to be clear, you are requesting 4 new shortcuts:  Auto Fill On/Off and Auto Group On/Off?  The current short cuts for Auto Fill/Group are toggles that also work on the Geometric Tool.  I think it should work there also, therefore I will move this request to a separate PR dealing with new shortcuts

melieconiek:
> The Auto Close shortcut is meant only to target that specific setting on the toolbar as if you checked/unchecked it on the toolbar manually. What happens with other settings are done by the tool itself as consequences of turning off that setting and doesn't matter if you are doing that through a shortcut or from the toolbar directly. Changing the existing shortcut should not be changed.

The request is for direct shortcuts that both enable the brush tool and apply the settings.  This request is not related to the geometry tool or the individual setting toggle shortcuts.

There are two existing direct shortcuts for the group/auto close setting on the brush tool:
- `Brush Tool - Auto Fill On` activates the brush tool and enables `Auto Fill`. This enables all three settings as result. Aside from the aforementioned bug, this is fine.
- `Brush Tool - Auto Fill Off` activates the brush tool and disables `Auto Close`. Before this disabled all three settings, the intent of the command. The change in this PR breaks this, it needs to be fixed so it disables all three and the command should be renamed to be clearer. Maybe `Brush Tool - Auto Close/Group/Fill Off`.

I requested two additional direct shortcuts:
- `Brush Tool - Auto Close On`
- `Brush Tool - Auto Group On`

So after the change we should have:
- New `Brush Tool - Auto Close On` - activates the brush tool with only `Auto Close` enabled.
- New `Brush Tool - Auto Group On` - activates the brush tool with only `Auto Group` enabled.
- Existing `Brush Tool - Auto Fill On` - activates the brush tool with both all three enabled.
- Fixed/Renamed `Brush Tool - Auto Close/Group/Fill Off` which activates the brush tool with all three disabled.

Sorry for the confusion.

manongjohn:
Ugh,  I just realized I overlooked the other shortcuts and only saw the `Tool Modifiers` shortcuts, not the  `Tools` shortcuts. Misreading/understanding your requests added to it.  Sorry.  

Seems weird to have just `On` commands and no `Off` commands.

I'll take care of the commands with this PR.

PR #2206, opened 2026-05-29
Fix shortcuts for toggling tool option checkboxes

manongjohn:
This fixes an issue where shortcuts set on tool option checkboxes (i.e. `Auto Fill`, `Pressure`, etc.) may not always work the 1st time the shortcut is used or, in at least 1 or 2 cases, not at all.

This can happen when you change the setting from the tool option bar and then try to use the shortcut later and did not switch tools.  Doing this causes the internal state of the tool option bar checkbox and the shortcut to be out of sync.

In most cases repeating the shortcut again will get the states in sync and resume working normally. There is 1 or 2 shortcuts that only triggers if the internal status is in a certain state so it may not work until it syncs up after switching tools.

This corrects the problem by updating the internal status of the assigned shortcut every time the tool option bar checkbox is changed.


beeheemooth:
I finally understood what was going on in #2203. By the way, switching the shortcut for Onion Skin also doesn't work the first time, so it's probably the same issue.

PR #2226, opened 2026-06-11
Remove Drawing/Cell Mark commands

manongjohn:
Follow up to #2188, this adds `Remove Drawing Mark` and `Remove Cell Mark` commands.


manongjohn:
Canceling this and moving the commits to PR #2223

PR #2263, opened 2026-08-09
Fix onion skin and column style support for some vector styles.

melieconiek:
image

This fixes onion skin and column/layer style support for Zigzag, Bump, Vector Brush, Vector Pattern, and Sponge fill styles. All vector line and fill styles on the 'Vector' tab should work properly when visible in an onion skin and will be affected by column/layer styles. 

This does not fix the vector texture brush and fills (the Texture tab.) There was some issue with calling the color function there I couldn't figure out. The texture brushes need support for colorization added and some other bug fixes anyways. These should be addressed in another PR.


manongjohn:
Closing and reopening to get the macOS build error to clear

manongjohn:
LGTM.

PR #2289, opened 2026-09-19
New Opentoonz 1.8 Profile Style and new command bar like 1.8

flaviohr:
Hi! These days, I created a new profile for Tahoma, updating the interface to make it look more like OpenToonz 1.8. I also modified the command bar to make it more similar to this new version.
image
image

The command bar is optional, I will try to create a switch for.


manongjohn:
Please remove the top command bar on any OT 1.8 room definitions that it was added to. It was meant to act as a main bar similar to most other software.

T2D now has a dedicated main bar that appears in all rooms so now there are 2 command bars at the top.  T2D's main bar will be available in 1.7.

manongjohn:
Since the main bar is now a fixed UI element, the Command bar changes you made is not necessary.  Please undo.

manongjohn (changes requested):
Please undo the Commandbar changes.

